PR #80, opened 2020-05-14
Add astmodel support for Functions and Methods

matthchr:


hack/generator/pkg/astmodel/fileDefinition_test.go
```diff
@@ -28,7 +28,7 @@ func NewTestStruct(name string, fields ...string) StructDefinition {
 	}
 
 	ref := NewStructReference(name, "group", "2020-01-01", false)
-	definition := NewStructDefinition(ref, fs...)
+	definition := NewStructDefinition(ref, *NewStructType(fs...))
```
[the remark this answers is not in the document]
matthchr: I think this is fixed now -- `NewStructDefinition` returns a ptr, and the other `New` accepts a `ptr` (since that's what we commonly have)

matthchr:
Created #97 to help track some of the object hierarchy improvement discussions that have come out of this

hack/generator/pkg/astmodel/structtype.go
```diff
@@ -132,5 +133,19 @@ func (structType *StructType) RelatedDefinitions(ref PackageReference, namehint
 		}
 	}
 
+	for name, function := range structType.functions {
+		funcDef := &FunctionDefinition{
+			Function: function,
+			DefinitionName: DefinitionName{ref, name, &namehint},
+			Receiver: &StructReference{DefinitionName{ref, namehint, nil}, false}} // TODO: This false right?
```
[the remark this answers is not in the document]
matthchr: I created an item for this and will leave it as is in this PR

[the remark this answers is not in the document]
Porges: For the StructReference it's not just cleanup it's a correctness issue (or will be in future when we actually use the referenced-resource thing).

hack/generator/pkg/astmodel/function.go
```diff
@@ -0,0 +1,55 @@
...
+
+func (f *FunctionDefinition) AsDeclarations() []ast.Decl {
+	result := f.Function.AsFunc(f.Receiver) // Get the func
+	result.Name = ast.NewIdent(f.Name()) // Give it a name
+
+	return []ast.Decl { result }
+}
+
+func (f *FunctionDefinition) References(t Type) bool {
+	return false // TODO: ??
```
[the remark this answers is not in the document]
Porges: > Yeah there is a tension here which is that, I really want to know the `StructReference` to place methods on, but at the time we are doing handling (i.e. in `oneOfHandler` we return a `Type` not a reference (it has no name yet).

But at the same time we don’t actually need the name yet, we only need it when we emit the definitions 🤔

hack/generator/pkg/astmodel/definition.go
```diff
@@ -13,8 +13,19 @@ type HasImports interface {
 	RequiredImports() []PackageReference
 }
 
+// ReferenceChecker describes an entity that can check if it references a given type
+type ReferenceChecker interface {
+	// Does this include any direct references to the given Type?
+	// "Direct" means we don't walk into any StructReferences (nor could we with these arguments),
+	// but we do walk into included StructTypes.
+	References(t Type) bool
```
theunrepentantgeek, line 21: As discussed, I think this needs to be renamed to `IsReferenceBy()` to make it clear both that this is a boolean test and the direction of that test.

hack/generator/pkg/astmodel/definition_name.go
```diff
@@ -27,6 +29,8 @@ func (dn *DefinitionName) Name() string {
 // it is simply a reference to the name.
 var _ Type = (*DefinitionName)(nil)
 
+// TODO: Should we remove this impl? It seems counter to definitions being types or funcs
+
```
theunrepentantgeek, line 33: Not sure if we can remove it though; perhaps check to see if it's ever called.

hack/generator/pkg/astmodel/file_definition.go
```diff
@@ -56,7 +56,7 @@ func (file *FileDefinition) generateImportSpecs() []ast.Spec {
 
 	var requiredImports = make(map[PackageReference]bool) // fake set type
 	for _, s := range file.definitions {
-		for _, requiredImport := range s.Type().RequiredImports() {
+		for _, requiredImport := range s.RequiredImports() {
```
theunrepentantgeek, line 59: This is nice.

hack/generator/pkg/astmodel/file_definition_test.go
```diff
@@ -28,7 +28,7 @@ func NewTestStruct(name string, fields ...string) StructDefinition {
 	}
 
 	ref := NewStructReference(name, "group", "2020-01-01", false)
-	definition := NewStructDefinition(ref, fs...)
+	definition := NewStructDefinition(ref, NewStructType(fs...))
```
theunrepentantgeek, line 31: Why does `NewStructType()` need to be called here vs embedding that detail within `NewStructDefinition()`?

hack/generator/pkg/astmodel/function.go
```diff
@@ -0,0 +1,64 @@
...
+	// This is only used at the moment to group definitions into files so just returning false
+	return false
+}
+
+// RequiredImports returns a list of packages required by this
+func (f *FunctionDefinition) RequiredImports() []PackageReference {
+	return f.Function.RequiredImports()
+}
+
+var _ Definition = (*FunctionDefinition)(nil)
```
theunrepentantgeek, line 64: In other files this assertion to ensure implementation implementation is before methods (close to the struct) and also commented.

hack/generator/pkg/astmodel/function.go
```diff
@@ -0,0 +1,64 @@
...
+	result := f.Function.AsFunc(f.Receiver) // Get the func
+	result.Name = ast.NewIdent(f.Name())    // Give it a name
+
+	return []ast.Decl{result}
+}
+
+// References indicates whether this function includes any direct references to the given Type
+func (f *FunctionDefinition) References(t Type) bool {
+	// This is only used at the moment to group definitions into files so just returning false
+	return false
```
theunrepentantgeek, line 56: This implementation makes sense for `IsReferencedBy()` - but as we discussed, adding `IsMemberOf()` would help allocate functions to files.

hack/generator/pkg/astmodel/package_definition.go
```diff
@@ -13,16 +13,17 @@ import (
 	"text/template"
 )
 
-// PackageDefinition is the definiton of a package
+// PackageDefinition is the definition of a package
```
theunrepentantgeek, line 16: Thanks!

hack/generator/pkg/astmodel/package_reference.go
```diff
@@ -5,31 +5,65 @@
...
+func (pr *PackageReference) isLocalPackage() bool {
+	return strings.HasPrefix(pr.packagePath, localPathPrefix)
 }
 
-// GroupName is the group name of the package reference
-func (pr *PackageReference) GroupName() string {
-	return pr.groupName
+func (pr *PackageReference) stripLocalPackagePrefix() (string, error) {
+	if !pr.isLocalPackage() {
+		return "", fmt.Errorf("cannot strip local package prefix from non-local package %v", pr.packagePath)
```
theunrepentantgeek, line 37: Capitalize `"Cannot strip ...` and quotes `... package '%v'"` #nitpick

hack/generator/pkg/astmodel/struct_type.go
```diff
@@ -127,5 +133,20 @@ func (structType *StructType) RelatedDefinitions(ref PackageReference, namehint
 		}
 	}
 
+	for name, function := range structType.functions {
+		funcDef := &FunctionDefinition{
+			Function:       function,
+			DefinitionName: DefinitionName{ref, name, &namehint},
+			Receiver:       &StructReference{DefinitionName{ref, namehint, nil}, false}} // TODO: This false right?
```
theunrepentantgeek, line 140: Does Go have named parameters? Anonymous boolean values are icky.

hack/generator/pkg/astmodel/struct_type.go
```diff
@@ -127,5 +133,20 @@ func (structType *StructType) RelatedDefinitions(ref PackageReference, namehint
...
+
 	return result
 }
+
+// WithFunction attaches a function to the struct type
+func (structType *StructType) WithFunction(name string, function Function) *StructType {
+	structType.functions[name] = function
+
+	return structType
+}
```
theunrepentantgeek, line 152: `StructType` is supposed to be immutable, so this function needs to return a copy of the struct, leaving the original unmodified. (This is a hedge against the same `StructType` (same `$ref`) being referenced with different names from different parts of the schema.)

hack/generator/pkg/jsonast/jsonast.go
```diff
@@ -420,7 +420,7 @@ func refHandler(ctx context.Context, scanner *SchemaScanner, schema *gojsonschem
 
 		description := "Generated from: " + url.String()
 
-		sd := astmodel.NewStructDefinition(structReference, structType.Fields()...).WithDescription(&description)
+		sd := astmodel.NewStructDefinition(structReference, structType).WithDescription(&description)
```
theunrepentantgeek, line 423: Agreed, wrapping the entire Struct Type makes more sense than extracting the fields.

hack/generator/pkg/astmodel/file_definition_test.go
matthchr: Because a struct type now has references to functions as well, and you can't have 2 varargs in a function I don't think, and even if we could, I don't think we'd want to.

The fact the definition has-a StructType isn't really a surprise as it's clear from the public definition, what's the goal of hiding it in the new?

hack/generator/pkg/astmodel/function.go
matthchr: Good catch

hack/generator/pkg/astmodel/package_reference.go
matthchr: According to my goland, error strings aren't supposed to begin with an uppercase character (it green-squiggles me if they aren't).

I see here: https://github.com/golang/go/wiki/CodeReviewComments#error-strings -- although it doesn't say why still...

Maybe @devigned knows.

hack/generator/pkg/astmodel/struct_type.go
matthchr: I think I fixed this

hack/generator/pkg/jsonast/exportConfiguration_test.go
```diff
@@ -14,15 +14,15 @@ import (
...
-var student2019 = astmodel.NewStructDefinition(astmodel.NewStructReference("student", "group", "2019-01-01", false))
-
-var address2020 = astmodel.NewStructDefinition(astmodel.NewStructReference("address", "group", "2020-01-01", false))
-var person2020 = astmodel.NewStructDefinition(astmodel.NewStructReference("person", "group", "2020-01-01", false))
-var professor2020 = astmodel.NewStructDefinition(astmodel.NewStructReference("professor", "group", "2020-01-01", false))
-var student2020 = astmodel.NewStructDefinition(astmodel.NewStructReference("student", "group", "2020-01-01", false))
-var tutor2020 = astmodel.NewStructDefinition(astmodel.NewStructReference("tutor", "group", "2020-01-01", false))
+var person2019 = astmodel.NewStructDefinition(astmodel.NewStructReference("person", "group", "2019-01-01", false), *astmodel.NewStructType())
+var post2019 = astmodel.NewStructDefinition(astmodel.NewStructReference("post", "group", "2019-01-01", false), *astmodel.NewStructType())
+var student2019 = astmodel.NewStructDefinition(astmodel.NewStructReference("student", "group", "2019-01-01", false), *astmodel.NewStructType())
```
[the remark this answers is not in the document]
theunrepentantgeek: Yeah, you should. Adding a new **func** should return a new `StructType`.

hack/generator/pkg/astmodel/package_reference.go
theunrepentantgeek: I'm not getting those squiggles.

_Sigh_. If lowercase is idiomatic Go, then we should follow that ... but it will take a while before it looks anything other than amateurish to me.

hack/generator/pkg/astmodel/definition.go
Porges: The direction is `References` not `IsReferencedBy`, since we only recurse through the receiver side, not the argument side.

hack/generator/pkg/astmodel/structtype.go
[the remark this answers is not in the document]
theunrepentantgeek: We need `RelatedDefintions` on _both_ `Definition` and `Type` because it's often (usually?) the Type that needs to generate the definitions - this is true at least for Enums; I've made the requisite change as a part of #101

hack/generator/pkg/astmodel/definition.go
theunrepentantgeek: Clearly we're each interpreting the direction of `References` and `IsReferencedBy` differently. 😁 

I think this proves we need a better name for the method, though I'm not sure quite what that is ...

hack/generator/pkg/astmodel/file_definition_test.go
theunrepentantgeek: That makes sense. Not sure I've got an alternative, but it looks a bit clunky to me.

hack/generator/pkg/astmodel/function.go
theunrepentantgeek: Alternatively, using `References()` in both directions when allocating definitions to files may be better than adding another method.

hack/generator/pkg/astmodel/package_reference.go
matthchr: Yeah I don't love it either but most of our error strings already seem to be following that

devigned:
hmm... I don't think the coverage report is including the code generator. I'll see about adding it.

hack/generator/pkg/astmodel/struct_definition.go
```diff
@@ -165,13 +159,26 @@ func (definition *StructDefinition) AsDeclarations() []ast.Decl {
...
+	return result
+}
+
 // Tidy the content of this struct before generating the AST
 func (definition *StructDefinition) Tidy() {
-	sort.Slice(definition.fields, func(left int, right int) bool {
-		return definition.fields[left].fieldName < definition.fields[right].fieldName
+	sort.Slice(definition.StructType.fields, func(left int, right int) bool {
+		return definition.StructType.fields[left].fieldName < definition.StructType.fields[right].fieldName
 	})
```
theunrepentantgeek, line 182: Given that `StructType` now incluces more than just fields, probably best if this method just delegates to a `Tidy()` method on the `StructType` itself - the methods will need to be sorted too, else we'll end up with diffs caused by them coming out in different orders.

hack/generator/pkg/astmodel/struct_type.go
```diff
@@ -103,6 +112,8 @@ func (structType *StructType) Equals(t Type) bool {
 			}
 		}
 
+		// TODO: Match funcs?
```
theunrepentantgeek, line 115: Yeah, probably should for correctness - we might not need it now, but leaving it out is creating a landmine that might go off later.

hack/generator/pkg/astmodel/function.go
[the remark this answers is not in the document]
matthchr: This tension has been mostly resolved after applying your suggestions @Porges

hack/generator/pkg/astmodel/definition.go
Porges: it’s the same direction as the normal verb isn’t it? `x.References(y)`

hack/generator/pkg/astmodel/struct_definition.go
matthchr: Our interface definitions have `Tidy` on `Definition` though, and right now there are no `Type`'s which have a `Tidy`, so I am not sure we want to violate that pattern?

theunrepentantgeek: We have `Tidy()` on `Definition` to force it, not to prevent it elsewhere. Maybe it needs a different name?

The principle in the back of my mind is that types should be responsible for their own state; having `StructDefinition` reach in to monkey around with the internals of `StructType` isn't desirable.

matthchr: I don't really disagree, just pointing out that if you look at the impl of `Tidy`, 2/3 of them reach into the underlying `Type` and fuss with it.

Specifically, `EnumDefinition` and `StructDefinition` both do - I'm just wondering if the solution to the problem is a separate PR addressing `Tidy` specifically -- although here I probably should at least impl it for functions in `StructType` and I'm not sure how to as I don't know if go maps are ordered...

matthchr: Also as an aside, isn't `Tidy` in `StructDefinition` today violating `StructTypes` "immutability" guarantees (since it's sorting the contents in-place?)

It is _really_ awkward that there's not at least a convention to indicate "Don't fuss with my fields" -- I guess the solution is to put it in a different package...

hack/generator/pkg/astmodel/definition.go
theunrepentantgeek: It's possible to read `x.References(y)` as a mutator that establishes the relationship that `x has a reference to y`, verses reading it as a test for whether that relationship already exists.

I know I'm splitting some mighty fine hairs here, let's just say that I'm scarred by experiences with developers who just don't care about the semantics of a name.

Taking into account the direction - how about `HasReferenceTo()` and also making the check bidirectional when allocating types to files?

hack/generator/pkg/astmodel/struct_definition.go
theunrepentantgeek: That's why I put the first model files into `astmodel` instead of next to the existing code.

matthchr: I have sent PR: #106 which removes Tidy entirely and just relies on immutability and does the sorting on create.

matthchr: If folks approve that I will rebase onto it and then add functions to it

matthchr: Did this

PR #93, opened 2020-05-17
Specification for Versioning of CRDs

theunrepentantgeek:
Initial draft for review and iteration.

2020-06-05 - Complete rewrite for new approach
2020-06-25 - Major update, including two case studies and diagrams


hack/generator/spec/versioning.md
```diff
@@ -0,0 +1,192 @@
...
+
+
+
+## Goals
+
+**Auto-generate conversions:** As far as practical, we want to autogenerate the schema for each hub version along with the conversions to and from the actual ARM API versions. Hand coding all the required conversions doesn't scale across all the different Azure sevices, especially with the ongoing rate of change. 
+
+**Allow for hand coded conversions:** While we expect to use code generation to handle the vast majority of needed conversions, we anticipate that some breaking API changes will require some of the conversion to be hand coded. We need to make it simple for these conversions to be introduced while still automatically generating the majority of the conversion.
+
+**No modification of generated files.** Manual modification of generated files is a known antipattern that greatly increases the complexity and burden of updates. If some files have been manually changed, every difference showing after code generation needs to be manually reviewed before being committed. This is tediuos and error prone because the vast majority of auto generated changes will be perfectly fine.
```
davefellows, line 11: typo: 'tediuos'

hack/generator/spec/versioning.md
```diff
@@ -0,0 +1,192 @@
...
+
+We'll already be generating the required `ConvertTo()` and `ConvertFrom()` methods defined by the `Convertible` interface, as a part of supporting the usual hub-spoke type conversion model.
+
+One possible approach would be to generate a custom interface for each type, with a name based on the ARM API being targetted
+
+``` go 
+package v20200101
+
+type AssignablePerson {}
+    AssignTo(person v1.Person) error
```
matthchr, line 154: minor: assuming `v1.Person` is a struct this should be a ptr

hack/generator/spec/versioning.md
theunrepentantgeek: Fixed.

hack/generator/spec/versioning.md
```diff
@@ -0,0 +1,192 @@
...
+
+## Non-Goals
+
+**Coverage of every case by code generation:** While it's likely that very high coverage will be achievable with code generation, we don't believe that it will be practical to handle every possible situation automatically. It's therefore necessary for the solution to have some form of extensibility allowing for the injection of hand written code.
+
+## Proposed Solution
+
+The "v1" storage version of each supported resource type will be created by merging all of the fields of all the distinct versions of the resource type, creating a *superset* type that includes every property declared across every version of the API.
+
+* To maintain backward compatibility as Azure APIs evolve over time, we will include properties across all versions of the API, even for versions we are not currently generating as output. This ensures that properties in use by older APIs are still present and available for forward conversion to newer APIs, even as those older APIs age out of use.
```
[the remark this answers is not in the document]
theunrepentantgeek: The argument for avoiding accumulation of "storage schema debt" is significant; I've rewritten the proposal to avoid this.

hack/generator/spec/versioning.md
```diff
@@ -0,0 +1,192 @@
...
+}
+```
+
+We can also anticipate changes going in the other direction, where a strongly typed value becomes a **string** in order to provide greater flexibility.
+
+#### Discussion
+
+Do we define the type using `interface{}` to allow any value to be stored, but at the cost of type safety? What does this do to our CRD generation based on these types?
+
+Can we reuse the work being done by @mattchr to handle OneOf schema types?
```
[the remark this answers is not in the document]
theunrepentantgeek: Removed.

hack/generator/spec/versioning.md
```diff
@@ -0,0 +1,192 @@
...
+## Issues
+
+Open issues
+
+* What if the type of a property changes?
+* How do we handle extensibility?
+
+Resolved issues:
+
+* Do we merge all properties from all versions of an API? [Resolved]
```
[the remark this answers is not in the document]
theunrepentantgeek: I've reworked how outstanding issues are shown.

hack/generator/spec/versioning.md
```diff
@@ -0,0 +1,192 @@
...
+    Ends Time
+    // elided
+}
+```
+
+We can also anticipate changes going in the other direction, where a strongly typed value becomes a **string** in order to provide greater flexibility.
+
+#### Discussion
+
+Do we define the type using `interface{}` to allow any value to be stored, but at the cost of type safety? What does this do to our CRD generation based on these types?
```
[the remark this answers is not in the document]
theunrepentantgeek: Yeah, there are a lot of difficult issues with this approach, which is why I've changed it significantly - interested in what you think of the new version.

hack/generator/spec/versioning.md
```diff
@@ -0,0 +1,192 @@
...
+
+**Allow for hand coded conversions:** While we expect to use code generation to handle the vast majority of needed conversions, we anticipate that some breaking API changes will require some of the conversion to be hand coded. We need to make it simple for these conversions to be introduced while still automatically generating the majority of the conversion.
+
+**No modification of generated files.** Manual modification of generated files is a known antipattern that greatly increases the complexity and burden of updates. If some files have been manually changed, every difference showing after code generation needs to be manually reviewed before being committed. This is tediuos and error prone because the vast majority of auto generated changes will be perfectly fine. 
+
+## Non-Goals
+
+**Coverage of every case by code generation:** While it's likely that very high coverage will be achievable with code generation, we don't believe that it will be practical to handle every possible situation automatically. It's therefore necessary for the solution to have some form of extensibility allowing for the injection of hand written code.
+
+## Proposed Solution
```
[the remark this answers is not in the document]
theunrepentantgeek: Good idea. Included.

hack/generator/spec/versioning.md
```diff
@@ -0,0 +1,192 @@
...
+}
+```
+
+This would be incompatible with any existing versions in storage, breaking the backward compatibility promise we should be adherent to.
+
+The problem occurs in different forms regardless of the number of supported versions being retained. A threshold of two versions was used above to keep the example constrained.
+
+#### Resolution
+
+The storage version will be generated by combining all available versions of the API to create a superset, regardless of the number of versions being emitted by the code generator.
```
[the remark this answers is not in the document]
theunrepentantgeek: No longer relevant.

hack/generator/spec/versioning.md
theunrepentantgeek: Fixed.

hack/generator/spec/versioning.md
```diff
@@ -0,0 +1,201 @@
...
+| NodeTypes                         | []NodeTypes                      | []NodeTypeDescription              | Change of Type   | Property Bag |
+| ReliabilityLevel                  | *Level                           | *ClusterPropertiesReliabilityLevel | Change of Type   | Property Bag |
+| ReverseProxyCertificate           |                                  | *CertificateDescription            | New Property     | Skip         |
+| UpgradeDescription                | *PaasClusterUpgradePolicy        | *ClusterUpgradePolicy              | Change of Type   | Property Bag |
+| UpgradeMode                       |                                  | *ClusterPropertiesUpgradeMode      | New Property     | Skip         |
+| VmImage                           | *string                          | *string                            | None             | Copy value   |
+
+Of the 15 properties listed:
+
+* 8 are unchanged (same name and type); these values are simply copied across.
```
matthchr, line 130: Same comment as above -- they all need autogenerated conversion functions don't they?

hack/generator/spec/versioning.md
```diff
@@ -0,0 +1,201 @@
...
+
+In summary (details below):
+
+* Base the storage version on the latest available API version for each resource type
+* Augment the storage version with a name/value property bag for storing additional information
+* Generate the required `ConvertTo()` and `ConvertFrom()` methods automatically
+* Allow configuration to support common changes from version to version
+* Define extension points for hand written conversions
+
+### Use the latest available API Version
```
matthchr, line 31: We aren't actually using the latest API version are we -- we're just using it _as a base_?

So for example if we have 3 API versions currently: `v1`, `v2`, `v3`, then we would also have a `v3storage` basically, right? Or is `v3` == `v3Storage`?

If we're just using it as a base we need to talk about what happens when the base changes. How do we get from `v3storage` to `v4storage`

hack/generator/spec/versioning.md
```diff
@@ -0,0 +1,201 @@
...
+
+As a starting point for defining the storage schema for each resource type, we will use the latest available version of the matching ARM API. This ensures maximum compatibility with that API when it is used, giving a good starting point for new adoption of the service operator.
```
davefellows, line 33: What do you mean by the "matching" ARM API? It's just the latest version of the ARM API, right?

hack/generator/spec/versioning.md
```diff
@@ -0,0 +1,201 @@
...
+
+If a resource type has been dropped from the ARM API, we will still generate a storage schema for it based on the last ARM API version where it existed; this ensures backward compatibility with existing service operator deployments.
+
+### Augment storage with a name/value property bag
+
+To facilitate easy and safe storage of additional configuration in storage, we'll include a name/value property bag for storage, along with helper methods that allow for strongly typed access to values stored within.
```
davefellows, line 39: I'm still a little confused about what you mean by "configuration"...

hack/generator/spec/versioning.md
```diff
@@ -0,0 +1,201 @@
...
+
+We could generate those anew for each type, making them conveniently available and avoiding the need for our generated types to depend on anything more than the standard Go library Alternatively, we could define the property bag as a new struct, wrapping up storage & helper methods into a reusable form.
+
+---
+
+### Generate conversion methods
+
+Each of the structs generated for ARM API will have the normal `ConvertTo()` and `ConvertFrom()` methods generated automatically, mapping like-for-like properties between the hub version and the current type.
+
+* When the name and type of properties match exactly, conversion in both directions will be simple assignments.
```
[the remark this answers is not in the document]
davefellows: Yeah, you have to deepcopy everything...

hack/generator/spec/versioning.md
```diff
@@ -0,0 +1,201 @@
...
+* Where the name matches but the type does not, the property will be skipped by default.
+* If missing on the hub version, the name/value property bag will also be used.
+
+### Configuration for common changes
```
davefellows, line 59: Can you elaborate more on how and where this configuration would be applied? I still feel like I'm not fully groking what you're suggesting...

hack/generator/spec/versioning.md
```diff
@@ -0,0 +1,201 @@
...
+We could generate those anew for each type, making them conveniently available and avoiding the need for our generated types to depend on anything more than the standard Go library Alternatively, we could define the property bag as a new struct, wrapping up storage & helper methods into a reusable form.
+
+---
+
+### Generate conversion methods
+
+Each of the structs generated for ARM API will have the normal `ConvertTo()` and `ConvertFrom()` methods generated automatically, mapping like-for-like properties between the hub version and the current type.
+
+* When the name and type of properties match exactly, conversion in both directions will be simple assignments.
+* Where the name matches but the type does not, the property will be skipped by default.
```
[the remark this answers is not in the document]
davefellows: What do you mean by "skipped"? Not dealt with at all? Is this where it would rely on us explicitly informing the generator how to handle this case? Is this where the configuration comes in?

hack/generator/spec/versioning.md
```diff
@@ -0,0 +1,201 @@
...
+**Property rename** - if a property is simply renamed between versions (say for a spelling correction), we can generate the required assignment for compatiblity between spoke and hub versions.
+
+**Type rename** - if a subtype is renamed between versions, maintaining largely the same structure, we can automatically map between those two types.
+
+
+---
+
+***Outstanding Issue:** What do we do for other kinds of version-to-version change?*
+
+Are there other cases of changes between versions that we may be able to handle automatically.
```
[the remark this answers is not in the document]
davefellows: Well, v3 is still the storage verison, right? As that's the one you're storing... To convert from v2 to v1, it would be 
v2 -> v3 -> v1. The conversion from v2 -> v3 needs to handling converting the real type to the property bag representation and then v3 -> v1 needs to do the opposite. Or am I missing something?

hack/generator/spec/versioning.md
```diff
@@ -0,0 +1,201 @@
...
+``` go
+type ClusterPropertiesv20160301Converter {
+    AssignTo(clusterProperties *ClusterPropertiesvStorage)
+    AssignFrom(clusterProperties *ClusterPropertiesvStorage)
+}
+```
+
+If implemented on the v20160301 version of ClusterProperties, it will be automatically invoked at the end of the standard `ConvertTo()` and `ConvertFrom()` methods, allowing for their behaviour to be modified/augmented.
+
+When the type of a property is changed between versions, we'll generate an interface allowing for conversion between those two types.
```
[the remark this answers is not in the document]
davefellows: Same question :)

hack/generator/spec/versioning.md
```diff
@@ -0,0 +1,201 @@
...
+**Coverage of every case by code generation:** While it's likely that very high coverage will be achievable with code generation, we don't believe that it will be practical to handle every possible situation automatically. It's therefore necessary for the solution to have some form of extensibility allowing for the injection of hand written code.
+
+## Proposed Solution
+
+In summary (details below):
+
+* Base the storage version on the latest available API version for each resource type
+* Augment the storage version with a name/value property bag for storing additional information
+* Generate the required `ConvertTo()` and `ConvertFrom()` methods automatically
+* Allow configuration to support common changes from version to version
```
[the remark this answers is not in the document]
theunrepentantgeek: This is supposed to be a very high level summary of the approach, with sections below going into details. I'll add more introductory text and also make the working clearer.

hack/generator/spec/versioning.md
babbageclunk: I'm confused about this. If a new version comes out and we regenerate the storage type and conversion functions, will the storage type be updated to match the latest API version type? 

Won't that mean that running controllers that wrote a `v2storage` into the k8s API and are then upgraded will try to read a `v3storage`? 

It seems like the storage types need to be versioned independently from the API types. Maybe that's ok - old storage types just become extra spokes for the new hub? Not sure whether the kubebuilder conversion stuff will handle that though.

hack/generator/spec/versioning.md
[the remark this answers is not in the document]
theunrepentantgeek: I definitely need to tweak the wording here as that's exactly what I thought I was saying! If both types have the same property, we can just generate an assignment to copy the data:

```
package vStorage
type Person struct {
    Name string
}

package v20200601
type Person struct {
    Name string
}

func (person *Person) ConvertTo() vStorage.Person {
    result := &vStorage.Person{}

    result.Name = person.Name

    return result
}
```

_Indicative syntax only_.

hack/generator/spec/versioning.md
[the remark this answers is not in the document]
theunrepentantgeek: Yeah, skipped is less than ideal because it's silent. I'll capture this as an outstanding issue.

hack/generator/spec/versioning.md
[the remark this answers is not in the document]
theunrepentantgeek: I was thinking of the situation where you have a collection of items, and we could then autogenerate the looping, requiring manual authoring of just the conversion. Perhaps I'm overthinking this - in the interests of MVP, I'll simplify.

hack/generator/spec/versioning.md
theunrepentantgeek: Why would they need separate functions?

We're going to be generating the central `ConvertTo()` function required by the controller, surely all we need is to include a simple assignment in that function to copy the value from one instance to the next. Yes?

hack/generator/spec/versioning.md
theunrepentantgeek: Correct, if we have 3 API versions currently (`v1`, `v2` & `v3`) then we'll create a hub/storage version based on `v3` called (say) `vStorage`. 

We do need to think through the upgrade scenario. The ASO repo uses a webhook to trigger the upgrade when you deploy a new version of the operator; I'm not sure of the exact details.

hack/generator/spec/versioning.md
theunrepentantgeek: For most resources, yes, it's just the latest version of the ARM API.

But. 

Sometimes services will drop a resource type - e.g. `Microsoft.Authorization` dropped `Locks` between versions.

We mustn't fail to generate the storage type - if we do, the operator will fail because it won't be able to load the item. So, for those, we use the last ARM API that does include the resource type.

hack/generator/spec/versioning.md
babbageclunk: Ah, ok - so there's a moment when the operator can load the old versions of specs and convert them to the new version?

I guess the key thing is that if `vStorage` is based on `v3` because that was the latest API version when `vStorage` was created, we don't necessarily need to update `vStorage` when `v4` comes out (unless there's something that can't be represented).

hack/generator/spec/versioning.md
```diff
@@ -0,0 +1,201 @@
...
+
+### Alternative: Superset storage schema
+
+The "v1" storage version of each supported resource type will be created by merging all of the fields of all the distinct versions of the resource type, creating a *superset* type that includes every property declared across every version of the API.
+
+To maintain backward compatibility as Azure APIs evolve over time, we will include properties across all versions of the API, even for versions we are not currently generating as output. This ensures that properties in use by older APIs are still present and available for forward conversion to newer APIs, even as those older APIs age out of use.
+
+#### Limitation: Old properties aging out
+
+When a changed property ages out and the schema is no longer available, the code generator will be unaware of its existence. Therefore, it will be omitted from our generated storage schema resulting in a breaking change for any resource already persisted.
```
davefellows, line 160: This would only happen if we're not supporting that older version, right? So I think it comes down to the decision on when to stop supporting an older version, recognizing that anyone using that version will likely be broken?

hack/generator/spec/versioning.md
davefellows: I think we would *always* update to v4 (the latest version), right?

hack/generator/spec/versioning.md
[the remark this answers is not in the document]
matthchr: yeah but that's the simplest case - holding true for structs comprised of simple (basic type) properties only.

If you imagine a more complicated struct like:
```
package vStorage
type Person struct {
    Name string
}

type School {
    Students []*Person
}

package v20200601
type Person struct {
    Name string
}

type School {
    Students []*Person
}

func (school *School) ConvertTo() vStorage.School {
    result := &vStorage.School{}

    // This is not just a basic assignment
    for _, person := school.Students {
        result.Students = append(result.Students, person.CovertTo())
    }
    return result
}
```

I'm not super concerned about the specifics of all the different conversions because I'm convinced we can do them all, I'm just highlighting that _a lot of the time_ it's not going to be a simple assignment because many objects in Azure are quite deeply nested, and we need to generate conversion for structs, maps, slices as well. Also (from this), _every single type_ in a version must be convertible to the corresponding type in the storage version if there is one (so there are going to be a lot of `ConvertTo`'s flying around)

hack/generator/spec/versioning.md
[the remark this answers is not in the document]
matthchr: No you have it right Dave -- my point is just that when we put the `v2` representation into the property bag, it's still a `v2` type, not a `v3` type (since there is no `v3` version of that type). So as far as _that particular type_ is concerned, it's kinda still the storage version and all conversions for that type need to go through `v2` (not `v3`, because there is no corresponding type in `v3`).

Basically I am saying I think we're going to end up in a world where: the storage version is not per-api (i.e. what we in Azure think of as an API - a related set of resources) rather it's per-type. This is basically what @porges was saying in the call Bevan/I with him earlier I think

hack/generator/spec/versioning.md
matthchr: See my larger example above -- every single type needs a `ConvertTo` I think?

hack/generator/spec/versioning.md
matthchr: Is there a win in having `v3` and `v3Storage` be different packages? How are they actually different? How is the generation of `ConvertTo`'s work when we come out with `v4` and are making `v4Storage` the storage shcema? It feels like we need a `v3` -> `v4Storage` and a `v3Storage` -> `v4Storage` and it's not clear to me how they will ever differ.

hack/generator/spec/versioning.md
davefellows: Can we fix the bug that causes properties to drop off? This seems like a really annoying thing to have to workaround... Additionally, how do we determine the difference between a property that's been mistakenly dropped (like `Locks`) vs a property that has legitimately been removed from the API?

Porges: (`Locks` is a whole resource type that was removed, not just a property)

Porges: The `Locks` resource that was removed from later API versions is still the correct one to use for subs & RGs. So we definitely need to preserve resource types that were “removed”.

hack/generator/spec/versioning.md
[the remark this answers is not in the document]
davefellows: Not sure I'm fully understanding the challenge. Might be good to discuss in this week's call.

hack/generator/spec/versioning.md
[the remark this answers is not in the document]
theunrepentantgeek: The proposal now recommends a rolling-storage version to avoid the unbounded accumulation of storage schema debt over time. It's more complex to set up initially, but doesn't degrade over time.

hack/generator/spec/versioning.md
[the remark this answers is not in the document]
theunrepentantgeek: I've completely reworked the lead-in to avoid digging too far into details too quickly. I've also pull out the examples into a pair of parallel case studies to allow comparison of the two major approaches.

hack/generator/spec/versioning.md
```diff
@@ -0,0 +1,201 @@
...
+
+If a resource type has been dropped from the ARM API, we will still generate a storage schema for it based on the last ARM API version where it existed; this ensures backward compatibility with existing service operator deployments.
+
+### Augment storage with a name/value property bag
+
+To facilitate easy and safe storage of additional configuration in storage, we'll include a name/value property bag for storage, along with helper methods that allow for strongly typed access to values stored within.
+
+Sequestering the information away within the storage schema is more robust than using separate annotations as they are less subject to arbitary modification by users. This allows us to largely avoid situations where well meaning (but uninformed) consumers of the service operator innocently make changes that result in the operator becoming failing. We particularly want to avoid this failure mode because recovery will be difficult - restoration of the modified/deleted information may be impractical or impossible.
```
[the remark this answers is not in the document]
theunrepentantgeek: Reworked this section.

hack/generator/spec/versioning.md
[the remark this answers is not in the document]
theunrepentantgeek: Yes, there will be a large number of conversion methods in flight.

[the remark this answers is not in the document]
theunrepentantgeek: Reworked this section.

hack/generator/spec/versioning.md
```diff
@@ -0,0 +1,201 @@
...
+
+
+### Define extension points
+
+Optional interfaces will be generated to allow injection of additional steps to the generated conversions. If implemented on the spoke versions (not the hub version), they will be automatically invoked as required.
+
+For each actual API version, an interface will be generated containing strongly typed `AssignTo()` and `AssignFrom()` methods:
+
+``` go
+type ClusterPropertiesv20160301Converter {
```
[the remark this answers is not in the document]
theunrepentantgeek: Fixed.

hack/generator/spec/versioning.md
theunrepentantgeek: Yes, every custom type needs its own `ConvertTo/From` methods generated.

hack/generator/spec/versioning.md
theunrepentantgeek: I think I've covered this off in the new spec, let me know if it reads clearly.

hack/generator/spec/versioning.md
```diff
@@ -0,0 +1,192 @@
...
+
+We can also anticipate changes going in the other direction, where a strongly typed value becomes a **string** in order to provide greater flexibility.
+
+#### Discussion
+
+Do we define the type using `interface{}` to allow any value to be stored, but at the cost of type safety? What does this do to our CRD generation based on these types?
+
+Can we reuse the work being done by @mattchr to handle OneOf schema types?
+
+We need to ensure that the final solution is backward compatible when we already have copies of an object in storage and the type is changed in a new version. For example, if we already had a `Meeting` in storage and the new version made the type change detailed above, we can't retrospectively change the name of the properties to avoid name clashes.
```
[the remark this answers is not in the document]
theunrepentantgeek: I've completely reworked this section.

hack/generator/spec/versioning.md
theunrepentantgeek: Yes, we will do this - and I've explicitly called it out in the spec.

hack/generator/spec/versioning.md
theunrepentantgeek: Reworked this section.

hack/generator/spec/versioning.md
theunrepentantgeek: Reworked the section on renaming.

hack/generator/spec/versioning.md
```diff
@@ -0,0 +1,273 @@
...
+
+***TBC***
+
+## Case Studies
+
+There are two case studies that accompany this specification, each one walking through one possible solution and showing how it will perform over time.
+
+The [**Rolling Versions**](case-study-rolling-storage-versions.md) case study shows how the preferred solution adapts to changes as the Azure Resources evolve over time.
+
+The [**Fixed Version**](case-study-rolling-storage-versions.md) case study shows how the primary alternative would fare, calling out some specific problems that will occur.
```
davefellows, line 37: This link is wrong - should be to the fixed version md.

hack/generator/spec/case-study-rolling-storage-versions.md
```diff
@@ -0,0 +1,934 @@
+# Case Study - Rolling Storage Versions
+
+This case study explores the recommended solution of using a *rolling storage version* where we update the storage schema of each resource each release of the service operator. We'll keep the storage version up to date with the latest GA release of each resource.
+
+For the purposes of discussion, we'll be following the version by version evolution of a theoretical ARM service that provides customer resource management (CRM) services. Synthetic examples are used to allow focus on specific scenarios one by one, providing motivation for specific features.
+
+Examples shown are deliberately simplified in order to focus, and therefore minutate should be considered motivational, not binding. Reference the formal specification for precise details.
```
davefellows, line 7: minutiae?

hack/generator/spec/case-study-rolling-storage-versions.md
```diff
@@ -0,0 +1,934 @@
...
+
+    return nil
+}
+
+// ConvertFrom converts from the Hub version to this version.
+func (person *Person) ConvertFromStorage(source storage.Person) error {
+    person.AlphaKey = source.AlphaKey
+    person.FamilyName = source.FamilyName
+    person.Id = source.Id
+    person.KnownAs = source.KnownAs"
```
davefellows, line 383: erroneous character
```suggestion
    person.KnownAs = source.KnownAs
```

hack/generator/spec/case-study-rolling-storage-versions.md
```diff
@@ -0,0 +1,934 @@
...
+
+Conversions with the upgraded storage version will be identical (except for the import statements for referenced types) to those shown above.
+
+## Version Map
+
+Our hub and spoke diagram is becoming useful for seeing the relationship between versions:
+
+![](images/case-study-rolling-storage-2012-02-02.png)
+
+Observe that the prior storage version is still shown, with a one way conversion to the current storage version. Existing users who upgrade their service operator will have their storage upgraded using this conversion. The conversion between storege versions will be generated with the same approach, and with the same structure, as all our other conversions.
```
davefellows, line 122:
```suggestion
Observe that the prior storage version is still shown, with a one way conversion to the current storage version. Existing users who upgrade their service operator will have their storage upgraded using this conversion. The conversion between storage versions will be generated with the same approach, and with the same structure, as all our other conversions.
```

hack/generator/spec/case-study-rolling-storage-versions.md
```diff
@@ -0,0 +1,934 @@
...
+
+    return nil
+}
+
+// ConvertFrom converts from the Hub version to this version.
+func (person *Person) ConvertFromStorage(source storage.Person) error {
+    person.AlphaKey = source.SortKey // *** Rename is automatically handled ***
+    person.FamilyName = source.FamilyName
+    person.Id = source.Id
+    person.KnownAs = source.KnownAs"
```
davefellows, line 553:
```suggestion
    person.KnownAs = source.KnownAs
```

hack/generator/spec/case-study-rolling-storage-versions.md
```diff
@@ -0,0 +1,934 @@
...
+
+These changes are entirely similar to those previously covered in version 2014-04-04, above.
+
+## Version Map
+
+In this release, we see that support for both `2014-04-04` and the preview version `2014-04-04preview` has been dropped:
+
+![](images/case-study-rolling-storage-2018-08-08.png)
+
+Users still running earlier releases of the service operator that are using `2014-04-04` or earlier will need to install an intermediate release in order to upgrade to this one.
```
davefellows, line 840: This could be a poor experience for people who find themselves in this situation. We should think about ways we can mitigate or simplify this (or retain all conversions).

hack/generator/spec/case-study-rolling-storage-versions.md
```diff
@@ -0,0 +1,934 @@
...
+    Longitide    double
+}
+```
+
+The storage version gets changed in a straightforward way:
+
+``` go
+package v20190909storage 
+
+type Location struct {
```
davefellows, line 864: two things;
* shouldn't this include Lattitude and Longitude?
* are properties meant to be alphabetically ordered on the generated storage struct as well as the conversions?

hack/generator/spec/case-study-rolling-storage-versions.md
```diff
@@ -0,0 +1,934 @@
...
+``` go
+package v20190909
+
+type Location struct {
+    FullAddress  string
+    City         string
+    Country      string
+    PostCode     string
+    Lattitude    double
+    Longitide    double
```
davefellows, line 855:
```suggestion
    Longitude    double
```

hack/generator/spec/case-study-fixed-storage-version.md
```diff
@@ -0,0 +1,1105 @@
+# Case Study - Fixed Storage Version
+
+This case study explores the alternative solution of using a *fixed storage version* where the schema of the storage version is modified to handle each additional release.
+
+For the purposes of discussion, we'll be following the version by version evolution of a theoretical ARM service that provides customer resource management (CRM) services. Synthetic examples are used to allow focus on specific scenarios one by one, providing motivation for specific features.
+
+Examples shown are deliberately simplified in order to focus on specific details, and therefore minutate should be considered motivational, not binding. Reference the formal specification for precise details.
```
davefellows, line 7:
```suggestion
Examples shown are deliberately simplified in order to focus on specific details, and therefore minutiae should be considered motivational, not binding. Reference the formal specification for precise details.
```

hack/generator/spec/case-study-fixed-storage-version.md
```diff
@@ -0,0 +1,1105 @@
...
+
+    return nil
+}
+```
+
+This provides round-trip support for the preview release, but does not provide backward compatibility with prior official releases. 
+
+The storage version of `Person` written by the preview release will have no values for `FirstName`, `LastName`, and `MiddleName`.
+
+Similarly, the storage version of `Person` written by an eariler release will have no values for `KnownAs`, `FamilyName`, or `FullName`.
```
davefellows, line 265:
```suggestion
Similarly, the storage version of `Person` written by an eariler release will have no values for `KnownAs`, `FamilyName`, or `FullName`.
```
```suggestion
Similarly, the storage version of `Person` written by an earlier release will have no values for `KnownAs`, `FamilyName`, or `FullName`.
```

hack/generator/spec/case-study-fixed-storage-version.md
```diff
@@ -0,0 +1,1105 @@
...
+
+These kinds of cross-version conversions cannot be automatically generated as they require more understanding the semantic changes between versions.
```
davefellows, line 267:
```suggestion
These kinds of cross-version conversions cannot be automatically generated as they require more understanding of the semantic changes between versions. 
```

hack/generator/spec/case-study-fixed-storage-version.md
```diff
@@ -0,0 +1,1105 @@
...
+
+## Issue: Property Bloat
+
+As our API evolves over time, our storage version is accumulating all the properties that have ever existed, bloating the storage version with obsolete properties that are seldom (if ever) used. 
+
+For example, only preview users would ever have used `FullName` as it became `LegalName` in the general release; most users will never use `FullName`.
+
+We have a problem, however. We can't remove the `FullName` property as that is a breaking change that will negatively impact users who *have* used the preview version. Both properties will need to be retained permanently.
+
+This violates the *pay for play* principle - even users who adopt the operator after the 2014-04-04 release will have to deal with the complexity even though they've never used the preview version.
```
davefellows, line 375: The *users* wouldn't have to deal with any additional complexity, would they?

hack/generator/spec/case-study-fixed-storage-version.md
```diff
@@ -0,0 +1,1105 @@
...
+
+References to `FirstName`, `MiddleName` and `LastName` will disappear across all the generated `ConvertFromStorage()` and `ConvertToStorage()` methods as well.
+
+**This is a breaking change.** Existing long time users of the service operator, with CRDs serialized containing any of the properties `FirstName`, `MiddleName` and `LastName` will find that their clusters break when this version of the service operator is deployed.
+
+Without explicit intervention in the operator, their only mitigation would be to revert back to the earlier version of the service operator and *never upgrade it ever again.*
+
+The root cause of the problem is that the CRD has never been modified (upgraded) - once first deployed the resource is never modified, only loaded and reconciled.
+
+To retain backward compatibility we would need manually merge the generated code with prior versions to retain both the property definitions and the conversion support previously generated.
```
davefellows, line 887:
```suggestion
To retain backward compatibility we would need to manually merge the generated code with prior versions to retain both the property definitions and the conversion support previously generated. 
```

hack/generator/spec/versioning.md
```diff
@@ -0,0 +1,201 @@
...
+
+To facilitate easy and safe storage of additional configuration in storage, we'll include a name/value property bag for storage, along with helper methods that allow for strongly typed access to values stored within.
+
+Sequestering the information away within the storage schema is more robust than using separate annotations as they are less subject to arbitary modification by users. This allows us to largely avoid situations where well meaning (but uninformed) consumers of the service operator innocently make changes that result in the operator becoming failing. We particularly want to avoid this failure mode because recovery will be difficult - restoration of the modified/deleted information may be impractical or impossible.
+
+---
+
+****Outstanding issue:** Where do the helper functions for the name/value property bag reside?**
+
+We could generate those anew for each type, making them conveniently available and avoiding the need for our generated types to depend on anything more than the standard Go library Alternatively, we could define the property bag as a new struct, wrapping up storage & helper methods into a reusable form.
```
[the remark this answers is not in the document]
theunrepentantgeek: Agreed - and it gives us a convenient place for supporting methods too.

hack/generator/spec/case-study-rolling-storage-versions.md
theunrepentantgeek: Damn. I thought I'd fixed all of those!

hack/generator/spec/case-study-rolling-storage-versions.md
theunrepentantgeek: Agreed - it comes down to answering the question "How long do we retain older storage versions?". We would want at least a year, as a lower bound. We might want to consider keeping them indefinitely - though there's not a zero cost to doing so.

hack/generator/spec/case-study-rolling-storage-versions.md
theunrepentantgeek: Yes to both.

hack/generator/spec/versioning.md
theunrepentantgeek: Fixed.

hack/generator/spec/case-study-rolling-storage-versions.md
theunrepentantgeek: Fixed. #hopefullyeverywhere

hack/generator/spec/case-study-rolling-storage-versions.md
theunrepentantgeek: Fixed.

hack/generator/spec/case-study-rolling-storage-versions.md
theunrepentantgeek: Fixed.

hack/generator/spec/case-study-rolling-storage-versions.md
theunrepentantgeek: Fixed.

hack/generator/spec/case-study-rolling-storage-versions.md
theunrepentantgeek: Fixed.

hack/generator/spec/case-study-fixed-storage-version.md
theunrepentantgeek: Fixed

hack/generator/spec/case-study-fixed-storage-version.md
theunrepentantgeek: Fixed.

hack/generator/spec/case-study-fixed-storage-version.md
theunrepentantgeek: Added.

hack/generator/spec/case-study-fixed-storage-version.md
theunrepentantgeek: Added.

hack/generator/spec/case-study-rolling-storage-versions.md
```diff
@@ -0,0 +1,936 @@
...
+With these methods available, implementing the interface `AssignableToPerson` becomes straightforward. For the `2011-01-01` release of `Person`:
+
+``` go
+package v20110101
+
+import storage "v20140404storage"
+
+func (person *Person) AssignTo(dest storage.Person) error {
+    dest.PopulateFromFirstMiddleLastName(
+        person.FirstName, "", person.LastName)
```
matthchr, line 466: minor: missing return nil which you have below

hack/generator/spec/case-study-fixed-storage-version.md
theunrepentantgeek: On the happy path, when everything works, no. In that case, nobody cares about the format.

But when something goes awry and is investigating issues, they may choose (or we may ask them) to directly inspect what's in etcd to work out what's going on; the more complex the information that's returned, the harder it becomes to troubleshoot the issues.

hack/generator/spec/versioning.md
```diff
@@ -0,0 +1,273 @@
...
+
+**Consistency of experience:** Early adopters should have a similar experience with the latest release of the service operator as new users who are adopting it for the first time. We don't want early adopters to be penalized for their enthusiasm. 
+
+## Non-Goals
+
+**Coverage of every case by code generation:** While it's likely that very high coverage will be achievable with code generation, we don't believe that it will be practical to handle every possible situation automatically. It's therefore necessary for the solution to have some form of extensibility allowing for the injection of hand written code.
+
+## Requirements
+
+***TBC***
```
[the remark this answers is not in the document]
theunrepentantgeek: Good idea.

hack/generator/spec/versioning.md
```diff
@@ -0,0 +1,273 @@
...
+Sequestering additional properties away within a property bag in the storage schema is more robust than using separate annotations as they are less subject to arbitary modification by users. This allows us to largely avoid situations where well meaning (but uninformed) consumers of the service operator innocently make changes that result in the operator becoming failing. We particularly want to avoid this failure mode because recovery will be difficult - restoration of the modified/deleted information may be impractical or impossible.
+
+### Generated conversion methods
+
+Each of the structs generated for ARM API will have the normal `ConvertTo()` and `ConvertFrom()` methods generated automatically, implementing the required [Convertible](https://book.kubebuilder.io/multiversion-tutorial/conversion.html) interface:
+
+``` go
+// ConvertTo converts this Person to the Hub storage version.
+func (person *Person) ConvertTo(raw conversion.Hub) error {
+    p := raw.(*storage.Person)
```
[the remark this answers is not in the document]
theunrepentantgeek: The examples I've seen are all using the hard cast as shown here; I believe the contract with the `Convertible` interface is that it won't be called with the wrong type.

hack/generator/spec/case-study-rolling-storage-versions.md
```diff
@@ -0,0 +1,936 @@
...
+We need to implement the [Convertible](https://book.kubebuilder.io/multiversion-tutorial/conversion.html) interface to allow conversion to and from the storage version:
+
+``` go
+package v20110101
+
+import storage "v20110101storage"
+
+// ConvertTo converts this Person to the Hub storage version.
+func (person *Person) ConvertTo(raw conversion.Hub) error {
+    p := raw.(*storage.Person)
```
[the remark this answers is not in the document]
theunrepentantgeek: I believe the contract for `Convertible` ensures this will never panic.

hack/generator/spec/versioning.md
```diff
@@ -0,0 +1,273 @@
...
+Can we find examples? Do we want to support these cases?
+
+### Standard extension points
+
+Code generation will include interfaces to allow easy injection of manual conversion steps. 
+
+For each storage type, two interfaces will be generated, one to be called by `ConvertToStorage()` and one to be called by `ConvertFromStorage()`:
+
+``` go
+type AssignableToPerson interface {
```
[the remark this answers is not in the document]
theunrepentantgeek: We did have a conversation - there was some confusion over what the methods actually did (e.g. were they additive to the process, or replacements for generated steps), which lead to some suggestions for renaming. I thought the place we ended up was to review the names once people were more familiar with the spec.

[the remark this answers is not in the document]
matthchr: Ah yeah, I had forgotten that's where we landed (and forgotten what I suggested as alternatives too!). You can just resolve this for now I think.

hack/generator/spec/versioning.md
theunrepentantgeek: There are two reasons why an old API version might drop out:

* By our policy, that we decide we're only supporting the 3 latest GA versions. We could decide to support _all_ of them, but that makes our support load greater.

* By ARM policy, when the older ARM JSON schema is no longer available because ARM are not publishing it. In this case we don't have a choice.

hack/generator/spec/versioning.md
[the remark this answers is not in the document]
theunrepentantgeek: When a type is removed, the latest version of that type will be used to define the storage version. Just as we do for other types, we convert to the storage type for storage - we'll put it into the PropertyBag as an `interface{}` and rely on the way it's serialized by the bag.

hack/generator/spec/case-study-rolling-storage-versions.md
theunrepentantgeek: Capture as a separate issue for later discussion; outside of the scope of this spec.

hack/generator/spec/case-study-rolling-storage-versions.md
```diff
@@ -0,0 +1,936 @@
...
+
+Conversions with the upgraded storage version will be identical (except for the import statements for referenced types) to those shown above.
+
+## Version Map
+
+Our hub and spoke diagram is becoming useful for seeing the relationship between versions:
+
+![](images/case-study-rolling-storage-2012-02-02.png)
+
+Observe that the prior storage version is still shown, with a one way conversion to the current storage version. Existing users who upgrade their service operator will have their storage upgraded using this conversion. The conversion between storage versions will be generated with the same approach, and with the same structure, as all our other conversions.
```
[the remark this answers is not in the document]
theunrepentantgeek: Nothing explicit - will capture in a related issues section.

theunrepentantgeek:
Abandoning thjs PR in favour of starting a fresh conversation with a wider audience.

PR #94, opened 2020-05-18
Run make all and commit formatted output

matthchr:


devigned:
Let use branches on main, rather than forks until we get the lower privileged builds sorted. I'll create an issue and assign myself. Probably able to get it done today.

PR #98, opened 2020-05-18
Generate type for oneOf instead of interface{}

matthchr:
- This generates a new type in the recommended
    Kubernetes discriminated union style (N properties, each
    mutually exclusive).


matthchr:
This is a draft until the other PR merges and then I'll retarget this to master

hack/generator/pkg/astmodel/definition_name.go
```diff
@@ -29,6 +29,8 @@ func (dn *DefinitionName) Name() string {
 // it is simply a reference to the name.
 var _ Type = (*DefinitionName)(nil)
```
theunrepentantgeek, line 30: We may be able to simplify `DefinitionName` by no longer implementing everything needed by the interface `Type` - would be worth exploring.

hack/generator/pkg/astmodel/identifier_factory.go
```diff
@@ -18,6 +18,8 @@ var filterRegex = regexp.MustCompile(`[\W_]`)
 type IdentifierFactory interface {
 	CreateIdentifier(name string) string
 	CreateFieldName(fieldName string) FieldName
+	CreatePrivateIdentifier(name string) string // TODO: Name?
+	CreatePrivateFieldName(fieldName string) FieldName
```
theunrepentantgeek, line 22: Should these be separate methods, or should we pass an enum requesting specific visibility:

``` go
CreateFieldName(fieldName string, visibility Visibility) FieldName
```
where 
``` go
type Visibility string

const (
    Public = Visibility("public")
    Internal = Visibility("internal")
)
```

hack/generator/pkg/jsonast/jsonast.go
```diff
@@ -496,8 +496,64 @@ func oneOfHandler(ctx context.Context, scanner *SchemaScanner, schema *gojsonsch
...
+				return nil, fmt.Errorf("couldn't find struct for definition: %v", concreteType)
+			}
+			fieldName := scanner.idFactory.CreateFieldName(concreteType.Name())
+
+			// TODO: JSON name is unimportant here because we will implement the JSON marshaller anyway
+			jsonName := scanner.idFactory.CreatePrivateIdentifier(concreteType.Name())
+			field := astmodel.NewFieldDefinition(fieldName, jsonName, concreteType).MakeOptional()
+			fields = append(fields, field)
+		case *astmodel.EnumType:
+			// Just a sanity check that we've already scanned this definition
```
theunrepentantgeek, line 523: Comment is misplaced

hack/generator/pkg/jsonast/jsonast.go
```diff
@@ -496,8 +496,64 @@ func oneOfHandler(ctx context.Context, scanner *SchemaScanner, schema *gojsonsch
...
+			name := fmt.Sprintf("object%v", i)
+			fieldName := scanner.idFactory.CreateFieldName(name)
+
+			// TODO: JSON name is unimportant here because we will implement the JSON marshaller anyway
+			jsonName := scanner.idFactory.CreatePrivateIdentifier(name)
+			field := astmodel.NewFieldDefinition(fieldName, jsonName, concreteType).MakeOptional()
+			fields = append(fields, field)
+		case *astmodel.PrimitiveType:
+			// TODO: This name sucks but what alternative do we have?
+			name := fmt.Sprintf("primitive%v", i)
```
theunrepentantgeek, line 541: Can you use the name of the primitive type to make the name less sucky - `string4` `int3` etc?

matthchr: Yeah, can do this. None of the options are really mind-bogglingly good, but I've added this at least until we come up with something better.

matthchr:
This needs more tests, but I need to sync with @Porges on the right way to write the golden file tests.

hack/generator/pkg/astmodel/identifier_factory.go
```diff
@@ -39,21 +46,37 @@ func NewIdentifierFactory() IdentifierFactory {
...
+	result := strings.Title(clean)
+
+	if visibility == Internal {
+		// TODO: This is a hack, as there are cases (acronyms, etc) where
+		// TODO: this doesn't work...
+		// Lowercase the first rune if visibility is internal
+		done := false
+		result = strings.Map(
+			func(r rune) rune {
+				if !done {
```
davefellows, line 66: nit: minor: personal preferance I guess but I prefer early exit:
`if done {
    return r
}
done = true
return unicode.ToLower(r)`

hack/generator/pkg/astmodel/identifier_factory.go
```diff
@@ -39,21 +46,37 @@ func NewIdentifierFactory() IdentifierFactory {
...
+					done = true
+					return unicode.ToLower(r)
+				}
+				return r
```
davefellows, line 70: nit: minor: personal preferance I guess but I prefer early exit:
```suggestion
if done {
    return r
}
done = true
return unicode.ToLower(r)
```

hack/generator/pkg/jsonast/jsonast.go
```diff
@@ -180,13 +180,28 @@ func (scanner *SchemaScanner) GenerateDefinitions(ctx context.Context, schema *g
...
 
 	// TODO: make safer:
-	root := astmodel.NewStructDefinition(rootStructRef, nodes.(*astmodel.StructType))
+	var root *astmodel.StructDefinition
+	switch n := nodes.(type) {
+	case *astmodel.DefinitionName:
+		// A root definition name can happen for example in the case of a oneOf of refs which were all deduped.
+		// When this happens we just need to look up the corresponding struct type
+		if def, ok := scanner.FindDefinition(*n); ok {
+			root = def.(*astmodel.StructDefinition) // TODO: is this safe?
```
theunrepentantgeek, line 195: If it's not, this will panic - should you perhaps use the `root, ok := ...` form and return an error if not?

hack/generator/pkg/jsonast/jsonast.go
```diff
@@ -655,3 +736,19 @@ func versionOf(url *url.URL) (string, error) {
 	// No version found, that's fine
 	return "", nil
 }
+
+func appendIfUniqueType(slice []astmodel.Type, item astmodel.Type) []astmodel.Type {
```
theunrepentantgeek, line 740: There's a standard Go idiom where you could use `map[Type]bool` as a set; would that make things easier/better?

hack/generator/pkg/jsonast/jsonast.go
```diff
@@ -492,16 +507,82 @@ func oneOfHandler(ctx context.Context, scanner *SchemaScanner, schema *gojsonsch
...
 	}
 
-	// bail out, can't handle this yet:
-	return astmodel.AnyType, nil
+	// If there's more than one option, synthesize a type.
+	// Note that this is required because Kubernetes CRDs do not support OneOf the same way
+	// OpenAPI does, see https://github.com/Azure/k8s-infra/issues/71
+	var fields []*astmodel.FieldDefinition
+
+	for i, t := range results {
```
theunrepentantgeek, line 523: A thought on naming - the types have their own names (e.g. `Foo`, `Bar` and `Baz` from your samples); can we use those to generate names for the fields?

hack/generator/pkg/astmodel/one_of_json_marshal_function.go
```diff
@@ -0,0 +1,132 @@
...
+				},
+			},
+		}
+
+		statements = append(statements, &ifStatement)
+	}
+
+	finalReturnStatement := &ast.ReturnStmt{
+		Results: []ast.Expr{
+			ast.NewIdent("nil, nil"), // TODO: Is this the right way to do this?
```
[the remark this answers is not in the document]
Porges: @theunrepentantgeek can you link that playground? Sounds useful.

hack/generator/pkg/jsonast/jsonast.go
Porges: It’s actually `map[Type]struct{}`; my version isn’t as simple 😆

hack/generator/pkg/astmodel/one_of_json_marshal_function.go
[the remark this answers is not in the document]
theunrepentantgeek: Not a Go Playground link, but I created a [gist](https://gist.github.com/theunrepentantgeek/c1cfbd8044bf273da61660d01995545c)

hack/generator/pkg/jsonast/jsonast.go
```diff
@@ -180,13 +180,28 @@ func (scanner *SchemaScanner) GenerateDefinitions(ctx context.Context, schema *g
...
-		scanner.idFactory.CreateIdentifier(rootStructName),
+		scanner.idFactory.CreateIdentifier(rootStructName, astmodel.Public),
 		scanner.idFactory.CreateGroupName(rootStructGroup),
 		scanner.idFactory.CreatePackageNameFromVersion(rootStructVersion),
 		false)
 
 	// TODO: make safer:
-	root := astmodel.NewStructDefinition(rootStructRef, nodes.(*astmodel.StructType))
+	var root *astmodel.StructDefinition
+	switch n := nodes.(type) {
```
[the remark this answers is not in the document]
Porges: this changed on master

hack/generator/pkg/astmodel/identifier_factory.go
```diff
@@ -39,21 +46,37 @@ func NewIdentifierFactory() IdentifierFactory {
 }
 
 // CreateIdentifier returns a valid Go public identifier
-func (factory *identifierFactory) CreateIdentifier(name string) string {
+func (factory *identifierFactory) CreateIdentifier(name string, visibility Visibility) string {
 	if identifier, ok := factory.renames[name]; ok {
-		return identifier
+		name = identifier
```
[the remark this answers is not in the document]
matthchr: Is the cache really a cache?

Right now the only thing we have in there is `"$schema"` -> `"Schema"` - the issue with returning based on this is that it assumes the cache has knowledge about the desired case. It matters if I want "Schema" or "schema" as the result, and that will depend on if I am looking for a private identifier or a public one?

Maybe I am saying the same as you (the cache is broken as it doesn't take visibility into account) - this change is making it take visibility into account (because now visibility always applies after the renames take place).

hack/generator/pkg/astmodel/identifier_factory.go
```diff
@@ -39,21 +46,37 @@ func NewIdentifierFactory() IdentifierFactory {
...
+		done := false
+		result = strings.Map(
+			func(r rune) rune {
+				if !done {
+					done = true
+					return unicode.ToLower(r)
+				}
+				return r
+			},
+			result)
```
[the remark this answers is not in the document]
matthchr: Yeah

hack/generator/pkg/astmodel/one_of_json_marshal_function.go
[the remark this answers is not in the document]
matthchr: Fixed this - Thanks for the link too Bevan.

hack/generator/pkg/jsonast/jsonast.go
matthchr: The problem is we want the results to be strictly ordered according to the order they were defined in the oneOf (to aid with deterministic code generation). They also need to be indexed because we sometimes use the index to help generate a unique property name. I think the combination of these two things means that using a map is a bit awkward as we lose our ordering guarantee.

hack/generator/pkg/jsonast/jsonast.go
matthchr: That's what the first case (`TypeName`) is already doing -- using the name of the ref to generate the property. All of the other cases (`EnumType`/`StructType`, etc) are for anonymous types embedded in the OneOf (see the `OneOf_generates_wrapper_type_with_named_properties_from_anonymous_types` golden file test) - there are not really any "good" names to use in this case unfortunately.

hack/generator/pkg/astmodel/one_of_json_marshal_function.go
```diff
@@ -0,0 +1,132 @@
...
+								},
+								Args: []ast.Expr{
+									fieldSelectorExpr,
+								},
+							},
+						},
+					},
+				},
+			},
+		}
```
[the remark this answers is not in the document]
matthchr: That's correct.

hack/generator/Makefile
```diff
@@ -17,11 +17,11 @@ build: tidy fmt vet lint ; $(info $(M) buiding ./bin/$(APP))
 
 .PHONY: lint
 lint: $(GOLANGCI_LINT) ; $(info $(M) running golangci configured linters…) ## Lint codebase
-	$(Q) $(GOLANGCI_LINT) run -v --timeout 5m
+	$(Q) $(GOLANGCI_LINT) run -v
```
theunrepentantgeek, line 20: Is there a timeout on the CI pipeline that will catch long running execution anyway?

hack/generator/pkg/astmodel/identifier_factory.go
```diff
@@ -11,12 +11,20 @@ import (
...
+// \W is all non-word characters (https://golang.org/pkg/regexp/syntax/)
 var filterRegex = regexp.MustCompile(`[\W_]`)
 
+type Visibility string
+
+const (
+	Public   = Visibility("public")
+	Internal = Visibility("internal")
+)
+
```
theunrepentantgeek, line 23: I'm conscious that _internal_ is really the .NET view on what to call that level of visibility; what's the Go nomenclature? perhaps _package_?

hack/generator/pkg/astmodel/identifier_factory.go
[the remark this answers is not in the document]
theunrepentantgeek: You're right, it's not really a cache - it's a dictionary of hand crafted transformations, to handle cases where the normal rules don't give the right result.

So the rules we're applying below shouldn't be applied if the dictionary is used, except for the case conversion to handle visibility.

hack/generator/pkg/astmodel/identifier_factory.go
```diff
@@ -39,21 +47,31 @@ func NewIdentifierFactory() IdentifierFactory {
...
+	for i, word := range cleanWords {
+		if visibility == Internal && i == 0 {
+			caseCorrectedWords = append(caseCorrectedWords, strings.ToLower(word))
+		} else {
+			caseCorrectedWords = append(caseCorrectedWords, strings.Title(word))
+		}
+	}
+
+	result := strings.Join(caseCorrectedWords, "")
+	result = strings.ReplaceAll(result, " ", "")
```
theunrepentantgeek, line 69: Do we need this? Under what circumstances would the `Join()` give us a string containing spaces?

hack/generator/pkg/astmodel/identifier_factory.go
```diff
@@ -128,16 +146,23 @@ func simplifyName(context string, name string) string {
...
 		preceedingLower := i > 0 && unicode.IsLower(chars[i-1])
 		succeedingLower := i+1 < len(chars) && unicode.IsLower(chars[i+1])
+		isSpace := unicode.IsSpace(chars[i])
 		foundUpper := unicode.IsUpper(chars[i])
 		if i > lastStart && foundUpper && (preceedingLower || succeedingLower) {
 			result = append(result, string(chars[lastStart:i]))
 			lastStart = i
+		} else if i > lastStart && isSpace {
+			result = append(result, string(chars[lastStart:i]))
+			lastStart = i + 1 // skip the space
```
theunrepentantgeek, line 165: Do any of the test cases  cover this branch?

hack/generator/pkg/astmodel/identifier_factory_test.go
```diff
@@ -79,6 +90,8 @@ func Test_TransformToSnakeCase_ReturnsExpectedString(t *testing.T) {
 		// Correctly splits all-caps acronyms
 		{string: "XMLDocument", expected: "xml_document"},
 		{string: "ResultAsXML", expected: "result_as_xml"},
+		{string: "AlreadyHas spaces", expected: "already_has_spaces"},
+		{string: "AlreadyHas spaces    ", expected: "already_has_spaces"},
```
theunrepentantgeek, line 94: This set of test cases needs a comment (they're not checking all-caps acronyms)

hack/generator/pkg/jsonast/testdata/OneOf/OneOf_generates_wrapper_type_with_marshal_helper.golden
```diff
@@ -0,0 +1,47 @@
...
+}
+
+/*Generated from: https://test.test/schemas/2020-01-01/test.json*/
+type Test struct {
+	Bar *Bar `json:"bar"`
+	Baz *Baz `json:"baz"`
+	Foo *Foo `json:"foo"`
+}
+
+// MarshalJSON marshals the object as JSON
```
theunrepentantgeek, line 35: Is there anything more meaningful we can include in this doc comment?

hack/generator/pkg/jsonast/jsonast.go
theunrepentantgeek: Ok, fair enough.

hack/generator/Makefile
matthchr: It's set in the config file I just forgot to remove it here

hack/generator/pkg/astmodel/identifier_factory.go
matthchr: Likely never - can remove

hack/generator/pkg/astmodel/identifier_factory.go
devigned: exported / not exported, perhaps

hack/generator/pkg/astmodel/identifier_factory.go
matthchr: Yes, I added a few that do

hack/generator/pkg/jsonast/jsonast.go
matthchr: It would be nice if maps worked like Python dicts (by default, they iterate through in insertion order) - unfortunately though they don't behave like that

hack/generator/pkg/jsonast/testdata/OneOf/OneOf_generates_wrapper_type_with_marshal_helper.golden
matthchr: Possibly - I realize this is a bit of a cop-out comment but unsure what else to say really.
We could make the documentation specific to the implementation by saying: `MarshalJSON defers marshaling to the first non-nil property to simulate OneOf`?

Do you have any suggestions?

hack/generator/pkg/astmodel/identifier_factory.go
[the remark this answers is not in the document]
matthchr: Fixed

hack/generator/pkg/jsonast/testdata/OneOf/OneOf_generates_wrapper_type_with_marshal_helper.golden
matthchr: `MarshalJSON defers JSON marshaling to the first non-nil property, because TestProperties represents a discriminated union (JSON OneOf)`

It's an improvement I think.

PR #101, opened 2020-05-19
Ensure related definitions are always created

theunrepentantgeek:
Having the `HasRelatedDefinitions` interface optional meant that container types (such as Array and Map) could easily neglect implementation, resulting in definitions being omitted from the generated code.

This PR makes it mandatory for both definitions and types to implement the interface by including it as a part of both `Definition` and `Type`. 

As a necessary part of this, `EnumDefinition` is no longer compatible with `EnumType` - this wasn't actually needed and becomes impossible with the change described above (any struct trying to implement both `Definition` and `Type` experiences a compiler error).

### DIFF
--- a/hack/generator/pkg/astmodel/enum_definition.go
+++ b/hack/generator/pkg/astmodel/enum_definition.go
@@ -19,6 +19,7 @@ type EnumDefinition struct {
 
 var _ Definition = (*EnumDefinition)(nil)
 
+// NewEnumDefinition is a factory method for creating new Enum Definitions
 func NewEnumDefinition(name DefinitionName, t *EnumType) *EnumDefinition {
 	return &EnumDefinition{DefinitionName: name, baseType: t}
 }
--- a/hack/generator/pkg/astmodel/enum_definition.go
+++ b/hack/generator/pkg/astmodel/enum_definition.go
@@ -13,10 +13,15 @@ import (
 
 // EnumDefinition generates the full definition of an enumeration
 type EnumDefinition struct {
-	EnumType
+	DefinitionName
+	baseType *EnumType
 }
 
 var _ Definition = (*EnumDefinition)(nil)
+
+func NewEnumDefinition(name DefinitionName, t *EnumType) *EnumDefinition {
+	return &EnumDefinition{DefinitionName: name, baseType: t}
+}
 
 // FileNameHint returns a desired name for this enum if it goes into a standalone file
 func (enum *EnumDefinition) FileNameHint() string {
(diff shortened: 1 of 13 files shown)


hack/generator/pkg/astmodel/map_type.go
```diff
@@ -62,19 +62,12 @@ func (m *MapType) Equals(t Type) bool {
...
-	if df, ok := m.key.(HasRelatedDefinitions); ok {
-		defns := df.RelatedDefinitions(ref, namehint, idFactory)
-		result = append(result, defns...)
-	}
-
-	if df, ok := m.value.(HasRelatedDefinitions); ok {
-		defns := df.RelatedDefinitions(ref, namehint, idFactory)
-		result = append(result, defns...)
-	}
+	otherdefinitions := m.value.CreateRelatedDefinitions(ref, namehint, idFactory)
```
matthchr, line 69: otherDefinitions?

hack/generator/pkg/astmodel/enum_definition.go
```diff
@@ -104,3 +110,14 @@ func (enum *EnumDefinition) createValueDeclaration(value EnumValue) ast.Spec {
 
 	return valueSpec
 }
+
+// FindRelatedDefinitions returns a set of definitions related to this one
```
matthchr, line 114: I am not seeing anywhere that this method is actually called...?

matthchr: Also since `EnumDefinition` implements `Definition`, and `Definition` `HasRelatedDefinitions`, shouldn't enum have a `CreateRelatedDefinitions`? I'm not seeing it.

matthchr: Oh, but `DefinitionName` is embedded and `DefinitionName` has an implementation of `CreateRelatedDefinitions` already, which we are getting because of struct embedding...?

But do we _really want_ that definition? It feels like we don't, we want our own definition here that farms out to `baseType` like `FindRelatedDefinitions` does below?

Which is maybe a long-winded way of saying, do we really want to embed `DefinitionName` here, or do we just want a property of type `DefinitionName`, but not embedded?

matthchr:
Is there a way we can add a test which ensures that this case is fixed?

hack/generator/pkg/astmodel/enum_definition.go
theunrepentantgeek: Ooo, good catch - this is debris from when I tried something else; the method is needed, but it has the wrong name, wrong type, and wrong implementation. 😜 

I think I'll just replace it with the right thing.

theunrepentantgeek: Done.

hack/generator/pkg/astmodel/map_type.go
theunrepentantgeek: Done.

hack/generator/pkg/astmodel/enum_value.go
```diff
@@ -0,0 +1,23 @@
...
+}
+
+// Equals tests to see if the passed EnumValue has the same name and value
+func (value *EnumValue) Equals(v *EnumValue) bool {
+	if value == v {
+		return true
+	}
+
+	return value.Identifier == v.Identifier && value.Value == v.Value
+}
```
matthchr, line 23: lacking newline?

PR #102, opened 2020-05-19
Use klog for logging and tame the volume of output 

theunrepentantgeek:
Uses the `klog` for logging instead of the `log` package, including integration with command line options to allow configuration of the verbosity of output.

Also extracts the top level functionality out of the `gen.go` file, allowing it concentrate on command line options.

Will need to be retargeted to `master` after #101 merges.

``` text
PS> .\generator.exe --help
Usage of D:\dev\go\src\github.com\k8s-infra\hack\generator\generator.exe:
  -add_dir_header
        If true, adds the file directory to the header of the log messages
  -alsologtostderr
        log to standard error as well as files
  -log_backtrace_at value
        when logging hits line file:N, emit a stack trace
  -log_dir string
        If non-empty, write log files in this directory
  -log_file string
        If non-empty, use this log file
  -log_file_max_size uint
        Defines the maximum size a log file can grow to. Unit is megabytes. If the value is 0, the maximum file size is unlimited. (default 1800)
  -logtostderr
        log to standard error instead of files (default true)
  -skip_headers
        If true, avoid header prefixes in the log messages
  -skip_log_headers
        If true, avoid headers when opening log files
  -stderrthreshold value
        logs at or above this threshold go to stderr (default 2)
  -v value
        number for the log level verbosity
  -vmodule value
        comma-separated list of pattern=N settings for file-filtered logging
```


hack/generator/cmd/gen/gen.go
```diff
@@ -29,103 +29,18 @@ func NewGenCommand() (*cobra.Command, error) {
...
-				log.Printf("Configuration invalid: %v\n", err)
-				return err
-			}
-
-			log.Printf("Loading schema %s", configuration.SchemaURL)
-			schema, err := loadSchema(configuration.SchemaURL)
-			if err != nil {
-				log.Printf("Failed to load schema: %v\n", err)
+			err = cg.Generate(ctx, "apis")
+			if err!=nil {
```
matthchr, line 39: minor: formatting, I believe you need spaces between the `err!=nil` bits -- CI will likely complain about this now but may not have before

hack/generator/pkg/astmodel/package_definition.go
```diff
@@ -44,13 +44,20 @@ func (pkgDef *PackageDefinition) EmitDefinitions(outputDir string) {
 	allocateTypesToFiles(otherDefinitions, filesToGenerate)
 	emitFiles(filesToGenerate, outputDir)
 	emitGroupVersionFile(pkgDef, outputDir)
+
+	return len(filesToGenerate)
```
matthchr, line 48: You may get a conflict here with my PR where I added some error handling/reporting here.

hack/generator/pkg/astmodel/package_definition.go
```diff
@@ -44,13 +44,20 @@ func (pkgDef *PackageDefinition) EmitDefinitions(outputDir string) {
...
+func (pkgDef *PackageDefinition) DefinitionCount() int {
+	return len(pkgDef.definitions)
 }
 
 func emitFiles(filesToGenerate map[string][]Definition, outputDir string) {
 	for fileName, defs := range filesToGenerate {
 		genFile := NewFileDefinition(defs[0].Reference().PackageReference, defs...)
 		outputFile := filepath.Join(outputDir, fileName+"_types.go")
-		log.Printf("Writing '%s'\n", outputFile)
+		klog.V(5).Infof("Writing '%s'\n", outputFile)
```
matthchr, line 60: Are there an unbounded number of verbosity levels? How far up do we want to use in code?

hack/generator/pkg/codegen/configuration.go
```diff
@@ -29,17 +30,17 @@ const (
 	Skip ShouldExportResult = "skip"
 )
 
-// NewExportConfiguration is a convenience factory for ExportConfiguration
-func NewExportConfiguration(filters ...*TypeFilter) *ExportConfiguration {
-	result := ExportConfiguration{
+// NewExportConfiguration is a convenience factory for Configuration
+func NewExportConfiguration(filters ...*jsonast.TypeFilter) *Configuration {
```
matthchr, line 34: NewConfiguration?

hack/generator/pkg/astmodel/package_definition.go
theunrepentantgeek: Yup. Merged.

hack/generator/pkg/astmodel/package_definition.go
theunrepentantgeek: It goes up to 5 and there are defined uses, see https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md

PR #104, opened 2020-05-19
Update References usage

matthchr:
- References only actually operates on DefinitionNames so make that
    clear.


Porges:
Nice 👍 I think this makes it a lot clearer, and it’s such a small change.

PR #105, opened 2020-05-19
Add codeowners

matthchr:
I think this should auto-add reviewers which is cool (I'm tired of clicking)


matthchr:
@devigned I am not sure if you're the only actual "owner" at the top level or if there are others so I added you and you can add whoever else you like.

devigned:
Note here the drop in coverage. That is due to integration tests adding to the code coverage and this PR coming from a fork only running the unit tests (no secrets).

/cc @theunrepentantgeek @Porges

PR #106, opened 2020-05-21
Remove Tidy() and just use immutability and ordering on creation

matthchr:


hack/generator/pkg/astmodel/enum_type.go
```diff
@@ -22,7 +25,11 @@ var _ Type = (*EnumType)(nil)
 
 // NewEnumType defines a new enumeration including the legal values
 func NewEnumType(baseType *PrimitiveType, options []EnumValue) *EnumType {
-	return &EnumType{BaseType: baseType, Options: options}
+	sort.Slice(options, func(left int, right int) bool {
+		return options[left].Identifier < options[right].Identifier
+	})
```
theunrepentantgeek, line 30: Sorting on creation is useful - means we can rely on `Equals()` without having to worry about things being out of order.

matthchr: Yep!

Porges:
I like this a lot, nice!

PR #107, opened 2020-05-21
Simplify & Improve Definition/Type interfaces

Porges:
Summary:
* `controller-gen` can now run without crashing
* all nested enums and structs are pulled out and named
* we generate names for everything that is named in a schema, not just structs/enums
* nomenclature should be a bit clearer
----
Details;
* `DefinitionName` renamed to `TypeName`
* `Definition` interface renamed to `TypeDefiner` (golangy!) and simplified, it has the `TypeName` of the type it is defining and the `Type` that it will be bound to. `AsDeclarations` is on this type.
  - Add `SimpleTypeDefiner` for when we want to give names to simple types that aren't structs/enums. This is functionality we didn't have before (see next point as well).
* `CreateRelatedDefinitions` is renamed and split/massaged to `CreateDefinitions` which will create both the primary definition and any associated definitions. 
This latter thing fixes a few issues:
  - previously we only named structs & enums. now a `refHandler` can give a name to anything like a map, etc (this is actually used a lot).
  - removes the special-casing of structs that we had in `refHandler` and at the root node
  - removes the mutation of `canonicalName` in `EnumType`
  - this is also necessary to enforce that internal structs are always named, which is needed to actually run controller-gen properly.
* There is also `CreateInternalDefinitions` which walks the innards of a type and creates any enums/structs that must be named in order to be used. With this change `controller-gen` can now generate CRDs without crashing.


hack/generator/pkg/astmodel/struct_type.go
```diff
@@ -110,14 +109,42 @@ func (structType *StructType) Equals(t Type) bool {
...
+	ref := NewStructReference(name.name, name.groupName, name.packageName, isResource)
+
+	var otherTypes []TypeDefiner
+	var newFields []*FieldDefinition
+
+	for _, field := range st.fields {
+		newField := field
+
+		// TODO: figure out a generic way to do this:
+		if et, ok := newField.FieldType().(*EnumType); ok {
```
theunrepentantgeek, line 123: 💥 
This will only work for enums directly used by fields. If the enum is wrapped by `MapType`, `ArrayType` or `OptionalType` it won't get a name. The previous implementation used recursion to handle this - and it broke when `OptionalType` was introduced.

Porges: right, that's the Outstanding bit that I need to reintroduce

hack/generator/pkg/astmodel/simple_type_definer.go
```diff
@@ -0,0 +1,42 @@
...
+// SimpleTypeDefiner is a TypeDefiner for simple cases (not structs or enums)
+type SimpleTypeDefiner struct {
+	name    *TypeName
+	theType Type
+}
+
+// SimpleTypeDefiner is a TypeDefiner
+var _ TypeDefiner = (*SimpleTypeDefiner)(nil)
+
+func (gtd *SimpleTypeDefiner) Name() *TypeName {
```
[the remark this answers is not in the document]
Porges: I renamed the type from GenericTypeDefiner, missed the receivers

hack/generator/pkg/astmodel/definition.go
```diff
@@ -5,58 +5,45 @@
...
-	// AsType renders the current instance as a Go abstract syntax tree
+	// AsType renders as a Go abstract syntax tree for a type
+	// (yes this says ast.Expr but that is what the Go 'ast' package uses for types)
 	AsType() ast.Expr
 
 	// Equals returns true if the passed type is the same as this one, false otherwise
 	Equals(t Type) bool
+
+	// CreateDefinitions gives a name to the type and might generate some asssociated definitions as well (the second result)
+	CreateDefinitions(name *TypeName, idFactory IdentifierFactory) (TypeDefiner, []TypeDefiner)
```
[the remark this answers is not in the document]
Porges: one is the actual type being defined, so I thought it was more obvious to make that clear rather than return a `[]TypeDefiner` which we then have to slice into parts

hack/generator/pkg/astmodel/enum_definition.go
```diff
@@ -69,7 +64,7 @@ func (enum *EnumDefinition) Tidy() {
 
 func (enum *EnumDefinition) createBaseDeclaration() ast.Decl {
 	var identifier *ast.Ident
-	identifier = ast.NewIdent(enum.name)
+	identifier = ast.NewIdent(enum.name.name)
```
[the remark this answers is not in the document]
Porges: that's better

hack/generator/pkg/astmodel/enum_type.go
```diff
@@ -27,20 +25,14 @@ func NewEnumType(baseType *PrimitiveType, options []EnumValue) *EnumType {
...
-}
-
-// CreateRelatedDefinitions returns a definition for our enumeration, with a name based on the referencing property
-func (enum *EnumType) CreateRelatedDefinitions(ref PackageReference, namehint string, idFactory IdentifierFactory) []Definition {
-	identifier := idFactory.CreateEnumIdentifier(namehint)
-	enum.canonicalName = DefinitionName{PackageReference: ref, name: identifier}
-	definition := NewEnumDefinition(enum.canonicalName, enum)
-	return []Definition{definition}
+func (enum *EnumType) References(d *TypeName) bool {
+	return false
```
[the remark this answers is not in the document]
Porges: Yeah, it will never reference anything. But perhaps for consistency...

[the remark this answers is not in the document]
Porges: ... done.

hack/generator/pkg/astmodel/map_type.go
```diff
@@ -62,6 +62,7 @@ func (m *MapType) Equals(t Type) bool {
 	return false
 }
 
+/*j
```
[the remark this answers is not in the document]
Porges: associated with TODO

hack/generator/pkg/astmodel/package_reference.go
```diff
@@ -14,6 +14,10 @@ type PackageReference struct {
 	packageName string
 }
 
+func NewPackageReference(groupName string, packageName string) PackageReference {
```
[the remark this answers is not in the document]
Porges: We use `PackageReference` as a value everywhere so I didn't want this to return a pointer. I think I'll rename the function.

hack/generator/pkg/astmodel/enum_type.go
```diff
@@ -34,20 +32,14 @@ func NewEnumType(baseType *PrimitiveType, options []EnumValue) *EnumType {
 
 // AsType implements Type for EnumType
 func (enum *EnumType) AsType() ast.Expr {
-	return ast.NewIdent(enum.canonicalName.name)
+	// TODO: should warn here, emitting a non-named enum
+	// when this is feature-complete this should "never" happen
+	return enum.BaseType.AsType()
```
theunrepentantgeek, line 39: Won't this just be emitting `string` or `int` - is this correct?

hack/generator/pkg/astmodel/identifier_factory.go
```diff
@@ -88,6 +88,19 @@ func sanitizePackageName(input string) string {
 	return string(builder)
 }
 
+// transformToSnakeCase transforms a string LikeThis to a snake-case string like_this
+func transformToSnakeCase(input string) string {
+	words := sliceIntoWords(input)
+
+	// my kingdom for LINQ
```
theunrepentantgeek, line 95: 😁

hack/generator/pkg/astmodel/identifier_factory_test.go
```diff
@@ -64,6 +64,34 @@ func Test_SliceIntoWords_GivenIdentifier_ReturnsExpectedSlice(t *testing.T) {
...
+		// Single name doesn't get split
+		{string: "Name", expected: "name"},
+		// Single Acronym doesn't get split
+		{string: "XML", expected: "xml"},
+		// Splits simple words
+		{string: "PascalCase", expected: "pascal_case"},
+		{string: "XmlDocument", expected: "xml_document"},
+		// Correctly splits all-caps acronyms
+		{string: "XMLDocument", expected: "xml_document"},
+		{string: "ResultAsXML", expected: "result_as_xml"},
```
theunrepentantgeek, line 81: Nice.

hack/generator/pkg/astmodel/map_type.go
```diff
@@ -62,12 +62,12 @@ func (m *MapType) Equals(t Type) bool {
...
 
-// CreateRelatedDefinitions returns any definitions required by our key or value types
-func (m *MapType) CreateRelatedDefinitions(ref PackageReference, namehint string, idFactory IdentifierFactory) []Definition {
-	result := m.key.CreateRelatedDefinitions(ref, namehint, idFactory)
-
-	otherDefinitions := m.value.CreateRelatedDefinitions(ref, namehint, idFactory)
-	result = append(result, otherDefinitions...)
+func (m *MapType) CreateInternalDefinitions(name *TypeName, idFactory IdentifierFactory) (Type, []TypeDefiner) {
+	newKeyType, otherTypes := m.key.CreateInternalDefinitions(name, idFactory)
+	newValueType, otherTypes2 := m.value.CreateInternalDefinitions(name, idFactory)
```
theunrepentantgeek, line 67: Suggestions for variable naming ...
```
keyType, otherKeyTypes := ...
valueType, otherValueTypes := ...
```

hack/generator/pkg/astmodel/package_reference_test.go
```diff
@@ -6,14 +6,13 @@
 package astmodel
 
 import (
-	"github.com/onsi/gomega"
 	"testing"
+
+	"github.com/onsi/gomega"
```
theunrepentantgeek, line 11: Isn't this supposed to be brought into scope?
```
. "github.com/onsi/gomega"
```

Then you can write
```
g := NewGomegaWithT(t)
```

hack/generator/pkg/astmodel/simple_type_definer.go
```diff
@@ -0,0 +1,65 @@
...
+	return &result
+}
+
+func (std *SimpleTypeDefiner) AsDeclarations() []ast.Decl {
+	var docComments *ast.CommentGroup
+	if std.description != nil {
+		docComments = &ast.CommentGroup{
+			List: []*ast.Comment{
+				{
+					Text: "\n/*" + *std.description + "*/",
```
theunrepentantgeek, line 47: Leading and trailing spaces need to be a part of the comment.

hack/generator/pkg/astmodel/struct_definition.go
```diff
@@ -109,15 +82,15 @@ func (definition *StructDefinition) AsDeclarations() []ast.Decl {
 		},
 	}
 
-	if definition.description != "" {
+	if definition.description != nil {
 		declaration.Doc.List = append(declaration.Doc.List,
-			&ast.Comment{Text: "\n/* " + definition.description + " */"})
+			&ast.Comment{Text: "\n/*" + *definition.description + "*/"})
```
theunrepentantgeek, line 87: Needs spaces around the content of the description

Porges:
> Broadly makes sense to me, though I'm unsure how it works now. #shrug

Two main loci of changes:

1) Enabling `controller-gen` by naming structs nested inside other types:

Where we used to have some special code in `jsonast` naming the `StructType` and then calling `CreateRelatedDefinitions` we now have a function `CreateDefinitions` which names the type and returns any extra definitions like what `CreateRelatedDefinitions` did (so `CreateDefinitions` is basically naming with `CreateRelatedDefinitions` attached). This means `refHandler` can now grant a name to any type.

In addition to that, `CreateDefinitions` names everything inside that type that has to be named for `controller-gen` to work (i.e. enums/structs), including types that are nested inside other types like maps/arrays/…;  `CreateInternalDefinitions` is the function that does this work (`CreateRelatedDefinitions` didn't do this before).

2) Clarification changes:

`TypeDefiner` takes over what `Definition` was doing before but it doesn’t participate in any of the `Type` stuff that was mixed into the `Definition` interface; I think this makes it cleaner. `DefinitionName` is a simple renaming of `TypeName` without much else changed. `StructReference` was removed entirely and the `isResource` bit moved into `StructDefinition`.

hack/generator/pkg/astmodel/identifier_factory.go
Porges: thanks @matthchr

hack/generator/pkg/astmodel/identifier_factory_test.go
Porges: again thanks to @matthchr

hack/generator/pkg/astmodel/package_reference_test.go
Porges: changed

hack/generator/pkg/astmodel/map_type.go
Porges: went with `keyOtherTypes` and `valueOtherTypes`

hack/generator/pkg/astmodel/enum_type.go
Porges: It should "never" happen as we should have named the enum and substituted-out the EnumType before this is called (so the `AsType` here won't be called). I added a warning to print out if it happens.

hack/generator/pkg/astmodel/enum_definition.go
```diff
@@ -80,9 +82,10 @@ func (enum *EnumDefinition) createBaseDeclaration() ast.Decl {
 
 func (enum *EnumDefinition) createValueDeclaration(value EnumValue) ast.Spec {
 
-	enumIdentifier := ast.NewIdent(enum.name)
+	enumIdentifier := ast.NewIdent(enum.typeName.name)
+
```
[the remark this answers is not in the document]
Porges: done

hack/generator/pkg/astmodel/array_type.go
```diff
@@ -52,7 +52,11 @@ func (array *ArrayType) Equals(t Type) bool {
 	return false
 }
 
-// CreateRelatedDefinitions returns any additional definitions that need to be created
-func (array *ArrayType) CreateRelatedDefinitions(ref PackageReference, namehint string, idFactory IdentifierFactory) []Definition {
-	return array.element.CreateRelatedDefinitions(ref, namehint, idFactory)
+func (array *ArrayType) CreateInternalDefinitions(name *TypeName, idFactory IdentifierFactory) (Type, []TypeDefiner) {
```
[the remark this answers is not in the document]
Porges: added doc-comments for all `Create(Internal)Definition` declarations

hack/generator/pkg/astmodel/simple_type_definer.go
Porges: At the moment this adversely affects CRD generation. Perhaps we can fix it in `controller-gen`.

hack/generator/pkg/astmodel/struct_definition.go
Porges: At the moment this adversely affects CRD generation. Perhaps we can fix it in `controller-gen`.

hack/generator/pkg/astmodel/simple_type_definer.go
theunrepentantgeek: Fair enough.

PR #110, opened 2020-05-21
Improve golden test formatting

Porges:
Make sure golden tests write all the generated definitions, and use the FileDefinition path so that they are formatted correctly.

I made the `CodeGenerator` less stateful by removing the `scanner` from it; it is no longer necessary to read the `Definitions` field from the `scanner` as `GenerateDefinitions` will return them directly.


hack/generator/cmd/gen/gen.go
```diff
@@ -58,55 +58,20 @@ func NewGenCommand() (*cobra.Command, error) {
...
-
-			err = os.Mkdir(rootOutputDir, 0700)
-			if err != nil {
-				log.Printf("Error %v\n", err)
 				return err
 			}
 
 			log.Printf("INF Checkpoint\n")
 
-			// group definitions by package
```
matthchr, line 84: Some of this may conflict with Bevan's PR #102

Porges: Rebased on that

PR #126, opened 2020-05-27
Use OneOf handler for AnyOf

matthchr:


matthchr:
I debated duplicating the OneOf tests for AnyOf but decided not to because it wasn't really going to add much and would introduce maintenance burden as we update the OneOf tests.

PR #127, opened 2020-05-27
TypeName references (in structs, maps, etc) must use package reference if not from the current package

matthchr:
This is for cases where one spec references another - the referenced specification is imported but the generated code was not actually using the module when defining structs, types, etc. Introduced concept of CodeGenerationContext (parameter to AsType/AsDeclarations) to inform various types as to the context they are in so that they can generate correct AST.

### DIFF
--- a/hack/generator/azure-cloud.yaml
+++ b/hack/generator/azure-cloud.yaml
@@ -6,3 +6,6 @@ typefilters:
   - action: exclude
     version: '*preview'
     because: preview SDK versions are excluded by default
+  - action: exclude
+    group: 'deploymenttemplate'
+    because: this is the "container" group that holds references to all of the other groups and as such doesn't make sense to generate
--- a/hack/generator/pkg/astmodel/code_generation_context.go
+++ b/hack/generator/pkg/astmodel/code_generation_context.go
@@ -10,21 +10,26 @@ package astmodel
 // in which the field declaration occurs - for example in a file with two conflicting package references
 // a disambiguation must occur and field types must ensure they correctly refer to the disambiguated types
 type CodeGenerationContext struct {
-	packageReferences map[PackageReference]struct{}
-	currentPackage    *PackageReference
+	packageImports map[PackageReference]PackageImport
+	currentPackage *PackageReference
 }
 
 // New CodeGenerationContext creates a new immutable code generation context
-func NewCodeGenerationContext(currentPackage *PackageReference, packageReferences map[PackageReference]struct{}) *CodeGenerationContext {
-	return &CodeGenerationContext{currentPackage: currentPackage, packageReferences: packageReferences}
+func NewCodeGenerationContext(currentPackage *PackageReference, packageImports map[PackageImport]struct{}) *CodeGenerationContext {
+	packageImportsMap := make(map[PackageReference]PackageImport)
+	for imp := range packageImports {
+		packageImportsMap[imp.PackageReference] = imp
+	}
+
+	return &CodeGenerationContext{currentPackage: currentPackage, packageImports: packageImportsMap}
 }
 
-// PackageReferences returns the set of package references in the current context
-func (codeGenContext *CodeGenerationContext) PackageReferences() map[PackageReference]struct{} {
+// PackageImports returns the set of package references in the current context
+func (codeGenContext *CodeGenerationContext) PackageImports() map[PackageReference]PackageImport {
 	// return a copy of the map to ensure immutability
-	result := make(map[PackageReference]struct{})
+	result := make(map[PackageReference]PackageImport)
 
-	for key, value := range codeGenContext.packageReferences {
+	for key, value := range codeGenContext.packageImports {
 		result[key] = value
 	}
 	return result
--- a/hack/generator/pkg/astmodel/code_generation_context.go
+++ b/hack/generator/pkg/astmodel/code_generation_context.go
@@ -14,8 +14,6 @@ type CodeGenerationContext struct {
 	currentPackage    *PackageReference
 }
 
-// TODO: How picky about immutability should we be -- should we put this in its own package?
-
 // New CodeGenerationContext creates a new immutable code generation context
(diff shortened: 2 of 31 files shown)


hack/generator/pkg/astmodel/code_generation_context.go
```diff
@@ -0,0 +1,33 @@
...
+// CodeGenerationContext stores context about the location code-generation is occurring.
+// This is required because some things (such as specific field types) are impacted by the context
+// in which the field declaration occurs - for example in a file with two conflicting package references
+// a disambiguation must occur and field types must ensure they correctly refer to the disambiguated types
+type CodeGenerationContext struct {
+	packageReferences map[PackageReference]struct{}
+	currentPackage    *PackageReference
+}
+
+// TODO: How picky about immutability should we be -- should we put this in its own package?
```
matthchr, line 17: Remove this before merge

hack/generator/pkg/astmodel/file_definition.go
```diff
@@ -37,37 +37,31 @@ func NewFileDefinition(packageRef PackageReference, definitions ...TypeDefiner)
...
 
-	for requiredImport := range requiredImports {
-		importSpecs = append(importSpecs, &ast.ImportSpec{
-			Name: nil,
-			Path: &ast.BasicLit{
-				Kind:  token.STRING,
-				Value: "\"" + requiredImport.PackagePath() + "\"",
-			},
-		})
+	// TODO: handle conflicting names
```
matthchr, line 57: Not sure if this actually happens now or not so I left it for a future item

hack/generator/pkg/astmodel/code_generation_context.go
```diff
@@ -0,0 +1,33 @@
...
+
+// New CodeGenerationContext creates a new immutable code generation context
+func NewCodeGenerationContext(currentPackage *PackageReference, packageReferences map[PackageReference]struct{}) *CodeGenerationContext {
```
matthchr, line 20: possibly break onto multiple lines

hack/generator/pkg/astmodel/package_reference.go
```diff
@@ -18,12 +20,26 @@ const (
 // a struct belongs to.
 type PackageReference struct {
 	packagePath string
+	name        *string
```
Porges, line 23: hmm, there's not much sense in having a `name` on a `PackageReference` when it is part of a `Type`. do we want to split this into `PackageName` (to use on a `Type`) and a `PackageImport` (= `PackageName` + friendly name) to be used for imports?

matthchr: Yeah, agree -- I thought about doing something like this but wasn't 100% sure. Can fix.

hack/generator/pkg/jsonast/jsonast.go
```diff
@@ -580,6 +584,7 @@ func oneOfHandler(ctx context.Context, scanner *SchemaScanner, schema *gojsonsch
 			// but we still need it for controller-gen
 			jsonName := scanner.idFactory.CreateIdentifier(name, astmodel.NotExported)
 			field := astmodel.NewFieldDefinition(fieldName, jsonName, concreteType).MakeOptional()
+			field = field.WithDescription(&fieldDescription)
```
theunrepentantgeek, line 587: Why not chain `.WithDescription()` after `.MakeOptional()`?

hack/generator/pkg/jsonast/testdata/OneOf/OneOf_generates_wrapper_type_with_marshal_helper.golden
```diff
@@ -27,8 +27,11 @@ type Foo struct {
 
 /*Generated from: https://test.test/schemas/2020-01-01/test.json*/
 type Test struct {
+	/* Bar: mutually exclusive with all other properties */
 	Bar *Bar `json:"bar"`
+	/* Baz: mutually exclusive with all other properties */
 	Baz *Baz `json:"baz"`
+	/* Foo: mutually exclusive with all other properties */
```
theunrepentantgeek, line 34: The other comments lack whitespace around the value, because of limitations in CRD generation, do these need to follow suit? e.g. `/*Bar: mutually exclusive with all other properties*/`

hack/generator/pkg/jsonast/jsonast.go
matthchr: Ah yes, now I can -- had structured it slightly differently originally so couldn't as easily. Will change

hack/generator/pkg/jsonast/testdata/OneOf/OneOf_generates_wrapper_type_with_marshal_helper.golden
matthchr: Fixed

hack/generator/pkg/astmodel/package_reference.go
matthchr: I left this one called PackageReference, and made a new one PackageImport.

hack/generator/pkg/astmodel/package_import.go
```diff
@@ -0,0 +1,62 @@
...
+		Name: name,
+		Path: &ast.BasicLit{
+			Kind:  token.STRING,
+			Value: "\"" + pi.PackageReference.PackagePath() + "\"",
+		},
+	}
+}
+
+// PackageName is the package name of the package reference
+func (pi *PackageImport) PackageName() string {
```
Porges, line 51: this works out nicely 👍

hack/generator/azure-cloud.yaml
```diff
@@ -6,3 +6,6 @@ typefilters:
   - action: exclude
     version: '*preview'
     because: preview SDK versions are excluded by default
+  - action: exclude
+    group: 'deploymenttemplate'
+    because: This is the "container" group that holds references to all of the other groups and as such doesn't make sense to generate
```
theunrepentantgeek, line 11: Given the way the message is output in logs, suggest you start it with lowercase: `this is the ...` #nit

PR #130, opened 2020-05-29
Add validation comment to enums

### LINKED ISSUE #119: Enums need a kubebuilder validation comment

matthchr:
This closes #119

### DIFF
--- a/hack/generator/pkg/jsonast/jsonast.go
+++ b/hack/generator/pkg/jsonast/jsonast.go
@@ -333,6 +333,10 @@ func getFields(ctx context.Context, scanner *SchemaScanner, schema *gojsonschema
 			fieldDefinition = fieldDefinition.MakeOptional()
 		}
 
+		if enum, ok := fieldDefinition.FieldType().(*astmodel.EnumType); ok {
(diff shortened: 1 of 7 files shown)


hack/generator/pkg/jsonast/jsonast.go
```diff
@@ -333,6 +333,16 @@ func getFields(ctx context.Context, scanner *SchemaScanner, schema *gojsonschema
 			fieldDefinition = fieldDefinition.MakeOptional()
 		}
 
+		// TODO: Should we be deferring this? To who? An enum used as a property is always supposed to have
+		// TODO: validation like this, but it doesn't necessarily make sense on the EnumType?
+		if enum, ok := fieldDefinition.FieldType().(*astmodel.EnumType); ok {
+			var values []interface{}
+			for _, opt := range enum.Options() {
+				values = append(values, opt.Value)
+			}
```
theunrepentantgeek, line 342: Perhaps turn this loop into a helper method on `astModel.EnumType`.

Porges:
One thing is that Kubebuilder allows validation comments on the `type` definition. Do we want to put it there instead of on each field using that `type`? It would stop us “down-casting” in `getFields` as well.

PR #134, opened 2020-06-04
Add support for type transformations, add additional kind of filter

matthchr:
There are 3 supported scenarios here:

1. `TypeFilter` 
    - `Include` - The default behavior.
    - `Exclude` - Prevent a type (struct) from being generated. This still allows other structs to reference this type (which will cause compilation issues down the road). This option is pretty "safe" because if you exclude a type you really actually wanted, you'll find out about it later when the code doesn't build.
2. `TypeFilter` - `Obliterate` (I am not sure the name is good...) - this totally removes the type and also will remove all references to it (properties won't get generated, oneOf/anyOf/allOf will get collapsed, etc). This is quite "dangerous" because you could accidentally remove properties that you weren't intending to. This is what we were doing before explicitly with `Expression`
3. `TypeTransformer` - This can instruct the generator to map a type to another type. It's particularly useful for types like `NumberOrExpression` where because of the `obliterate` of `Expression` above, that type is really just a `Float64`, but it doesn't really make sense to say: "Don't allow type alias's" generally, instead it's just that specific alias we want to remap to `float64`. It may have other uses like for manually crafting bits of the object hierarchy that aren't getting generated right (because we could type-transform the references of the bad type to our manually crafted one in a config).


matthchr:
I probably need to add a few tests to `type_filter_tests.go` for the new type of filter (although really there's pretty low risk it doesn't work given that it takes the same codepaths as the others...)

hack/generator/pkg/config/type_filter.go
```diff
@@ -19,8 +19,13 @@ type TypeFilterAction string
...
 	// IncludeType indicates that any type matched by the filter should be exported to disk by the generator
 	IncludeType TypeFilterAction = "include"
-	// ExcludeType indicates that any type matched by the filter should be skipped and not exported
+	// ExcludeType indicates that any type matched by the filter should not be exported as a struct.
+	// This skips generation of types but may not prevent references to the type by other structs (which
+	// will cause compiler errors).
 	ExcludeType TypeFilterAction = "exclude"
+	// ObliterateType indicates that any reference to the type should be entirely removed, this includes
+	// the type itself as well as properties referencing the type via a ref
+	ObliterateType TypeFilterAction = "obliterate" // TODO: Naming
```
matthchr, line 28: Better name suggestions accepted.

hack/generator/pkg/astmodel/package_reference.go
```diff
@@ -70,3 +70,11 @@ func (pr *PackageReference) PackageName() string {
...
 	return pr.packagePath == ref.packagePath
 }
+
+// String returns the string representation of the package reference
+func (pr *PackageReference) String() string {
+	return pr.packagePath
+}
+
+// Ensure we implement Stringer
+var _ fmt.Stringer = &PackageReference{}
```
babbageclunk, line 80: I've seen this more often as 
```suggestion
var _ fmt.Stringer = (*PackageReference)(nil)
```

... which avoids allocating. Obviously it isn't a huge deal since it's package level but can add up if you've got a big codebase!

Porges: we had been doing that but someone pointed out that uneffectful discards like this will be removed by the compiler. will send you the link

matthchr: Yeah the thing I liked about this format was that it works the same if the receiver is not a ptr (just drop the `&`).

hack/generator/pkg/codegen/code_generator.go
```diff
@@ -35,6 +36,14 @@ func NewCodeGenerator(configurationFile string) (*CodeGenerator, error) {
 		return nil, fmt.Errorf("configuration loaded from '%v' is invalid (%w)", configurationFile, err)
 	}
 
+	// TODO: Should there be a config.Init() rather than this?
```
theunrepentantgeek, line 39: Yeah, reaching inside another type to do this kind of thing is suboptimal.

matthchr:
Okay so I played with doing what @theunrepentantgeek suggested, and the main issue is that we rely on include/exclude happening at the end right now because with the current specification we want to exclude the root of the graph (deploymentTemplate). That works today because even though we're excluding it, we still walk the whole graph first, and only then exclude generating it at the end.

If I change things to where exclude == obliterate (graph pruning), we won't generate anything, because the first thing we run into is something we're excluding and so we will stop.

I can't think up a way that results in us only supporting the graph-pruning exclude with the current root we're using (because we want to prune it). I think we could do only graph pruning if we took in a list of configurations because then we don't want to prune the root (it just wouldn't be in the list), but as it stands we need both graph-pruning (for Expression) and exclude as it exists today (for deploymentTemplate).

I also am not sure that I agree with obliterate being implementation focused. There is (at least in my mind) a clear distinction between it and exclude:
Exclude: I want to prevent this type from being generated.
Obliterate: I want to prevent this type, and all types that are referenced only by it, from being generated.

It's the difference between dropping a single node from the graph (exclude) or dropping a sub-tree from the graph (obliterate).

*note that I am not advocating for the name obliterate because that really doesn't explain what it does very well…

hack/generator/pkg/config/export_filter.go
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+package config
+
+// ExportFilterAction defines the possible actions that should happen for types matching the filter
+type ExportFilterAction string
+
+const (
+	// ExportFilterActionInclude indicates that any type matched by the filter should be exported to disk by the generator
+	ExportFilterActionInclude ExportFilterAction = "include"
```
matthchr, line 13: The prefix here was because I was struggling with conflicting enums and this was _a way_ to differentiate different enums in the same package. If anybody has a better way I'm open to it

hack/generator/pkg/jsonast/jsonast.go
```diff
@@ -397,10 +387,29 @@ func refHandler(ctx context.Context, scanner *SchemaScanner, schema *gojsonschem
 			scanner.idFactory.CreatePackageNameFromVersion(version)),
 		scanner.idFactory.CreateIdentifier(name, astmodel.Exported))
 
+	shouldPrune, because := scanner.configuration.ShouldPrune(typeName)
+	// Only handle obliterate here because it's the only thing that modifies the type tree-walking
```
matthchr, line 391: Fix comment

hack/generator/pkg/config/export_filter.go
Porges: I'm not sure how golangy it is but you could make a very small package so that usage is like `exportfilter.Include`? @babbageclunk is that a done thing?

Porges:
Looking good :+

hack/generator/pkg/config/export_filter.go
matthchr: Yeah I'm not really sure what to do here -- totally open to ideas.

hack/generator/pkg/config/configuration.go
```diff
@@ -0,0 +1,124 @@
...
+	if config.SchemaURL == "" {
+		return errors.New("SchemaURL missing")
+	}
+
+	for _, transformer := range config.TypeTransformers {
+		err := transformer.Init()
+		if err != nil {
+			return err
+		}
+	}
```
theunrepentantgeek, line 68: From a user experience perspective, I'm wondering whether we should collate errors together and return all of them at once; playing whack-a-mole fixing one error at a time can get pretty tedious.

hack/generator/pkg/config/configuration.go
```diff
@@ -0,0 +1,124 @@
...
+
+	return nil
+}
+
+// ShouldExport tests for whether a given struct should be exported
```
theunrepentantgeek, line 73:
```suggestion
// ShouldExport tests for whether a given type should be exported as Go code
```

hack/generator/pkg/config/configuration.go
```diff
@@ -0,0 +1,124 @@
...
+				panic(fmt.Errorf("unknown exportfilter directive: %s", f.Action))
+			}
+		}
+	}
+
+	// By default we export all types
+	return Export, ""
+}
+
+// ShouldPrune tests for whether a given type should be processed or pruned
```
theunrepentantgeek, line 93:
```suggestion
// ShouldPrune tests for whether a given type should be extracted from the JSON schema or pruned
```

hack/generator/pkg/config/configuration.go
```diff
@@ -0,0 +1,124 @@
...
+				return Prune, f.Because
+			case TypeFilterIncludeType:
+				return Include, f.Because
+			default:
+				panic(fmt.Errorf("unknown typefilter directive: %s", f.Action))
+			}
+		}
+	}
+
+	// By default we export all types
```
theunrepentantgeek, line 108:
```suggestion
	// By default we include all types
```

hack/generator/pkg/config/configuration.go
```diff
@@ -0,0 +1,124 @@
...
+	return Include, ""
+}
+
+// TransformType uses the configured type transformers to transform a type name (reference) to a different type.
+// If no transformation is performed, nil is returned
```
theunrepentantgeek, line 113: Returning the original type if there is no transformation might make the consuming code simpler.

hack/generator/pkg/config/export_filter.go
theunrepentantgeek: I'm conflicted. I agree they should be separate enums (because they have different reasons to change), but the current naming is clunky. Moving to a different package wouldn't really help, unless you created several - one per enum, and that's even worse.

Delphi had a similar problem - unlike C# where enum values are namespaced within the type, Delphi enum values were global identifiers. The common idiom was to use a two letter, lowercase, prefix as a pseudo namespace.

Since filter actions are within-package, we could do the same ... so for ExportFilterAction, use `ea`:

``` go
eaInclude ExportFilterAction = "include"
eaExclude ExportFilterAction = "exclude"
```

and for `TypeFilterAction`, use `ta`:

``` go
taInclude TypeFilterAction = "include"
taPrune TypeFilterAction = "prune"
```

hack/generator/pkg/config/type_filter.go
```diff
@@ -0,0 +1,23 @@
...
+
+// TypeFilterAction defines the possible actions that should happen for types matching the filter
+type TypeFilterAction string
+
+const (
+	// TypeFilterIncludeType indicates that any type matched by the filter should be included in the type graph
+	TypeFilterIncludeType TypeFilterAction = "include"
+	// TypeFilterPruneType indicates that any type matched by the filter, and any types only referenced by that type
+	// should not be included in the type graph
+	TypeFilterPruneType TypeFilterAction = "prune"
```
theunrepentantgeek, line 16: If we keep the longer naming, should use `TypeFilterAction` as the prefix: `TypeFilterActionPrune` etc

hack/generator/pkg/config/type_transformer.go
```diff
@@ -0,0 +1,123 @@
...
+import (
+	"fmt"
+	"github.com/Azure/k8s-infra/hack/generator/pkg/astmodel"
+	"regexp"
+)
+
+type TransformTarget struct {
+	PackagePath string `yaml:",omitempty"`
+	Name        string `yaml:",omitempty"`
+}
```
theunrepentantgeek, line 17: Do these satisfy the "Is-A" guideline for when to embed vs name?

hack/generator/pkg/jsonast/jsonast.go
```diff
@@ -295,6 +290,15 @@ func generateFieldDefinition(ctx context.Context, scanner *SchemaScanner, prop *
 		return field, nil
 	}
 
+	// This can happen if the property type was prune away by a type filter.
```
theunrepentantgeek, line 293:
```suggestion
	// This can happen if the property type was pruned away by a type filter.
```

hack/generator/pkg/jsonast/jsonast.go
```diff
@@ -295,6 +290,15 @@ func generateFieldDefinition(ctx context.Context, scanner *SchemaScanner, prop *
...
+	// There are a few options here: We can skip this property entirely, we can emit it
+	// with no type (won't compile), or we can emit with with interface{}.
+	// TODO: Currently setting this to anyType as that's easiest to deal with and will generate
+	// TODO: a warning during controller-gen
+	if propType == nil {
```
theunrepentantgeek, line 298: I think a missing type should result in the property being omitted, with a log message at the [appropriate level](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md) ... maybe v2 or v3?

Emitting it with no type requires either manual fixes (yeuch!) or some other mechanism to individually suppress the affected properties. 

Using `interface{}` is problematic because controller gen doesn't really know what to do with it.

So I think suppressing the property, with visibility via logging, is the most viable approach.

hack/generator/pkg/config/type_transformer.go
theunrepentantgeek: I'm not sure why `TransformTarget` needs to be a nested struct, why not have optional fields directly in `TypeTransformer` to capture the changes?

hack/generator/pkg/config/type_transformer.go
```diff
@@ -0,0 +1,123 @@
...
+			transformer.Name)
+	}
+
+	transformer.groupRegex = createGlobbingRegex(transformer.Group)
+	transformer.versionRegex = createGlobbingRegex(transformer.Version)
+	transformer.nameRegex = createGlobbingRegex(transformer.Name)
+
+	// Type has no package -- must be a primitive type
+	if transformer.Transform.PackagePath == "" {
+		switch transformer.Transform.Name {
```
Porges, line 72: suggest extracting a `primitiveNameToType` (or something like that) function

Porges:
Still looking good 👍  :)

hack/generator/pkg/config/configuration.go
```diff
@@ -0,0 +1,124 @@
...
+
+const (
+	// Include indicates the specified type should be included in the type graph
+	Include ShouldPruneResult = "include"
+	// Prune indicates the type (and all types only referenced by it) should be pruned from the type graph
+	Prune ShouldPruneResult = "prune"
+)
+
+// NewConfiguration is a convenience factory for Configuration
+func NewConfiguration(filters ...*ExportFilter) *Configuration {
```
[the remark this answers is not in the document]
matthchr: Reworked to the `WithXxx` paradigm we have.

hack/generator/pkg/config/configuration.go
matthchr: Hmm okay, so I did this but of course the actual functionality to do this easily isn't built into the language so I added a dependency which supports it.

I'm not 100% sure that's ok so will need to check around about what the process is (i.e. license of dependency, etc...)

hack/generator/pkg/config/configuration.go
matthchr: I originally had actually done that but moved to this model instead because I felt like this was simpler - either way the caller needs to do a check (at least currently if we want logging), so the two options end up being:


```
	transformation, because := scanner.configuration.TransformType(typeName)
	if transformation != nil {
		klog.V(2).Infof("Transforming %s -> %s because %s", typeName, transformation, because)
		return transformation, nil
	}
```
vs
```
	transformation, because := scanner.configuration.TransformType(typeName)
	if transformation != typeName {
		klog.V(2).Infof("Transforming %s -> %s because %s", typeName, transformation, because)
		return transformation, nil
	}
```

I think you're thinking we could get to something like this:
```
	typeName, because := scanner.configuration.TransformType(typeName)
        // code that uses typename, no check required
```
but unfortunately the `typeName` is a more specific type than `ast.Type`, so the types don't work out (and that also makes the `transformation != typeName` check have more congnative load because you have do the "oh yea, typeName is a type too" song and dance). Given that the types don't work out to make it a one-liner, and we need to do a check for logging anyway, I think checking against nil (which is quite common/idiomatic in go) is clearer (or at least equally as clear) here.

hack/generator/pkg/config/export_filter.go
matthchr: horizontal space isn't really at a premium though, so why bother with the abbreviations at all?

`eaInclude` is significantly less clear (at the usage site) than `exportFilterActionInclude` (or `exportFilterInclude` if we prefer that), or `exportfilter.include` if we do the package-per-enum thing George is mentioning (which does suck because I think you can only have 1 package per directory so we'd have to pull it off into its own directory which I don't like).

hack/generator/pkg/config/configuration.go
matthchr: Actually, it might be possible to do as you say (or at least get closer to doing as you say) if `PrimitiveType` was just a `TypeName`. Part of the reason why the return of the transform function is `astmodel.Type` is because it can be a primitive type, OR a type name, and the only relationship between those two today is `astmode.Type`

If you squint and look at it, a `PrimitiveType` really is just a reference to a type defined by the standard library/language, so maybe we could do away with it entirely - will bring it up in teams chat where we can discuss it more.

hack/generator/pkg/jsonast/jsonast.go
matthchr: We're in a bit of a difficult situation with respect to logging here because a type `Person` can have a property of type `Hobby`, where `Hobby` is pruned, and then (later) the entire `Person` type can be excluded from emission.

The result is that... there really is no problem - we excluded the `Person` type from being emitted and `Hobby` was pruned as well. If we emit an error _right here_ though, we can't know if `Person` is going to be emitted or not. 

The reason I landed on `interface{}` was because:
1. It's a type filter not a property filter and I wasn't sure we wanted to conflate the two (note: There is no such thing as a property filter right now at least).
2. at controller-gen runtime, it sees what was actually emitted and so forcing a warning at that time solves the timing problem of the error.
3. controller-gen doesn't actually stop working if given a property with type `interface{}`, it just emits an error/warning and skips it.

hack/generator/pkg/config/type_transformer.go
matthchr: The reason I did it this way is so that the transformer was clearly:
```
// Properties to select types to apply to (what I am now calling TypeMatcher)

// A single property describing the type to transform to
```

I'm not really married to it though, we could do away with this entirely and promote it up a level

hack/generator/pkg/config/export_filter.go
babbageclunk: I think `exportFilterInclude` and `exportFilterExclude` are probably the best option here - `exportFilterActionInclude` feels like it crosses some psychological word-complexity boundary. It doesn't really merit a small package, given the whole having to put it in a separate directory thing. You could use an anonymous type like this:

```go
type ExportFilterAction string

var ExportFilter = struct {
	Include ExportFilterAction
	Exclude ExportFilterAction
}{
	Include: "include",
	Exclude: "exclude",
}
```

... and that's not *too* bad - it's ok to use at least. It does require having something named `ExportFilter` that isn't an export filter though. And they're not constants either, which is dicey.

babbageclunk: Oh, and then you've got `ExportFilter` just below. Anyway, I was just including that as a "you could do this but don't" :).

matthchr: I agree about the psychological word-complexity boundary for sure. I'll update the pattern for these two to be `exportFilterInclude`/`exportFilterExclude` and the equivalent for `typeFilter`.

hack/generator/pkg/config/type_filter.go
```diff
@@ -0,0 +1,23 @@
...
+ */
+
+package config
+
+// TypeFilterAction defines the possible actions that should happen for types matching the filter
+type TypeFilterAction string
+
+const (
+	// TypeFilterActionIncludeType indicates that any type matched by the filter should be included in the type graph
+	TypeFilterActionIncludeType TypeFilterAction = "include"
```
babbageclunk, line 13: And then these would be `TypeFilterInclude` and `TypeFilterPrune`, right?

hack/generator/pkg/config/configuration.go
matthchr: Poked around and confirmed that was okay

hack/generator/pkg/config/configuration.go
```diff
@@ -0,0 +1,145 @@
...
+ */
+
+package config
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/Azure/k8s-infra/hack/generator/pkg/astmodel"
+	"github.com/hashicorp/go-multierror" // TODO: Not sure what test I should be applying before importing a random package...
```
matthchr, line 13: Remove this

hack/generator/pkg/config/type_filter.go
matthchr: Yep! much cleaner

hack/generator/pkg/config/type_transformer.go
```diff
@@ -0,0 +1,99 @@
...
+type TransformTarget struct {
+	PackagePath string `yaml:",omitempty"`
+	Name        string `yaml:",omitempty"`
+}
+
+// A TypeTransformer is used to remap types
+type TypeTransformer struct {
+	TypeMatcher `yaml:",inline"`
+
+	Transform TransformTarget // TODO: Don't love this property name or type name...
```
babbageclunk, line 22: I'd use `Target` here - it's more specific than `Transform` (which is implied by the context anyway) and I think it reads better where it's used.

hack/generator/pkg/config/configuration.go
Porges: > If you squint and look at it, a `PrimitiveType` really is just a reference to a type defined by the standard library/language, so maybe we could do away with it entirely - will bring it up in teams chat where we can discuss it more.

Works for me 🙂

hack/generator/pkg/jsonast/jsonast.go
Porges: Perhaps we could attach warnings to types somehow (another map in scanner) and then only emit them if they don't get excluded later?

Porges: But for now it should be fine to just emit a warning...

babbageclunk: Yeah, I was going to say what @Porges said.

hack/generator/pkg/config/export_filter.go
```diff
@@ -0,0 +1,25 @@
...
+
+	// exportFilterExclude indicates that any type matched by the filter should not be exported as a struct.
+	// This skips generation of types but may not prevent references to the type by other structs (which
+	// will cause compiler errors).
+	exportFilterExclude exportFilterAction = "exclude"
+)
+
+// A ExportFilter is used to control which types should be exported by the generator
+type ExportFilter struct {
+	Action      exportFilterAction
```
babbageclunk, line 23: It seems a bit weird for the type and values for this field be unexported - how will client code construct `ExportFilter`s?

hack/generator/pkg/jsonast/jsonast.go
matthchr: Created an issue to track the overall improvement (#138). Will fix behavior here to emit warning and skip property for now.

hack/generator/pkg/config/export_filter.go
matthchr: Doh, it's only used in this package so I thought I could make it private, but realizing now that doesn't work because yaml would deserialize to it properly.

matthchr: Yeah, I was thinking that this is only used internally by `Configuration` (which is true), but it doesn't really make sense to make this internal without making a whole lot of other stuff internal too...

reverting it to be public like I had it before.

hack/generator/pkg/config/type_transformer.go
matthchr: Done

PR #135, opened 2020-06-05
Delete commented out jsonast tests

matthchr:


Porges:
Coveralls is doing some nonsense here...

matthchr:
@theunrepentantgeek - There's one test which we don't have coverage for already, which I have just added (it did actually find a bug too).

The others are already covered though (they're exercising `OneOf`/`AnyOf`/`AllOf` which we have test coverage for already)

PR #136, opened 2020-06-05
Improve file generation naming and deletion

### LINKED ISSUE #122: Use a specific naming convention to allow generated files to be separated from hand coded files
* Define the convention (this may be mandated by kubernetes or Go tooling)
* Ensure that only these files are deleted when we run generation

matthchr:
- There is a marker near the top of each code generated file stating
    that it was code generated by our generator.
  - When deleting files, we only delete those which were generated
    by our generator (based on the marking mentioned above)
  - This fixes #122


hack/generator/pkg/astmodel/file_definition.go
```diff
@@ -7,6 +7,7 @@ package astmodel
 
 import (
 	"bytes"
+	"github.com/Azure/k8s-infra/hack/generator/pkg/consts"
```
matthchr, line 10: Do we like this pattern?

hack/generator/pkg/astmodel/package_definition.go
```diff
@@ -33,6 +34,10 @@ func (pkgDef *PackageDefinition) AddDefinition(def TypeDefiner) {
 	pkgDef.definitions = append(pkgDef.definitions, def)
 }
 
+// TODO: I feel like we should move a lot of this "create the files" stuff out of package_definition because
```
matthchr, line 37: Thoughts on this?

hack/generator/pkg/astmodel/package_definition.go
```diff
@@ -234,7 +242,8 @@ func emitGroupVersionFile(pkgDef *PackageDefinition, outputDir string) error {
 		return err
 	}
 
-	gvFile := filepath.Join(outputDir, "groupversion_info.go")
+	// TODO: The name of this file doesn't matter, right?
```
matthchr, line 245: I dont think that it does give that controller-gen seems happy with it named with a prefix still, although we actually don't need the `zz_generated` prefix - do we want it or no?

hack/generator/pkg/codegen/code_generator.go
```diff
@@ -165,16 +171,119 @@ func loadSchema(source string) (*gojsonschema.Schema, error) {
...
+	}
+	err := filepath.Walk(path, walkFunction)
+	if err != nil {
+		return err
+	}
+
+	// Now order the directories by deepest first - we have to do this because otherwise a directory
+	// isn't empty because it has a bunch of empty directories inside of it
+	sortFunction := func(i int, j int) bool {
+		// TODO: Wow, somehow the slice I'm iterating over isn't a parameter to this func I have to capture it?
```
matthchr, line 263: Remove....

hack/generator/pkg/consts/generation_consts.go
```diff
@@ -0,0 +1,7 @@
+package consts
+
+const (
+	// Note that this format is actually an official specification in go:  https://github.com/golang/go/issues/13560
+	CodeGenerationComment = "Code generated by k8s-infra-gen. DO NOT EDIT."
+	CodeGeneratedFilePrefix = "zz_generated." // TODO: Technically this is not required now, could remove it?
```
matthchr, line 6: Do we like the zz_generated prefix or no?

hack/generator/pkg/astmodel/file_definition.go
theunrepentantgeek: Looks Ok to me. Needs a `// TODO` comment reminding us to update the path when we move the generator.

hack/generator/pkg/astmodel/package_definition.go
theunrepentantgeek: See my other comment. `zz_generated` should be bathed in high velocity hot plasma:

image

hack/generator/pkg/astmodel/package_definition.go
```diff
@@ -64,8 +69,9 @@ func (pkgDef *PackageDefinition) DefinitionCount() int {
 
 func emitFiles(filesToGenerate map[string][]TypeDefiner, outputDir string) error {
 	for fileName, defs := range filesToGenerate {
+		fullFileName := consts.CodeGeneratedFilePrefix + fileName + "_types.go"
 		genFile := NewFileDefinition(&defs[0].Name().PackageReference, defs...)
-		outputFile := filepath.Join(outputDir, fileName+"_types.go")
+		outputFile := filepath.Join(outputDir, fullFileName)
 		klog.V(5).Infof("Writing '%s'\n", outputFile)
 		err := genFile.SaveToFile(outputFile)
 		if err != nil {
```
davefellows, line 77: Minor: Some white space would be nice here. I found this quite hard to read.

hack/generator/pkg/astmodel/package_definition.go
davefellows: I definitely don't think we should copy that naming scheme. For one, I think it could confuse people into thinking those files had been generated by Kubebuilder. Secondly, it's pretty ghastly.

hack/generator/pkg/astmodel/package_definition.go
Porges: one thing is we would have to make the types have more robust public interfaces since we can't define 'member functions' in separate packages

hack/generator/pkg/jsonast/testdata/Validations/Enums_have_required_validation.golden
```diff
@@ -1,6 +1,6 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
-// This is a generated file. Please do not make manual changes.
+// Code generated by k8s-infra-gen. DO NOT EDIT.
```
Porges, line 3: ... or what? 🙂

hack/generator/pkg/consts/generation_consts.go
matthchr: We don't actually need any changes to the file naming (I had originally written things so that we did and then realized that go had an official "how to detect generated files" which is based on a comment in the file so I moved to use that.

I can just revert the `zz_generated` and leave things as is. We should correctly detect what files are generated by us vs other tools/handcrafted and it should in theory all play nicely together without any specific file naming convention at all.

hack/generator/pkg/codegen/code_generator.go
```diff
@@ -165,16 +171,119 @@ func loadSchema(source string) (*gojsonschema.Schema, error) {
...
+	sort.Slice(dirs, sortFunction)
+
+	// Now clean things up
+	for _, dir := range dirs {
+		files, err := ioutil.ReadDir(dir)
+		if err != nil {
+			return fmt.Errorf("error reading directory '%v' (%w)", dir, err)
+		}
+
+		if len(files) == 0 {
```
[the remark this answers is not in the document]
matthchr: It's actually _only_ folders, will improve the name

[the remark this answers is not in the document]
matthchr: Yes, it does

hack/generator/pkg/jsonast/testdata/Validations/Enums_have_required_validation.golden
matthchr: You will rue the day (or something).

This is just the standard format that the go standard indicates. (Matching the following pattern):
`^// Code generated .* DO NOT EDIT\.$`

hack/generator/pkg/astmodel/file_definition.go
matthchr: Added the todo.

We don't need another package - we could put this in `astmodel`. 
It's used in both  the `codegen` and `astmodel` packages currently because `codegen` performs the directory scan for deletion, and `astmodel` performs the generation of the files.

`codegen` has a reference to `astmodel` (which is why we could put this there).

I'll go ahead and move this for now (and remove this extra package) - I think it was created because I didn't like `astmodel` being the thing that defines file naming (because then `codegen needs to reach into astmodel to figure that out in order to do deletion), but we can solve that in another change.

hack/generator/pkg/consts/generation_consts.go
theunrepentantgeek: I do think there's value in having generated files detectable by filename - it's another clue to nudge future devs who might otherwise manually change the files.

I'm partial to `_gen.go` - it seems more in line with the way Go source files are named than `.gen.go`.

matthchr: Pushed an updated to move to `_gen.go` suffix

hack/generator/pkg/astmodel/package_definition.go
matthchr: I've removed this comment (for cleanliness), and will poke around with it locally.

hack/generator/pkg/codegen/code_generator.go
```diff
@@ -49,9 +55,9 @@ func (generator *CodeGenerator) Generate(ctx context.Context, outputFolder strin
 	}
 
 	klog.V(0).Infof("Cleaning output folder '%v'", outputFolder)
-	err = cleanFolder(outputFolder)
+	err = deleteGeneratedCodeFromFolder(outputFolder)
 	if err != nil {
-		return fmt.Errorf("error cleaning output folder '%v' (%w)", generator.configuration.SchemaURL, err)
+		return fmt.Errorf("error cleaning output folder '%v' (%w)", outputFolder, err)
```
theunrepentantgeek, line 60: Good catch!

hack/generator/pkg/codegen/code_generator.go
```diff
@@ -165,17 +171,118 @@ func loadSchema(source string) (*gojsonschema.Schema, error) {
 	return schema, nil
 }
 
-//TODO: Only clean generated files
-func cleanFolder(outputFolder string) error {
-	err := os.RemoveAll(outputFolder)
+func deleteGeneratedCodeFromFolder(outputFolder string) error {
+	globPattern := path.Join(outputFolder, "**", "*", "*"+astmodel.CodeGeneratedFileSuffix) + "*"
```
theunrepentantgeek, line 175: Should this finish with `.go`?

matthchr: Yeah, probably. I'll move it into the constant so that we don't repeat ourselves all over too.

PR #137, opened 2020-06-08
Use map for struct fields

theunrepentantgeek:
A minor change for consistency between `functions` and `fields`, treating both as maps.

* Simplifies equality comparison a little
* Improves symmetry between fields and functions
* Removes the wart that `NewStructType()` accepted fields but not functions

Grew out of a branch I have where I'm trying to simplify code for easier comprehension.

### DIFF
--- a/hack/generator/pkg/astmodel/struct_definition_test.go
+++ b/hack/generator/pkg/astmodel/struct_definition_test.go
@@ -22,7 +22,7 @@ func Test_NewStructDefinition_GivenValues_InitializesFields(t *testing.T) {
 	knownAsField := createStringField("knownAs", "Commonly known as")
 
 	ref := NewTypeName(*NewLocalPackageReference(group, version), name)
-	definition := NewStructDefinition(ref, NewStructType(fullNameField, familyNameField, knownAsField), false)
+	definition := NewStructDefinition(ref, NewStructType().WithFields(fullNameField, familyNameField, knownAsField), false)
 
 	definitionGroup, definitionPackage, err := definition.Name().PackageReference.GroupAndPackage()
 	g.Expect(err).ShouldNot(HaveOccurred())
--- a/hack/generator/pkg/astmodel/struct_definition_test.go
+++ b/hack/generator/pkg/astmodel/struct_definition_test.go
@@ -18,7 +18,7 @@ func Test_NewStructDefinition_GivenValues_InitializesFields(t *testing.T) {
 	const group = "group"
 	const version = "2020-01-01"
 	fullNameField := createStringField("fullName", "Full legal name")
-	familyNameField := createStringField("familiyName", "Shared family name")
+	familyNameField := createStringField("familyName", "Shared family name")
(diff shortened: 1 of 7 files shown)


hack/generator/pkg/astmodel/struct_definition_test.go
```diff
@@ -18,11 +18,11 @@ func Test_NewStructDefinition_GivenValues_InitializesFields(t *testing.T) {
 	const group = "group"
 	const version = "2020-01-01"
 	fullNameField := createStringField("fullName", "Full legal name")
-	familyNameField := createStringField("familiyName", "Shared family name")
+	familyNameField := createStringField("familyName", "Shared family name")
```
matthchr, line 21: hah, nice catch

hack/generator/pkg/astmodel/struct_definition_test.go
```diff
@@ -18,11 +18,11 @@ func Test_NewStructDefinition_GivenValues_InitializesFields(t *testing.T) {
...
 	knownAsField := createStringField("knownAs", "Commonly known as")
 
 	ref := NewTypeName(*NewLocalPackageReference(group, version), name)
-	definition := NewStructDefinition(ref, NewStructType(fullNameField, familyNameField, knownAsField), false)
+	definition := NewStructDefinition(ref, NewStructType().WithFields(fullNameField, familyNameField, knownAsField), false)
```
matthchr, line 25: minor/style: Wondering at what point it makes sense to break this on to multiple lines. Might not be there quite yet but seems to be getting close.

hack/generator/pkg/astmodel/struct_type.go
```diff
@@ -23,18 +23,26 @@ var EmptyStructType = NewStructType()
...
+func NewStructType() *StructType {
+	return &StructType{
+		fields:    make(map[FieldName]*FieldDefinition),
+		functions: make(map[string]Function),
+	}
 }
 
 // Fields returns all our field definitions
-// A copy of the slice is returned to preserve immutability
+// A sorted copy of the slice is returned to preserve immutability and provide determinism
```
matthchr, line 34: minor: Since there is no longer "a slice" by default, might be worth rephrasing this to not say: "a sorted copy of the slice"
```suggestion
// A sorted slice is returned to preserve immutability and provide determinism
```

hack/generator/pkg/astmodel/struct_type.go
```diff
@@ -98,19 +106,14 @@ func (structType *StructType) Equals(t Type) bool {
 			return false
 		}
 
-		ourFields := make(map[FieldName]*FieldDefinition)
```
matthchr, line 101: This cleaned up nicely.

hack/generator/pkg/astmodel/struct_definition_test.go
theunrepentantgeek: Every time I've tried to break Go code across multiple lines, the compiler barks at me. #compilersaysno

PR #145, opened 2020-06-12
Support cancellation via Ctrl-C

### LINKED ISSUE #116: Should support Control-C to interrupt the generator
By default, Go programs ignore Control-C and other instructions to stop; we should make the required changes to ensure a user can interrupt generation if it starts running amuck.

@Porges speculated that this ties in with the use of `context` in the current JSON AST handlers.

Porges:
The support for signal handling was already hooked up but we didn’t check for cancellation anywhere.

With this change we check for cancellation:

* During JSON Schema loading: for each HTTP request or file open.
* While cleaning the output directory: once per file or directory to be cleaned, also while walking the tree.
* While converting the JSON AST: once per handler.
* While emitting the output files: once per package.

The JSON Schema library doesn’t support cancellation so this is implemented there by supplying a filesystem implementation that supports cancellation; there is no way to supply an implementation for the HTTP requests so instead we add cancellation to the global HTTP transport (in `gen.go`). We should try and push a change upstream to either add cancellation overall or allow overriding the HTTP client/transport.

Fixes #116.

### DIFF
--- a/hack/generator/cmd/gen/gen.go
+++ b/hack/generator/cmd/gen/gen.go
@@ -7,6 +7,7 @@ package gen
 
 import (
 	"context"
+	"net/http"
 
 	"github.com/Azure/k8s-infra/hack/generator/pkg/codegen"
 	"github.com/Azure/k8s-infra/hack/generator/pkg/xcobra"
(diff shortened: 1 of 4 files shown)


hack/generator/cmd/gen/gen.go
```diff
@@ -15,13 +16,30 @@ import (
 	"k8s.io/klog/v2"
 )
 
+type cancellableTransport struct {
+	ctx   context.Context
+	inner http.RoundTripper
+}
+
+func (transport *cancellableTransport) RoundTrip(request *http.Request) (*http.Response, error) {
```
matthchr, line 24: Is this impl-ing an interface? Should we have an assertion that the interface we want is impled?

hack/generator/pkg/codegen/code_generator.go
```diff
@@ -93,6 +94,9 @@ func (generator *CodeGenerator) Generate(ctx context.Context, outputFolder strin
 	// emit each package
 	klog.V(0).Infof("Writing output files into %v", outputFolder)
 	for _, pkg := range packages {
+		if ctx.Err() != nil {
```
matthchr, line 97: minor: worth annotating each of these requests with a `// check for cancellation`?

I ask just because `ctx.Err()` doesn't scream "cancellation" to me... although maybe that's just a failing of my go knowledge and this is standard go cancellation.

hack/generator/pkg/codegen/code_generator.go
```diff
@@ -282,17 +286,32 @@ func loadConfiguration(configurationFile string) (*config.Configuration, error)
 	return result, nil
 }
 
-func loadSchema(source string) (*gojsonschema.Schema, error) {
+type cancellableFS struct {
```
matthchr, line 289: minor: IMO, expand FS to... FileSystem?

hack/generator/pkg/jsonast/jsonast.go
```diff
@@ -8,11 +8,12 @@ package jsonast
 import (
 	"context"
 	"fmt"
-	"github.com/Azure/k8s-infra/hack/generator/pkg/config"
 	"net/url"
 	"regexp"
 	"strings"
 
+	"github.com/Azure/k8s-infra/hack/generator/pkg/config"
```
matthchr, line 15: super minor: put this down by other other hack/generator/pkg imports?

Porges: I'm surprised go fmt didn't do this. I don't manage the imports myself!

devigned:
Great use of the context! So glad to see the commands respecting cancellation.

PR #146, opened 2020-06-12
Build improvements

### LINKED ISSUE #144: Allow output directory to be configured
At the moment we're hard coding output to go into an `\apis\` subfolder under the current directory when the generator runs.

We should augment our configuration to allow the output folder to be explicitly specified, using `.\apis` as a default. 

This is probably also a good time to allow the config file to be specified on the command line too.

matthchr:
- Add makefile target for running k8sinfra-gen.
- Allow user to specify output directory in configuration file, and configuration file on cmdline
  - This fixes #144.
- Improved configuration yaml file property casing.
- Add new dependency because golang filepath.glob doesn't support recursive matching.


hack/generator/Makefile
```diff
@@ -17,6 +17,10 @@ all: build test
 manifests: $(CONTROLLER_GEN)
 	$(Q) $(CONTROLLER_GEN) $(CRD_OPTIONS) rbac:roleName=manager-role webhook paths="./apis/..." output:crd:artifacts:config=config/crd/bases
 
+.PHONY: gen-arm
```
babbageclunk, line 20: I tried running `make gen-arm` with the output from a previous run still in `apis/` and I got lots of errors (compile errors in the generated code) - is it worth having a `make clean` or having this get rid of the previous output before building?

matthchr: I merged your PR which fixes this.

PR #147, opened 2020-06-12
Specification for type references and ownership

matthchr:


spec/type_references_and_ownership.md
```diff
@@ -0,0 +1,295 @@
...
+1. [RouteTable -> Route](https://github.com/Azure/k8s-infra/blob/master/apis/microsoft.network/v1/routetable_types.go#L18) ([json schema](https://schema.management.azure.com/schemas/2020-03-01/Microsoft.Network.json))
+2. BatchAccount -> Pool ([json schema](https://schema.management.azure.com/schemas/2017-09-01/Microsoft.Batch.json))
+3. ResourceGroup -> Any resource
+
+This relationship tells us two things:
+1. Where to create/manage the dependent resource (this `Route` goes in this particular `RouteTable`, this `RouteTable` has this `Route`)
+2. That the dependent resource should be deleted when the parent resource is deleted. TODO: Race conditions here
+
+Note that sometimes an owning type has its dependent type embedded as a property 
+(for example: [RouteTable has a property routes - look for "RouteTablePropertiesFormat"](https://schema.management.azure.com/schemas/2020-03-01/Microsoft.Network.json)
```
theunrepentantgeek, line 35:
```suggestion
(for example: `RouteTable` has the property [`RouteTablePropertiesFormat`](https://schema.management.azure.com/schemas/2020-03-01/Microsoft.Network.json))
```

spec/type_references_and_ownership.md
```diff
@@ -0,0 +1,295 @@
...
+Most types don't embed the dependent resource directly in the owning resource. We need to cater to both scenarios.
```
theunrepentantgeek, line 36:
```suggestion
Most types do not embed the dependent resource directly in the owning resource. We need to cater to both scenarios.
```

Suggest avoiding the contraction here in order to provide emphasis - it's possibly the most important part of the sentence.

spec/type_references_and_ownership.md
```diff
@@ -0,0 +1,295 @@
...
+
+## What does expressing these relationships look like in existing solutions?
+
```
theunrepentantgeek, line 39: Should put a summary in here so that readers have an overview of what the following subsections cover.

spec/type_references_and_ownership.md
```diff
@@ -0,0 +1,295 @@
...
+### Related/Linked resources
+
+#### What ARM does
+These are just properties (often but not always called `id`) which refer to the fully qualified ARM ID of another resource. For example see a
+[sample deployment template for a VMSS refering to an existing vnet](https://github.com/Azure/azure-quickstart-templates/blob/master/201-vmss-existing-vnet/azuredeploy.json#L136).
```
theunrepentantgeek, line 44: Suggest copying a fragment of that example in here so that readers don't have to look elsewhere for details.

spec/type_references_and_ownership.md
```diff
@@ -0,0 +1,295 @@
...
+
+#### What ASO does
+Similar to what ARM templates do, ASO uses the decomposition of fully qualified resource id to reference another resource, as seen [here for VMSS -> VNet](https://github.com/Azure/azure-service-operator/blob/92240406aff3863f3a267d8a1dc1e28aa3e841ae/api/v1alpha1/azurevmscaleset_types.go#L25)
+```go
```
theunrepentantgeek, line 48: Blank line before code blocks. #nit

spec/type_references_and_ownership.md
```diff
@@ -0,0 +1,295 @@
...
+    Group string
+
+    // The "Kind" of the owning resources GVK. There may be an annotation on the property 
+    // that specifies Kind, in which case this is optional (if supplied it must match the kind
+    // on the annotation). For resources which do not have a kind annotation, this is required.
+    // TODO: It's possibly going to be confusing to customers when they have to specify this and when they don't... 
+    Kind string
+
+    // This is the name of the kubernetes resource to reference. 
+	// References across namespaces are not supported.
```
theunrepentantgeek, line 151:
```suggestion
    // References across namespaces are not supported.
```

spec/type_references_and_ownership.md
```diff
@@ -0,0 +1,295 @@
...
+    // references are. There are RBAC considerations here though so probably easier to just start by 
+    // disallowing cross-namespace references
+}
+
+type AzureReference struct {
+    // The subscription the Azure resource is in. If empty, this defaults to the subscription the resource 
+    // holding the reference is in. 
+    // TODO: If we're worried about this we could also remove it initially, ASO currently doesn't support cross sub references anyway
+    Subscription string
+    // The resource group the Azure resource is in. If empty, this defaults to the resource group the resource
```
theunrepentantgeek, line 164: Add gaps between members. #nit

spec/type_references_and_ownership.md
```diff
@@ -0,0 +1,295 @@
...
+7. How does this property (which doesn't exist in the Azure ARM spec) get transformed into the right ARM JSON?
+
+## Questions
+These are questions I am posing to the group - I don't expect to have an answer without input from the group.
+
+1. Basically everything above is extremely ARM specific. IDs are represented in a different way for data planes and non-ARM endpoints, 
+   and those endpoints JSON schemas may not have the magical `resources` property which allows us to identify owner-dependent relationships.
+   How do we structure the code to solve our ARM problems but allow for alternate solutions for non-ARM?
+2. Is subscription a parameter to in YAMLs, or is the operator configured to act only on a certain subscription?
+   If we only operate on a single sub do we not support cross-sub? Do we care?
```
theunrepentantgeek, line 245: I think starting with a single subscription is probably good enough for a starting point, as long as we don't box ourselves in so that cross subscription is impossible in the future. AFAIK, most Azure customers only use a single subscription (I don't know if this is true for likely ASO customers though)

spec/type_references_and_ownership.md
```diff
@@ -0,0 +1,295 @@
...
+#### Cons
+1. This requires that we detect and label all references appropriately.
+2. It means that you can only reference other resources which are being tracked by Kubernetes.
+   It likely will mean we need to introduce the notion of tracked 
+   resources (which the operator watches but doesn't own) purely for the purpose of 
+   allowing references to those resources. 
+
+### Use built-in OwnerReference for owner references
+#### Pros
+1. Basically none
```
theunrepentantgeek, line 279: 😁

spec/type_references_and_ownership.md
```diff
@@ -0,0 +1,295 @@
...
+3. Define how Kubernetes interacts with Azure resources not created/managed by Kubernetes, for example resources which were created prior to the 
+   customer onboarding to the Azure Service Operator.
+ 
+## Non-goals
+??
+
+## Different kinds of resource relationships in Azure
+#### Related/linked resources 
+Two resources are related to one another ("has-a" relationship), but there is no ownership.
+Example: [VMSS -> Subnet](https://github.com/Azure/k8s-infra/blob/master/apis/microsoft.compute/v1/virtualmachinescaleset_types.go#L169) ([json schema](https://schema.management.azure.com/schemas/2019-07-01/Microsoft.Compute.json))
```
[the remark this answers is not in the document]
matthchr: No, relationships are monodirectional in both cases (a owns b, or a refers to b). If B also refers back to A, that would be specified on the definition of B - but I am not aware of any resources which actually have such a relationship in Azure.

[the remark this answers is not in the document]
matthchr: Will update document to clarify this a bit

spec/type_references_and_ownership.md
```diff
@@ -0,0 +1,295 @@
...
+## Related reading
+1. [Kubernetes garbage collection](https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/)
+
+## Goals
+1. Provide a way for customers to express relationships between Azure resources in Kubernetes in an idiomatic way.
+2. Provide automatic ownership and garbage collection (using [Kubernetes garbage collection](https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/)) 
+   where appropriate (e.g. ResourceGroup as an owner of all the resources inside of it)
+   1. Ideally ResourceGroup is handled the same as other owners and isn't special cased. 
+3. Define how Kubernetes interacts with Azure resources not created/managed by Kubernetes, for example resources which were created prior to the 
+   customer onboarding to the Azure Service Operator.
```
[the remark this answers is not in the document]
davefellows: Ha, interesting, I'd never thought of numbered lists implying priorities. I generally use numbered lists over bullets if I might refer back to specific items later on.

spec/type_references_and_ownership.md
davefellows: We definitely had a few Batch customers that used multiple subscriptions but not sure how likely this will be in the AKS/ASO world. Would be a good thing to try and validate with others that are more familiar with common usage across AKS customers.

spec/type_references_and_ownership.md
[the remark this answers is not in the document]
matthchr: > I generally use numbered lists over bullets if I might refer back to specific items later on.

This is basically my intent too - although [the APA guidelines](https://apastyle.apa.org/style-grammar-guidelines/lists/bulleted) agree with Bevan

spec/type_references_and_ownership.md
```diff
@@ -0,0 +1,295 @@
...
+ 
+## Non-goals
+??
+
+## Different kinds of resource relationships in Azure
+#### Related/linked resources 
+Two resources are related to one another ("has-a" relationship), but there is no ownership.
```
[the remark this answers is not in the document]
matthchr: This isn't actually supposed to be a separate paragraph. The example (line 2 in markdown) really goes with the above line (which is how its rendered when you actually view the rendered markdown)

spec/type_references_and_ownership.md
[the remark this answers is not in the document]
theunrepentantgeek: FWIW, if I need to refer back to specific items I try to name them - otherwise all the references break when I modify/reorder the list.

spec/type_references_and_ownership.md
matthchr: Yeah, I think there are some RBAC/auth issues with cross subscription which we would have to navigate, so I agree that as a starting point we probably should focus on single sub (which is what ASO is already doing so we're aligned there)

spec/type_references_and_ownership.md
```diff
@@ -0,0 +1,295 @@
...
+it's similar to the current approach that `k8s-infra` is taking but with a few changes.
+
+```go
+type AugmentedKnownTypeReference struct {
+    Kubernetes KubernetesReference
+    Azure      AzureReference
+}
+
+type KubernetesReference struct {
+    // The "Group" of the owning resources GVK. There may be an annotation on the property
```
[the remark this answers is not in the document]
matthchr: I've added some details to the document explaining why it isn't needed. The short of it is, the purpose of this is to link to another Kubernetes resource. Because K8s resources are all stored in the storage version in etcd, and all this really is doing is giving us a link to the etcd content (so we can look up the name of the resource in question and then build the ARM ID for it) we don't need the version because it's just an etcd lookup.

Put another way, my understanding is that while there can be different versions of an object in Kubernetes, those versions are just views on the one underlying object - here we only need a link to the one underlying object.

spec/type_references_and_ownership.md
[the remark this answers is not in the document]
matthchr: I fixed this I think

spec/type_references_and_ownership.md
```diff
@@ -0,0 +1,331 @@
...
+    "subnets",
+    subnetName,
+)
+```
+
+Currently ASO does not support cross-subscription references (and some of the resources such as VMSS above don't even support cross-resource group references), but it in theory could by adding some more parameters.
+
+#### What k8s-infra does
+k8s-infra is a bit different in that resource references are in Kubernetes language (namespace + name) and not Azure language (resource group + resource name).
+All resource references are done using a special type `KnownTypeReference` which contains the fully qualified Kubernetes name for the resource.
```
devigned, line 104: Perhaps, consider adding the linkage to the Azure resource, which is the `status.id` field on each resource. This might help to clarify the relationship between the Azure resource and the K8s resource, which is more direct in the aforementioned examples.

spec/type_references_and_ownership.md
```diff
@@ -0,0 +1,331 @@
...
+
+### Dependent Resources
+#### What ARM does
+ARM template deployments support [two different ways of deploying dependent resources](https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/child-resource-name-type):
```
devigned, line 108: I've experimented with these 3 ways of deploying dependent resources. There might even be more ways...

Sibling level resources with a dependsOn:
```json
"resources": [
  {
    "apiVersion": "2019-09-01",
    "type": "Microsoft.Network/virtualNetworks",
    "name": "foo",
    "location": "[resourceGroup().location]",
    "properties": {
      "addressSpace": {
        "addressPrefixes": [
          "10.0.0.0/16"
        ]
      }
    }
  },
  {
    "apiVersion": "2019-09-01",
    "type": "Microsoft.Network/virtualNetworks/subnets",
    "name": "foo/foo-subnet",
    "properties": {
      "addressPrefix": "10.0.0.0/24"
    },
    "dependsOn": [
      "Microsoft.Network/virtualNetworks/foo"
    ]
  }
]
```

Embedded within a set of child resources:
```json
"resources": [
  {
    "apiVersion": "2019-09-01",
    "type": "Microsoft.Network/virtualNetworks",
    "name": "foo",
    "location": "[resourceGroup().location]",
    "properties": {
      "addressSpace": {
        "addressPrefixes": [
          "10.0.0.0/16"
        ]
      }
    },
    "resources": [
      {
        "apiVersion": "2019-09-01",
        "type": "subnets",
        "name": "foo-subnet",
        "properties": {
          "addressPrefix": "10.0.0.0/24"
        },
        "dependsOn": [
          "Microsoft.Network/virtualNetworks/foo"
        ]
      }
    ]
  }
]
```

Embedded directly as an object within the owning resource:
```json
"resources": [
  {
    "apiVersion": "2019-09-01",
    "type": "Microsoft.Network/virtualNetworks",
    "name": "foo",
    "location": "[resourceGroup().location]",
    "properties": {
      "addressSpace": {
        "addressPrefixes": [
          "10.0.0.0/16"
        ]
      },
      "subnets": [
        {
          "name": "foo-subnet",
          "properties": {
            "addressPrefix": "10.0.0.0/24"
          }
        }
      ]
    }
  }
]
```

spec/type_references_and_ownership.md
```diff
@@ -0,0 +1,331 @@
...
+type AugmentedKnownTypeReference struct {
+    Kubernetes KubernetesReference
+    Azure      AzureReference
+}
+
+type KubernetesReference struct {
+    // The "Group" of the owning resources GVK. There may be an annotation on the property 
+    // that specifies Group, in which case this is optional (if supplied it must match the group
+    // on the annotation). For resources which do not have a group annotation, this is required.
+    // TODO: It's possibly going to be confusing to customers when they have to specify this and when they don't...
```
devigned, line 175: I was worried about this being confusing that is why there is [`KnownTypeReference`](https://github.com/Azure/k8s-infra/blob/317e01fb2bbcb1e1b74d346cc1f470c8307f9514/apis/core/v1/types.go#L39-L45) and [`TypedReference`](https://github.com/Azure/k8s-infra/blob/317e01fb2bbcb1e1b74d346cc1f470c8307f9514/apis/core/v1/types.go#L27-L37) in k8s-infra. The latter includes the group and kind, where as the former expects its included in the associated metadata of the struct tag. Turns out we usually know the type based on the location of the references.

spec/type_references_and_ownership.md
```diff
@@ -0,0 +1,331 @@
...
+    // This is the name of the kubernetes resource to reference. 
+    // References across namespaces are not supported.
+    Name string
+
+    // Note that ownership across namespaces in Kubernetes is not allowed, but technically resource
+    // references are. There are RBAC considerations here though so probably easier to just start by 
+    // disallowing cross-namespace references
+}
+
+type AzureReference struct {
```
devigned, line 193: Did you consider just making this an Azure resourceID string? I think you are going to run into problems with only having Sub, Group, and Name.

spec/type_references_and_ownership.md
```diff
@@ -0,0 +1,331 @@
...
+**We need to code generate a field which allows users of the CRDs to express ownership**.
+As discussed in [what ARM does](#What-ARM-does), this is the `resources` property in the ARM deployment templates.
+
+When we determine that a resource is a dependent resource of another resource kind, we will code-generate
+an `OwnerRef` (actual name TBD) property in the dependent resource `Spec`. This will also include an annotation 
+about the expected type of the resource.
+
+```go
+type SubnetSpec struct {
+    OwnerRef AugmentedKnownTypeReference `json:"ownerRef" group:"microsoft.network.infra.azure.com" kind:"VirtualNetwork"`
```
devigned, line 240: Why would `OwnerRef` exist if the `Subnet` resource has an `subnetResource.OwnershipReferences` contains a `corev1.OwnershipReference` corresponding to the VNET? This field seems superfluous.

spec/type_references_and_ownership.md
```diff
@@ -0,0 +1,331 @@
...
+    ...
+}
+```
+
+When users submit a dependent object we should validate that the provided owner reference is present.
+This can be accomplished by making the property required in the CRD.
+One major advantage of this approach is that the customer cannot really get the link type wrong, because we've autogenerated the expected
+group/kind information all names they supply must point to the right kind of resource. If they are using the `AzureReference` portion
```
devigned, line 248: Why not just create an "unmanaged" resource for the Azure referenced resource. For example, the user could create a resource by specifying the only the ID for a `VirtualNetwork`.

```yaml
apiVersion: microsoft.network.infra.azure.com/v20191101
kind: VirtualNetwork
metadata:
  name: sample-virtualnetwork
spec:
  id: /subscriptions/1234/resourcegroups/foo/microsoft.network/virtualnetworks/bazz
```
Perhaps, the controller could fetch the data from the ID and populate the spec from that data. The K8s resource could be marked with an annotation which would keep the Azure resource from being deleted when the K8s resource is deleted.

The above solution would allow for importing Azure resources, simplifies the resource reference in K8s by not needing to create a `AzureReference` b/c everything will be a `KnowTypeReference`, and it provides a simple way to know what K8s need to clean up and what is an "unmanaged" resource.

spec/type_references_and_ownership.md
```diff
@@ -0,0 +1,331 @@
...
+## Questions
+These are questions I am posing to the group - I don't expect to have an answer without input from the group.
+
+- Basically everything above is extremely ARM specific. IDs are represented in a different way for data planes and non-ARM endpoints, 
+  and those endpoints JSON schemas may not have the magical `resources` property which allows us to identify owner-dependent relationships.
+  How do we structure the code to solve our ARM problems but allow for alternate solutions for non-ARM?
+- Is subscription a parameter to in YAMLs, or is the operator configured to act only on a certain subscription?
+  If we only operate on a single sub do we not support cross-sub? Do we care?
+- ARM templates have `dependsOn` for specifying creating order of related resources (where A refers to B) - we don't have that, what does that mean?
+- What to do with awkward resources where the owner requires at least 1 dependent to also be created with it?
```
devigned, line 283: One thing I was thinking about is adding an "embedded: true" tag to the `KnownTypeReferences` where the resource must be embedded and having the K8s to ARM converter fetch embedded resources and include them in the deployment template body of the parent.

spec/type_references_and_ownership.md
```diff
@@ -0,0 +1,331 @@
...
+#### Cons
+- This requires that we detect and label all references appropriately.
+- It means that you can only reference other resources which are being tracked by Kubernetes.
+  It likely will mean we need to introduce the notion of tracked 
+  resources (which the operator watches but doesn't own) purely for the purpose of 
+  allowing references to those resources. 
+
+### Use built-in OwnerReference for owner references
+#### Pros
+- Basically none
```
devigned, line 315: K8s garbage collection seems like a pretty big bonus.

spec/type_references_and_ownership.md
```diff
@@ -0,0 +1,331 @@
...
+
+#### Cons
+- OwnerReference requires the object UID, which cannot be known at template authoring time.
```
devigned, line 318: A user should not be setting these owner references. These owner references should be set by the controller.

spec/type_references_and_ownership.md
```diff
@@ -0,0 +1,331 @@
...
+- OwnerReference only works for ownership relationships, not for
```
devigned, line 319: Were you going to finish with "references"? If so, that is intentional since they are for garbage collection, which a simple reference does not require.

spec/type_references_and_ownership.md
```diff
@@ -0,0 +1,331 @@
...
+- OwnerReference only works for ownership relationships, not for 
+
+## Where ownership references are specified
+### Ownership is from owner to dependent
+#### Pros
+- It makes getting a list of all resources under a particular owner very easy.
+
+#### Cons
+- Adding/deleting a new dependent resource requires an update to the owner.
+- The owner can be in a failed state because dependent resources are missing. It feels like we're repeating our intent here:
```
devigned, line 328: You have specified in a resource spec that is requires a given resource. I believe being in a failed or waiting state is appropriate and is something that is common. For example, what happens to a deployment waiting on a secret reference to be mounted, the deployment waits. If the secret is not available, the deployment will not proceed. I think this is a similar situation.

spec/type_references_and_ownership.md
matthchr: Ah, yeah -- the 3rd option doesn't always work (it depends on the resource type involved). Some list their dependent resource types as a property (in this case `subnets`), but often they are not listed as a property on the parent type.

I didn't include the third option in the list because it's not really applicable to all resources, whereas the other two ways are.

spec/type_references_and_ownership.md
matthchr: Ah yeah, two different types might be a good idea.

> Turns out we usually know the type based on the location of the references.
Yeah based on my hacking so far, for _all_ ownership (parent -> child) relationships we can autogenerate the group and kind as metadata in the field tag.

For just "references" (i.e. VMSS -> Subnet) it's a bit trickier both to identify the properties in question as well as know for sure what the types are supposed to be.

spec/type_references_and_ownership.md
matthchr: Sub, Group and Name (with how I possibly poorly defined name) is == to fully qualified ARM ID.
but yes we could just make this FQArmId and it might make our life easier.

The main advantage of this was just to have defaulting where if you're in a single sub you don't have to keep repeating the sub in all of your YAML (which presumably would get annoying if you wanted to deploy to a different sub, although maybe Kustomize or something can help with that and we don't want to solve it here?)

I suppose the other argument for "just make this the full ARM ID" is that if they really want nice transportable YAMLs they should be using the Kubernetes flavored references, which will all translate quite nicely between namespaces

matthchr: I'm certainly open to changing this - there are some advantages for "similarity to ARM templates" too

spec/type_references_and_ownership.md
matthchr: Let me have a call with you and discuss - We will always be using this under the covers

spec/type_references_and_ownership.md
matthchr: Yeah, this option was exploring "What if we let the user set it" -- we will always be using this under the covers.

spec/type_references_and_ownership.md
matthchr: @devigned and I talked about this offline.

This is basically inverting the ownership approach from what k8s-infra is doing today. Instead of the parent including references to its dependents, the dependent resource includes a reference to its parent. This is in-line with what Azure is actually doing in the ARM templates too (where 

The takeaways from the discussion were:
- The name `OwnerRef` probably needs to be reconsidered. It's confusing/collides with `OwnerReferences` in the Kubernetes object metadata.
- The reason that the `KnownTypeReference` in k8s-infra went from parent -> child is because there are a number of resources where the child resource _is a property_ of the parent. A good example of this is `RouteTable`. David and I went through `RouteTable` as an exercise though and confirmed that you can create a route either by putting it side-by-side and using `dependsOn` in the ARM template, OR by putting the resource embedded in the parent resource. Both are actually identical in terms of what they do in ARM. Not every resource has this kind of relationship though (for example `BatchAccount` -> `Pool` doesn't) so we can't rely on that. After much discussion it was agreed that dependent resources referencing their owners (as they do in ARM templates today when deployed side-by-side with `dependsOn` seems a reasonable approach to proceed with).

The other key highlight is that, this field (called `OwnerRef` here, final name TBD) is what we will use in the controller as metadata to set the actual `ownershipReferences` property in `k8s` metadata.

spec/type_references_and_ownership.md
matthchr: See comment above -- we will 100% be using `OwnerReference` - but we will not be making the customer set it manually, we will figure what resources own what based on `AugmentedKnownTypeReference` (actual final name TBD).

spec/type_references_and_ownership.md
matthchr: yes... fixed.

spec/type_references_and_ownership.md
devigned: Related, but the authoring experience could become disjointed if one is often working with existing infrastructure via Azure'isms when they are used to K8s'isms. Just something to consider when bridging the gap.

spec/type_references_and_ownership.md
matthchr: This is in the list of open questions now (should ownership and references use the same type or different types).

I tend you agree with you @devigned that it would be clearer if there were two types, so I'll tentatively plan on updating the specification to include that. Will leave this comment unresolved though so others can see it and weigh in.

spec/type_references_and_ownership.md
matthchr: I've added a whole section on this (`Open Design Question - shape of Azure References`).

I've asked my team to weigh in as well.

spec/type_references_and_ownership.md
matthchr: As discussed I'm sorta 50-50 on this. I think this approach has some advantages and some disadvantages over what I originally proposed. I've tried to cover those in the "Open Design Question - shape of Azure References" section of the document and have asked some other folks to weigh in to see what makes the most sense.

spec/type_references_and_ownership.md
```diff
@@ -0,0 +1,503 @@
...
+```json
+"properties": {
+  "subnet": {
+    "id": "[resourceId(parameters('existingVnetResourceGroupName'), 'Microsoft.Network/virtualNetworks/subnets', parameters('existingVnetName'), parameters('existingSubNetName'))]"
+  },
+}
+```
+
+#### What ASO does
+Similar to what ARM templates do, ASO uses the decomposition of fully qualified resource id to reference another resource, as seen [here for VMSS -> VNet](https://github.com/Azure/azure-service-operator/blob/92240406aff3863f3a267d8a1dc1e28aa3e841ae/api/v1alpha1/azurevmscaleset_types.go#L25)
```
theunrepentantgeek, line 78:
```suggestion
Similar to how ARM templates behave, ASO uses the decomposition of fully qualified resource id to reference another resource, as seen [here for VMSS -> VNet](https://github.com/Azure/azure-service-operator/blob/92240406aff3863f3a267d8a1dc1e28aa3e841ae/api/v1alpha1/azurevmscaleset_types.go#L25)
```

spec/type_references_and_ownership.md
```diff
@@ -0,0 +1,503 @@
...
+subnetIDInput := helpers.MakeResourceID(
+    client.SubscriptionID,
+    resourceGroupName,
+    "Microsoft.Network",
+    "virtualNetworks",
+    vnetName,
+    "subnets",
+    subnetName,
+)
+```
```
theunrepentantgeek, line 103: Perhaps extend the example to show what the result looks like, for those not so familiar with ARM.

spec/type_references_and_ownership.md
```diff
@@ -0,0 +1,503 @@
...
+
+Currently ASO does not support cross-subscription references (and some of the resources such as VMSS above don't even support cross-resource group references), but it in theory could by adding some more parameters.
```
theunrepentantgeek, line 105:
```suggestion
Currently ASO does not support cross-subscription references (and some of the resources such as VMSS don't allow cross-resource group references), but it in theory could by adding parameters.
```

spec/type_references_and_ownership.md
```diff
@@ -0,0 +1,503 @@
...
+
+#### What k8s-infra does
+k8s-infra is a bit different in that resource references are in Kubernetes language (namespace + name) and not Azure language (resource group + resource name).
+All resource references are done using a special type `KnownTypeReference` which contains the fully qualified Kubernetes name for the resource.
```
theunrepentantgeek, line 109:
```suggestion
All resource references are done using the special type `KnownTypeReference` which contains the fully qualified Kubernetes name for the resource.
```

spec/type_references_and_ownership.md
```diff
@@ -0,0 +1,503 @@
...
+    vnetName,
+    "subnets",
+    subnetName,
+)
+```
+
+Currently ASO does not support cross-subscription references (and some of the resources such as VMSS above don't even support cross-resource group references), but it in theory could by adding some more parameters.
+
+#### What k8s-infra does
+k8s-infra is a bit different in that resource references are in Kubernetes language (namespace + name) and not Azure language (resource group + resource name).
```
theunrepentantgeek, line 108:
```suggestion
k8s-infra is a bit different in that resource references are in Kubernetes style (namespace + name) and not Azure style (resource-group + resource-name).
```

spec/type_references_and_ownership.md
```diff
@@ -0,0 +1,503 @@
...
+#### What ARM does
+ARM template deployments support [two different ways of deploying dependent resources](https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/child-resource-name-type):
+
+- Embed the dependent resource directly in the same ARM template in which the owning resource is created, using the `resources` property.
+    - This requires that the owning resource and the dependent resource are created at the same time.
+- Deploy the dependent resource separately.
+    - When doing this, you should specify `dependsOn` stating that your new dependent resource depends on its owner. This is not strictly required in cases where
+      the owner already exists, but is required when it is still being created (possibly in parallel).
+    - The `name` field of the dependent resource in this mode serves a dual purpose - to name the dependent resource and to identify the parent resources.
+      Each segment of the name corresponds to an owning resource, so creating a Batch Pool `foo` in Batch account `account` would have `name` = `account/foo`.
```
theunrepentantgeek, line 121: The two approaches (embed-directly vs deploy-separately) are described using completely different styles. The former is declarative, the later is first person. Parallel branches (as seen here in bullet lists) should follow similar construction to highlight the similarities/differences between the alternatives.

spec/type_references_and_ownership.md
```diff
@@ -0,0 +1,503 @@
...
+    ...
+}
+```
+
+The other pattern is where the owning resource has links to the dependent resources it is expecting to have:
+
+```go
+type RouteTableSpecProperties struct {
+    DisableBGPRoutePropagation bool                          `json:"disableBgpRoutePropagation,omitempty"`
+    RouteRefs                  []azcorev1.KnownTypeReference `json:"routeRefs,omitempty" group:"microsoft.network.infra.azure.com" kind:"Route" owned:"true"`
```
theunrepentantgeek, line 158: Perhaps word-wrap for clarity, with a comment that you'd normally see it all on the one line.

spec/type_references_and_ownership.md
```diff
@@ -0,0 +1,503 @@
...
+}
+```
+
+If the dependent resources aren't there, the status of the owning resource reflects an error explaining that.
+
+## Proposal
+
+### How to represent references
+We construct a new type which is our reference holder. I am calling this approach `AugmentedKnownTypeReference` (real type name TBD), because
```
theunrepentantgeek, line 167: How about `ResourceReference` ?

And should it be an interface, implemented by both `KubernetesReference` and `AzureReference` - or does this not show well in the CRD?

spec/type_references_and_ownership.md
```diff
@@ -0,0 +1,503 @@
...
+    // is that every version of the Foo type has the same ID shape (it's fundamentally not the same kind of resource
+    // if the ID has changed). Since all we need to do is look that entity up in etcd and get its name details so that
+    // we can build a fully qualified name, we don't actually need version here.
+
+    // This is the name of the kubernetes resource to reference. 
+    // References across namespaces are not supported.
+    Name string
+
+    // Note that ownership across namespaces in Kubernetes is not allowed, but technically resource
+    // references are. There are RBAC considerations here though so probably easier to just start by
```
theunrepentantgeek, line 200: I had to google a definition for this - perhaps use the full term, or at least add to a glossary.

spec/type_references_and_ownership.md
```diff
@@ -0,0 +1,503 @@
...
+#### Advantages compared the proposal above
+- References always look very Kubernetes native. This has an advantage especially when a customer is referring to a single
+  shared resource (Storage Account or the like) across many YAMLs. The more usages they have of it the more awkward it
+  gets to refer to it by fully qualified ARM ID all the time.
+- The structure of the reference type would be simpler because it's not `Azure` or `Kubernetes`, it could just look like
+  `KnownTypeReference` does today in k8s-infra.
+
+#### Disadvantages compared to the proposal above
+- It means that you can only reference other resources which are being tracked by Kubernetes,
+  which means we likely need to introduce the notion of unmanaged
```
theunrepentantgeek, line 268:
```suggestion
- Only able to reference other resources which are being tracked by Kubernetes. 
  We would likely need to introduce the notion of unmanaged 
```

spec/type_references_and_ownership.md
```diff
@@ -0,0 +1,503 @@
...
+- References always look very Kubernetes native. This has an advantage especially when a customer is referring to a single
+  shared resource (Storage Account or the like) across many YAMLs. The more usages they have of it the more awkward it
+  gets to refer to it by fully qualified ARM ID all the time.
+- The structure of the reference type would be simpler because it's not `Azure` or `Kubernetes`, it could just look like
+  `KnownTypeReference` does today in k8s-infra.
+
+#### Disadvantages compared to the proposal above
+- It means that you can only reference other resources which are being tracked by Kubernetes,
+  which means we likely need to introduce the notion of unmanaged 
+  resources (which the operator watches but doesn't own) for the purpose of
```
theunrepentantgeek, line 269:
```suggestion
  resources (which the operator _watches_ but does not _own_) for the purpose of 
```

(Expanded `doesn't` to `does not` for emphasis.)

spec/type_references_and_ownership.md
```diff
@@ -0,0 +1,503 @@
...
+
+#### Disadvantages compared to the proposal above
+- It means that you can only reference other resources which are being tracked by Kubernetes,
+  which means we likely need to introduce the notion of unmanaged 
+  resources (which the operator watches but doesn't own) for the purpose of 
+  allowing references to those resources. This puts some additional burden on the customers mental model
+  as in order to know what an action will do on a resource they need to know if it's managed or unmanaged.
+
+### Implementation
+There are three key pillars required for implementing the proposal as discussed
```
theunrepentantgeek, line 274:
```suggestion
There are three key pillars required for implementing the proposal as described
```

spec/type_references_and_ownership.md
```diff
@@ -0,0 +1,503 @@
...
+
+#### Identify resource relationships
+
+##### Resource references (for related resources)
+We must identify all resource references in the JSON schema and transform their type to `AugmentedKnownTypeReference`.
+There is no specific marker which means: "This field is a reference" - most are called `id` but that's not a guarantee.
+For example on the [VirtualMachineScaleSetIPConfigurationProperties](https://schema.management.azure.com/schemas/2019-07-01/Microsoft.Compute.json) 
+the `subnet` field is of custom type `ApiEntityReference`, which has an `id` field where you put the ARM ID for a subnet.
+This may require some manual work. One thing we can investigate doing long term
+is see if there's a way to get teams to annotate "links" in their Swagger somehow...
```
theunrepentantgeek, line 284: At this level of nesting, I found using a bold sentence fragment at the start of a paragraph read better. 

For example:

> **For references to related resources** found in the JSON schema, we must transform their type to `AugmentedKnownTypeReference`.
> There is no specific marker which means: "This field is a reference" - most are called `id` but that's not a guarantee.
> ...
> **For references to owned/owning resources**, we ...

spec/type_references_and_ownership.md
```diff
@@ -0,0 +1,503 @@
...
+    if armTransformer, ok := newEtcdObject.(ARMTransformer); ok {
+        toGet = armTransformer.ToArm(ownerArmId) // This just converts from an empty etcdObject shape to an empty arm object shape
+        armClient.GetIt(armId, toGet)
+    
+        armTransformer.FromArm(toGet)
+    }
+}
+```
+
+### A note on the differences between owner references and arbitrary resource references
```
theunrepentantgeek, line 400:
```suggestion
### Owner references vs Arbitrary references
```

spec/type_references_and_ownership.md
```diff
@@ -0,0 +1,503 @@
...
+Because we are code-generating all of the `ownerRef` fields, we can always supply the annotations for group and kind automatically, using the information
+provided to us by the ARM json specification. **This is not the case for abitrary references (`id`'s) to other resources**. We do not actually know
+programmatically what type that reference is, and it some cases it may actually be allowed to point to multiple different types 
+(think custom image vs shared image gallery).
+
+Currently, the plan is to trust the customer gets it right and ensure we have good
+failure modes for resources if they actually got it wrong (which will cause their resource to move to an unhealthy state).
```
theunrepentantgeek, line 407: Good diagnostics will be critical to making this work for customers.

spec/type_references_and_ownership.md
```diff
@@ -0,0 +1,503 @@
...
+(think custom image vs shared image gallery).
+
+Currently, the plan is to trust the customer gets it right and ensure we have good
+failure modes for resources if they actually got it wrong (which will cause their resource to move to an unhealthy state). 
+
+## FAQ
+
+#### What happens when a dependent resource specifies an `ownerRef` that doesn't exist?
+The dependent resource will be stuck in an unprovisioned state with an error stating that the owner doesn't exist.
+If the owner is created, the dependent resource will also be created automatically by the reconciliation loop.
```
theunrepentantgeek, line 413:
```suggestion
If the owner is created, the dependent resource will then be created by the reconciliation loop automatically.
```

spec/type_references_and_ownership.md
```diff
@@ -0,0 +1,503 @@
...
+
+#### What happens when a resource contains a link to another resource which doesn't exist?
+The resource with the link will be stuck in an unprovisioned state with an error stating that the linked resource doesn't exist.
+This behavior is the same as for a dependent resource with a non-existant owner.
```
theunrepentantgeek, line 417: Might this result in deadlock if two resources reference each other? We might need to know which one to create first.

spec/type_references_and_ownership.md
```diff
@@ -0,0 +1,503 @@
...
+  is managed by the service operator, only those parts can have their lifecycles managed.  
+
+## Different kinds of resource relationships in Azure
+#### Related/linked resources 
+Two resources are related to one another ("has-a" relationship), but there is no ownership.
+Example: [VMSS -> Subnet](https://github.com/Azure/k8s-infra/blob/master/apis/microsoft.compute/v1/virtualmachinescaleset_types.go#L169) ([json schema](https://schema.management.azure.com/schemas/2019-07-01/Microsoft.Compute.json)).
+
+This relationship is stereotypically one-way (a VMSS refers to a Subnet, but a Subnet _does not_ refer to a VMSS).
+
+#### Owner and dependent
```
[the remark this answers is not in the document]
matthchr: Dependent is the word used by the [official Kubernetes documentation](https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/)
> Some Kubernetes objects are owners of other objects. For example, a ReplicaSet is the owner of a set of Pods. The owned objects are called dependents of the owner object.

spec/type_references_and_ownership.md
matthchr: If that was allowed by ARM, yes that could be an issue. I don't think that's allowed by ARM... will do some reading to see if I can find official documentation on that.

spec/type_references_and_ownership.md
matthchr: Have updated the spec to default to this approach (two types)

spec/type_references_and_ownership.md
matthchr: Added a link at the top of the documented for related reading

spec/type_references_and_ownership.md
matthchr: Suggestion applied

spec/type_references_and_ownership.md
```diff
@@ -0,0 +1,675 @@
...
+  the operator is not watching/monitoring the resource in question
+  and as such cannot propagate deletes. Put another way: for the operator to manage ownership/object lifecycles, the entire resource hierarchy must exist _within Kubernetes_. If only part of the resource hierarchy
+  is managed by the service operator, only those parts can have their lifecycles managed.  
+
+## Different kinds of resource relationships in Azure
+#### Related/linked resources 
+Two resources are related to one another ("has-a" relationship), but there is no ownership.
+Example: [VMSS -> Subnet](https://github.com/Azure/k8s-infra/blob/master/apis/microsoft.compute/v1/virtualmachinescaleset_types.go#L169) ([json schema](https://schema.management.azure.com/schemas/2019-07-01/Microsoft.Compute.json)).
+
+This relationship is stereotypically one-way (a VMSS refers to a Subnet, but a Subnet _does not_ refer to a VMSS).
```
Porges, line 34:
```suggestion
This relationship is always one-way (a VMSS refers to a Subnet, but a Subnet _does not_ refer to a VMSS).
```

spec/type_references_and_ownership.md
```diff
@@ -0,0 +1,675 @@
...
+```json
+"properties": {
+  "subnet": {
+    "id": "[resourceId(parameters('existingVnetResourceGroupName'), 'Microsoft.Network/virtualNetworks/subnets', parameters('existingVnetName'), parameters('existingSubNetName'))]"
+  },
+}
+```
+
+#### What does ASO do?
+Similar to how ARM templates behave, ASO uses the decomposition of fully qualified resource id to reference another resource, as seen [here for VMSS -> VNet](https://github.com/Azure/azure-service-operator/blob/92240406aff3863f3a267d8a1dc1e28aa3e841ae/api/v1alpha1/azurevmscaleset_types.go#L25)
```
Porges, line 78:
```suggestion
Similar to how ARM templates behave, ASO uses the decomposition of fully qualified resource id to reference another resource, as seen [here for VMSS → VNet](https://github.com/Azure/azure-service-operator/blob/92240406aff3863f3a267d8a1dc1e28aa3e841ae/api/v1alpha1/azurevmscaleset_types.go#L25)
```

spec/type_references_and_ownership.md
```diff
@@ -0,0 +1,675 @@
...
+  by the service operator.
+  While this proposal allows references to point to external resources not managed by the service operator,
+  the operator is not watching/monitoring the resource in question
+  and as such cannot propagate deletes. Put another way: for the operator to manage ownership/object lifecycles, the entire resource hierarchy must exist _within Kubernetes_. If only part of the resource hierarchy
+  is managed by the service operator, only those parts can have their lifecycles managed.  
+
+## Different kinds of resource relationships in Azure
+#### Related/linked resources 
+Two resources are related to one another ("has-a" relationship), but there is no ownership.
+Example: [VMSS -> Subnet](https://github.com/Azure/k8s-infra/blob/master/apis/microsoft.compute/v1/virtualmachinescaleset_types.go#L169) ([json schema](https://schema.management.azure.com/schemas/2019-07-01/Microsoft.Compute.json)).
```
Porges, line 32:
```suggestion
Example: [VMSS → Subnet](https://github.com/Azure/k8s-infra/blob/master/apis/microsoft.compute/v1/virtualmachinescaleset_types.go#L169) ([json schema](https://schema.management.azure.com/schemas/2019-07-01/Microsoft.Compute.json)).
```

spec/type_references_and_ownership.md
```diff
@@ -0,0 +1,675 @@
...
+In the `ResourceReference` case, we cannot know the type we're looking for, so we must accept what the customer has provided and ensure that we have good error messages
+if they have provided a link to an invalid resource (usually the error from Azure should suffice).
+
+### ResourceLifeCycle and unmanaged resources
+In order to keep references Kubernetes-native, allow a "single pane of glass" for customers looking at their Azure resources through Kubernetes, and 
+allow references to resources that were created before the customer onboarded to the operator, we introduce a new mode to each resource: `ResourceLifeCycle`.
+
+`ResourceLifeCycle` can be either `Managed` or `Unmanaged`.
+
+`ResourceLifeCycle` is not specified by the customer explicitly in the `Spec`, instead
```
Porges, line 306: sentence here is unfinished?

spec/type_references_and_ownership.md
```diff
@@ -0,0 +1,675 @@
...
+
+### ResourceLifeCycle and unmanaged resources
+In order to keep references Kubernetes-native, allow a "single pane of glass" for customers looking at their Azure resources through Kubernetes, and 
+allow references to resources that were created before the customer onboarded to the operator, we introduce a new mode to each resource: `ResourceLifeCycle`.
+
+`ResourceLifeCycle` can be either `Managed` or `Unmanaged`.
+
+`ResourceLifeCycle` is not specified by the customer explicitly in the `Spec`, instead 
+
+RouteTable + Routes issue (multiple routes of the same name are allowed)
```
Porges, line 308: Some heading markup missing in this and the following paragraphs, I think?

spec/type_references_and_ownership.md
```diff
@@ -0,0 +1,675 @@
...
+
+Options for this:
+Note that all of these options share this restriction: Each resource must be imported, i.e. to import a VNET you may need to import
```
Porges, line 311:
```suggestion
Note that all of these options share this restriction: Each resource must be imported, e.g. to import a VNET you may need to import
```

spec/type_references_and_ownership.md
```diff
@@ -0,0 +1,675 @@
...
+the resource group the VNET is in, and then the VNET (with an `Owner` reference pointing to the imported `ResourceGroup`). 
+
+Option 1: Users must create a _valid_ resource with the same name as the resource they want to track. If this resources spec differs from what is in Azure, an error is logged
```
Porges, line 314:
```suggestion
Option 1: Users must create a _valid_ resource with the same name as the resource they want to track. If this resource’s spec differs from what is in Azure, an error is logged 
```

spec/type_references_and_ownership.md
```diff
@@ -0,0 +1,675 @@
...
+
+RouteTable + Routes issue (multiple routes of the same name are allowed)
+
+Options for this:
+Note that all of these options share this restriction: Each resource must be imported, i.e. to import a VNET you may need to import
+the resource group the VNET is in, and then the VNET (with an `Owner` reference pointing to the imported `ResourceGroup`). 
+
+Option 1: Users must create a _valid_ resource with the same name as the resource they want to track. If this resources spec differs from what is in Azure, an error is logged 
+but we never actually apply any state to Azure (i.e. we don't try to sync to the spec). A tag in the metadata must be added to inform the operator not to sync.
+Note that because of how we managed names in Azure and names in Kubernetes,
```
Porges, line 316: Unfinished sentence?

spec/type_references_and_ownership.md
```diff
@@ -0,0 +1,675 @@
...
+Note that because of how we managed names in Azure and names in Kubernetes,  
+
+Advantages
+- Swapping from unmanaged to managed is super easy, just remove the tag blocking the reconciliation loop.
+
+Disadvantages
+- There is possibly a significant amount of extra effort required to re-specify a resource whose shape we really just want to "import" from ARM. Worse for large trees of objects
+  or deeply nested objects.
+- If the tag is forgotten (or has a typo) we will try to manage a resource which we shouldn't be managing. This could be _very_ problematic depending on how different
+  the specification is from what's there in Azure.
```
Porges, line 325:
```suggestion
  the specification is from what exists in Azure.
```

spec/type_references_and_ownership.md
```diff
@@ -0,0 +1,675 @@
...
+        - required: ["owner", "foo"]
+```
+
+Note that it has to be `anyOf` because `oneOf` disallows multiple matches, and `owner` + `foo` matches both sets in the example above.
+
+Advantages
+- Represents what we want and maintains better automatic validation.
+
+Disadvantages
+- Kubebuilder doesn't generating this, so we would have to come up with another way to do it, or possibly upstream changes to Kubebuilder to support it.
```
Porges, line 363:
```suggestion
- Kubebuilder doesn't support generating this, so we would have to come up with another way to do it, or possibly upstream changes to Kubebuilder to support it.
```

spec/type_references_and_ownership.md
```diff
@@ -0,0 +1,675 @@
...
+
+## FAQ
+
+### What happens when a dependent resource specifies an `Owner` that doesn't exist?
+The dependent resource will be stuck in an unprovisioned state with an error stating that the owner doesn't exist.
+If the owner is created, the dependent resource will then be created by the reconciliation loop automatically.
+
+### What happens when a resource contains a link to another resource which doesn't exist?
+The resource with the link will be stuck in an unprovisioned state with an error stating that the linked resource doesn't exist.
+This behavior is the same as for a dependent resource with a non-existant owner.
```
Porges, line 535:
```suggestion
This behavior is the same as for a dependent resource with a non-existent owner.
```

spec/type_references_and_ownership.md
```diff
@@ -0,0 +1,675 @@
...
+There are a few different ways to perform ARM deployments as [discussed in Dependent Resources](#dependent-resources).
+Due to the nature of Kubernetes CRDs, each resource is managed separately and has its own reconcilation loop. It doesn't make sense to try to 
+deploy a single ARM template with the entire resource graph. Each resource will be done in its own deployment (with
+a `dependsOn` specified if required).
+
+### Aren't there going to be races in resource creation?
+Yes. If you have a complex hierarchy of resources (where resources have involved relationships between one another) and submit 
+all of their YAMLs to the operator at the same time it is likely that some requests when sent to ARM will fail because of missing dependencies.
+Those resources that failed to deploy initially will be in an unprovisioned state in Kubernetes, and eventually
+all the resources will be created through multiple iterations of the reconciliation loop.
```
Porges, line 547: We could also use `DependsOn` in the deployment template to get an ordering that we desire. This might not be needed but it could speed up initial deployment rather than having to wait for the reconciliation loop to fix everything.

spec/type_references_and_ownership.md
```diff
@@ -0,0 +1,675 @@
...
+the structural schema defined in the CRD, the generic controller will attempt to look up
+the owning resource in etcd (or in ARM if it's an `AzureReference`).
+
+If the generic controller finds the owning resource, it updates the `ownerReference` in the object metadata
+to include the `uid` of the owning resource and then submits an ARM template to ARM using the 
+name of the owner and the name of the resource to build the name specified in the ARM template. It will
+include the name of the owner in the `dependsOn` field. 
+
+### What happens if an owning resource is deleted and immediately recreated?
+Kubernetes garbage collection is based on object `uid`'s. As discussed above we bind to that `uid` on
```
Porges, line 572:
```suggestion
Kubernetes garbage collection is based on object `uid`'s. As discussed above, we bind to that `uid` on 
```

spec/type_references_and_ownership.md
```diff
@@ -0,0 +1,675 @@
...
+### What happens if an owning resource is deleted and immediately recreated?
+Kubernetes garbage collection is based on object `uid`'s. As discussed above we bind to that `uid` on 
+dependent resource creation. If a resource is deleted and then recreated Kubernetes will still understand
+that the new resource is fundamentally different than the old resource and garbage collection will happen
+as expected. The result will be that there is a new owning resource but all of its dependent resources were 
+deleted (in Azure and in k8s).
+
+## TODOs
+- How can we allow customers to easily find all dependents for a particular owner (i.e. all subnets of a vnet) using `kubectl`?
+- Cross subscription refs?
```
Porges, line 580: Note here that these aren't supported by Azure.

Porges (changes requested):

spec/type_references_and_ownership.md
matthchr: Oops! Fixed

spec/type_references_and_ownership.md
matthchr: Good eye

spec/type_references_and_ownership.md
matthchr: I think that `DependsOn` requires multiple resources submitted in a single ARM template in order to get that ordering - since our controller instances will each be managing a single type I am pretty sure we don't have an easy path to a unified ARM template with N resources in it?

> Within your template, the dependsOn element enables you to define one resource as a dependent on one or more resources.

https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/define-resource-dependency

spec/type_references_and_ownership.md
matthchr: They are in some cases I believe? https://stackoverflow.com/questions/61001675/azure-vnet-cross-subscription-peering

Porges: Right, but in general we can’t do it. That example is specific to VNets and needed for their specific functionality.

Porges (approved):

spec/type_references_and_ownership.md
matthchr: Ahh, got it. Will fix up document to clarify.

PR #150, opened 2020-06-17
Report the git commit hash in groupversion_info_gen.go

### LINKED ISSUE #118: Improvement: Include generator version in generated files
To allow code generation issues to be correctly traced back to the responsible version of the code generator, we desire for the files to be stamped with the version of the generator.

Marking all files would result in all files showing as changed when regenerated, which isn't good for tracking actual changes. Instead, let's embed the code generator version in the `groupversion_info.go`. file

babbageclunk:
This will help us trace a generated package back to the code that produced it. The git commit hash (and whether the tree is dirty) is fed into the build from the Makefile.

Fixes #118


babbageclunk:
Ha, got CLA'd.

hack/generator/Makefile
```diff
@@ -19,7 +20,7 @@ manifests: $(CONTROLLER_GEN)
 
 .PHONY: gen-arm
 gen-arm: rm-apis build ; $(info $(M) running k8sinfra-gen gen…) @ ## Running gen
-	$(Q) ./bin/$(APP) gen azure-arm.yaml
+	$(Q) ./bin/$(APP) -v 1 gen azure-arm.yaml
```
matthchr, line 23: This is setting log verbosity right?

Is there a reason we want more than the default level of logging during the build?
Phrased another way: if there is information we think is useful to be displayed all the time would we rather just put that stuff in v=0 (which I think is the default?)

hack/generator/pkg/codegen/version.go
```diff
@@ -0,0 +1,25 @@
...
+)
+
+func combinedVersion() string {
+	result := GitCommit
+	if GitTreeState != "clean" {
+		result += fmt.Sprintf(" (tree is %s)", GitTreeState)
+	}
+	return result
+
+}
```
Porges, line 25: minor: some odd formatting here?

hack/generator/Makefile
babbageclunk: Agreed - changing it back.

hack/generator/pkg/codegen/version.go
babbageclunk: Doh, fixed.

PR #152, opened 2020-06-19
Improve comment formatting

### LINKED ISSUE #139: Reformat comments on struct fields
Currently the comments we generate on each struct field are in different formats, giving a jarring result. Change to use a consistent style 

* Use `// prefix style` for all comments
* Handle line breaks and `<br>` in docs from the JSON schema
* Introduce a blank line prior to each field
* Wordwrap documentation comments at 120 characters
* Remove Markdown bold `**` if present

Final result should look something like this:
``` go
/*Generated from: https://schema.management.azure.com/schemas/2018-10-01/Microsoft.Compute.json#/definitions/VirtualMachineScaleSetDataDisk*/
type VirtualMachineScaleSetDataDisk struct {
    
    //Caching: Specifies the caching requirements. 
    //Possible values are: 
    //None
    //ReadOnly
    //ReadWrite
    //Default: None for Standard storage. ReadOnly for Premium storage.
    Caching *VirtualMachineScaleSetDataDiskCaching `json:"caching"`

    // +kubebuilder:validation:Required
    //CreateOption: The create option.
    CreateOption VirtualMachineScaleSetDataDiskCreateOption `json:"createOption"`
    //DiskSizeGB: Specifies the size of an empty data disk in gigabytes. This element can be used to overwrite the size of 
	//the disk in a virtual machine image. 
    //This value cannot be larger than 1023 GB
    DiskSizeGB *int `json:"diskSizeGB"`

    // +kubebuilder:validation:Required
    //Lun: Specifies the logical unit number of the data disk. This value is used to identify data disks within the VM and 
    //therefore must be unique for each data disk attached to a VM.
    Lun int `json:"lun"`

    //ManagedDisk: The managed disk parameters.
    ManagedDisk *VirtualMachineScaleSetManagedDiskParameters `json:"managedDisk"`

    //Name: The disk name.
    Name *string `json:"name"`

    //WriteAcceleratorEnabled: Specifies whether writeAccelerator should be enabled or disabled on the disk.
    WriteAcceleratorEnabled *bool `json:"writeAcceleratorEnabled"`
}
```

theunrepentantgeek:
Reformats comments to make them easier to read. Resolves #139 

Result looks like this:
```
type RollingUpgradePolicy struct {

	//MaxBatchInstancePercent: The maximum percent of total virtual machine instances
	//that will be upgraded simultaneously by the rolling upgrade in one batch. As
	//this is a maximum, unhealthy instances in previous or future batches can cause
	//the percentage of instances in a batch to decrease to ensure higher reliability.
	//The default value for this parameter is 20%.
	MaxBatchInstancePercent *int `json:"maxBatchInstancePercent"`

	//MaxUnhealthyInstancePercent: The maximum percentage of the total virtual machine
	//instances in the scale set that can be simultaneously unhealthy, either as a
	//result of being upgraded, or by being found in an unhealthy state by the virtual
	//machine health checks before the rolling upgrade aborts. This constraint will be
	//checked prior to starting any batch. The default value for this parameter is
	//20%.
	MaxUnhealthyInstancePercent *int `json:"maxUnhealthyInstancePercent"`
```

Warning ... the technique for putting a blank line prior to each comment is not pretty.


hack/generator/pkg/astmodel/comments.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+}
+
+// formatDocComment splits the supplied comment string up ready for use as a documentation comment
+func formatDocComment(comment string, width int) []string {
+	// Remove markdown bolding,
+	text := strings.ReplaceAll(comment, "**", "")
+
+	// Turn <br> and <br/> into \n
+	text = strings.ReplaceAll(text, "<br/>", "\n")
```
Porges, line 33: suggest living in sin and using `regex.ReplaceAllString` with `<br[^/>]*/?>`

hack/generator/pkg/astmodel/comments.go
```diff
@@ -0,0 +1,102 @@
...
+		if line[index] == ' ' {
+			return index
+		}
+
+		index--
+	}
+
+	// Line contains continuous text, we don't want to break it in two, so find the end of it
+	index = start + width
+	for index < len(line) {
```
Porges, line 93: `strings.Index*` could be useful here

hack/generator/pkg/astmodel/comments.go
```diff
@@ -0,0 +1,102 @@
...
+// If breaking at a space, this will give a trailing space, but allows for
+// breaking at other points too as no characters will be omitted.
+func findBreakPoint(line string, start int, width int) int {
+	index := start + width
+	if index >= len(line) {
+		return len(line)-1
+	}
+
+	// Look for a word break within the line
+	for index > start {
```
Porges, line 83: `strings.LastIndex*` could be useful here

hack/generator/pkg/astmodel/comments.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+
+	return result
+}
+
+// findBreakPoint finds the character at which to break two lines
+// Returned index points the last character that should be included on the line
+// If breaking at a space, this will give a trailing space, but allows for
+// breaking at other points too as no characters will be omitted.
+func findBreakPoint(line string, start int, width int) int {
```
Porges, line 76: I would probably make this return the character _after_ so the caller doesn’t need to adjust the result

hack/generator/pkg/astmodel/comments.go
```diff
@@ -0,0 +1,102 @@
...
+
+		*commentList = append(*commentList, &ast.Comment{
+			Text: line,
+		})
+	}
+}
+
+// formatDocComment splits the supplied comment string up ready for use as a documentation comment
+func formatDocComment(comment string, width int) []string {
+	// Remove markdown bolding,
```
davefellows, line 29: Rogue comma :)

hack/generator/pkg/astmodel/comments.go
davefellows: Nice

hack/generator/pkg/astmodel/comments.go
```diff
@@ -0,0 +1,102 @@
...
+		finish := findBreakPoint(text, start, width)
+		result = append(result, text[start:finish+1])
+		start = finish+1
+	}
+
+	return result
+}
+
+// findBreakPoint finds the character at which to break two lines
+// Returned index points the last character that should be included on the line
```
davefellows, line 73:
```suggestion
// Returned index points to the last character that should be included on the line
```

hack/generator/pkg/astmodel/comments.go
```diff
@@ -0,0 +1,102 @@
...
+	for index < len(line) {
+		if line[index] == ' ' {
+			return index
+		}
+
+		index++
+	}
+
+	return len(line)-1
+}
```
davefellows, line 102: No newline at the end of the file apparently...

hack/generator/pkg/astmodel/file_definition.go
```diff
@@ -204,3 +208,33 @@ func (file FileDefinition) SaveToFile(filePath string) error {
 
 	return file.SaveToWriter(filePath, f)
 }
+
+func (file FileDefinition) AddBlankLinesBeforeComments(buffer bytes.Buffer) bytes.Buffer {
```
davefellows, line 212: Aren't function comments required for public funcs?

hack/generator/pkg/astmodel/file_definition.go
```diff
@@ -171,25 +173,27 @@ func createComments(lines ...string) ([]*ast.Comment, int) {
 func (file FileDefinition) SaveToWriter(filename string, dst io.Writer) error {
 	original := file.AsAst()
 
-	// Write generated source into a memory buffer
+	// Write generated source into a memory unformattedBuffer
```
matthchr, line 176: minor: Seems weird here to say `unformattedBuffer` when the type is actually `Buffer`?  Maybe phrasing could be improved (or just left how it was?)

hack/generator/pkg/astmodel/comments.go
```diff
@@ -0,0 +1,102 @@
...
+	"strings"
+)
+
+// Utility methods for adding comments
+
+func addDocComment(commentList *[]*ast.Comment, comment string, width int) {
+	for _, c := range formatDocComment(comment, width) {
+		line := strings.TrimSpace(c)
+		if !strings.HasPrefix(line, "//") {
+			line = "//"+line
```
matthchr, line 14: Is the reason we're doing `//test` rather than `// text` because of controller-tools behavior with comments and how it renders them?

Might be worth raising an issue there if there already isn't one?

hack/generator/pkg/astmodel/file_definition.go
davefellows: copy/paste bug?

hack/generator/pkg/astmodel/file_definition.go
davefellows: Although, I believe we're using golangci-lint rather than golint which I think defaults to not include this check. The ASO repo doesn't use golangci-lint from what I can see so we need to be aware there may be situations like this where suddenly our code is throwing warnings once we migrate it to ASO. We could of course enable golangci-lint in ASO as well...

hack/generator/pkg/astmodel/comments.go
theunrepentantgeek: I did it this way to make it easier to add other word wrap rules - if we have too much problem with `long_identifiers_with_underscores` we could break at `_` as well, but we wouldn't want to remove that character. #wontfix

hack/generator/pkg/astmodel/comments.go
theunrepentantgeek: Fixed.

hack/generator/pkg/astmodel/comments.go
theunrepentantgeek: Indeed they were.

hack/generator/pkg/astmodel/comments.go
theunrepentantgeek: Yup.

hack/generator/pkg/astmodel/comments.go
theunrepentantgeek: Done.

hack/generator/pkg/astmodel/comments.go
theunrepentantgeek: Thanks.

hack/generator/pkg/astmodel/comments.go
theunrepentantgeek: Fixed.

hack/generator/pkg/astmodel/file_definition.go
theunrepentantgeek: This method doesn't need to be exported, so renamed it to make it internal. Added a documentation comment anyway to clarify intent.

hack/generator/pkg/astmodel/file_definition.go
theunrepentantgeek: search/replace bug when I renamed the variable. #whoops 😊

hack/generator/pkg/astmodel/comments.go
theunrepentantgeek: I thought @Porges had already done that - I'll leave this open for him to confirm.

hack/generator/pkg/astmodel/comments.go
```diff
@@ -0,0 +1,95 @@
...
+	// Line contains continuous text, we don't want to break it in two, so find the end of it
+	index = strings.Index(line[limit:], " ")
+	if index >= 0 {
+		return limit + index
+	}
+
+	return len(line)-1
+}
+
+var brRegex = regexp.MustCompile("<br[^/>]*/?>")
```
Porges, line 95: would put this just before the func that uses it, it’s a bit orphaned down here

theunrepentantgeek: Fair enough. Moved.

hack/generator/pkg/astmodel/comments.go
Porges: I have not 🙂

matthchr:
header check failing for a few new files added:
```
./hack/generator/pkg/astmodel/comments.go
./hack/generator/pkg/astmodel/comments_test.go
##[error]Makefile:78: recipe for target 'header-check' failed
##[error]Process completed with exit code 2.
```

I've opened a PR to make these easier to catch locally: #153

Issue #123, opened 2020-05-27
Consider using go-cmp to simplify implementation of Equals


matthchr:

babbageclunk:
It turns out that go-cmp slows down generation sufficiently that we don't want to use it.

PR #157, opened 2020-06-24
Replace most Equals methods with go-cmp Equal

babbageclunk:
This removes most of the Equals methods from `astmodel` types - only the ones that need some non-default behaviour are kept.

This was done to explore #123 but we might not want to actually land it, for reasons that are called out in the [go-cmp docs](https://pkg.go.dev/github.com/google/go-cmp/cmp).
1. cmp.Equal panics when it encounters a problem comparing things - in our codebase it panics on structs with unexported fields. I've handled this by adding all the types encountered to an allow list, but we might see this for new types that are added.
2. It's noticeably slower - on my machine `time bin/k8sinfra-gen gen azure-arm.yaml` went from about 15s to about 45s.

Neither of these are deal-breakers for the code generation but the savings in terms of removed Equals methods isn't huge either. What do people think?

### DIFF
--- a/hack/generator/go.mod
+++ b/hack/generator/go.mod
@@ -5,6 +5,7 @@ go 1.13
 require (
 	github.com/bmatcuk/doublestar v1.3.1
 	github.com/devigned/tab v0.1.1
+	github.com/google/go-cmp v0.3.0
 	github.com/onsi/gomega v1.8.1
 	github.com/pelletier/go-toml v1.6.0 // indirect
 	github.com/sebdah/goldie/v2 v2.3.0
--- a/hack/generator/go.mod
+++ b/hack/generator/go.mod
@@ -5,7 +5,7 @@ go 1.13
 require (
 	github.com/bmatcuk/doublestar v1.3.1
 	github.com/devigned/tab v0.1.1
-	github.com/google/go-cmp v0.3.0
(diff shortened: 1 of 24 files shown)


babbageclunk:
I don't understand what's happening with the coverage numbers - this PR is a net removal of code lines but not test lines. Shouldn't  coverage have gone up?

theunrepentantgeek:
Every build gets reported as a reduction in coverage by ~3.5%, not sure why.

devigned:
@theunrepentantgeek, I think any PR that comes from a fork only runs the unit tests (tests without secrets). If a PR is opened with a main repo branch, then it will be a "trusted" branch and secrets will be available and integration tests will run also.

theunrepentantgeek:
We've got a bunch of work to come on the code generator - serialization of fields with fixed values (e.g. `api-version`), ownership & references, versioning & conversion methods, unit tests, etc.

A short runtime is going to be critical to our inner-loop so that we can iterate quickly and get these features delivered. Pushing the runtime out to close to a minute therefore seems like a deal breaker to me.

babbageclunk:
After discussion we agreed that the increased runtime would push this into the "Oh, I'll just check email/whatever while this is running..." zone which is definitely a productivity killer. It was worth trying but we don't want to merge it.

PR #158, opened 2020-06-25
TypeDefiners must also support RequiredImports

matthchr:
- Definitions have code associated with them, so they need to have the ability to modify a files set of imports. Specifically struct_definition can produce the metav1 import, and may produce other imports in the future.
- Also a minor modification to how we are excluding some types to be slightly more efficient.


hack/generator/pkg/astmodel/file_definition.go
```diff
@@ -40,18 +40,19 @@ func NewFileDefinition(packageRef *PackageReference, definitions ...TypeDefiner)
...
-	requiredImports[*metav1Import] = struct{}{}
 
 	for _, s := range file.definitions {
-		for _, requiredImport := range s.Type().RequiredImports() {
+		for _, requiredImport := range s.RequiredImports() {
 			// no need to import the current package
 			if !requiredImport.Equals(file.packageReference) {
 				newImport := NewPackageImport(*requiredImport)
+
+				// TODO: Should we change signature of RequiredImports to return this type instead?
```
matthchr, line 51: I'm wondering if we should change RequiredImports to return the `PackageImport` type rather than the `PackageReference` type - that way the type itself can specify what it wants the import name to be, rather than this hack we have here where the renaming happens above.

If there were multiple imports that were the same but with different names we can either error on that, or just pick one and let the types figure it out using `codeGenerationContext` (which they should be able to since the context will have the winner)

matthchr: If folks like this plan/idea I'll do it in this PR too

hack/generator/pkg/astmodel/struct_definition.go
```diff
@@ -56,6 +56,18 @@ func (definition *StructDefinition) WithIsStorageVersion(isStorageVersion bool)
 	return &result
 }
 
+// RequiredImports returns a list of packages required by this type
+func (definition *StructDefinition) RequiredImports() []*PackageReference {
+	typeImports := definition.StructType.RequiredImports()
+
+	if definition.isResource {
+		metav1Import := NewPackageReference("k8s.io/apimachinery/pkg/apis/meta/v1")
```
matthchr, line 64: Should fix the usages of metav1 in here to confirm to the pattern with codeGenContext

hack/generator/pkg/astmodel/importer.go
```diff
@@ -0,0 +1,11 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ */
+
+package astmodel
+
+type PackageImporter interface {
```
Porges, line 8: I don’t much like this as an interface because we never abstract over it (`PackageImporter` is never used as an argument/return type)

matthchr: You prefer just to define 2x the same method (one on each?)

hack/generator/pkg/astmodel/file_definition.go
matthchr: Actually on second thought maybe this doesn't make sense.

definitions/types identify what packages they require. `CodeGenerationContext` is built from that set, naming and disambiguating as required. The `AsAst` methods then all use `CodeGenerationContext` to pick out the name of the package chosen - so there shouldn't be any expectation that a particular package got mapped to a particular name, should always just defer to `CodeGenerationContext` probably, which means what we're doing now is right.

hack/generator/pkg/astmodel/resource.go
```diff
@@ -77,13 +78,25 @@ func (definition *ResourceDefinition) WithDescription(description *string) TypeD
...
+
+	return typeImports
+}
 
 // AsDeclarations converts the ResourceDefinition to a go declaration
 func (definition *ResourceDefinition) AsDeclarations(codeGenerationContext *CodeGenerationContext) []ast.Decl {
 
+	packageName, err := codeGenerationContext.GetImportedPackageName(MetaV1PackageReference)
+	if err != nil {
+		panic(err)
```
Porges, line 95: it would be good to provide a message as well

PR #159, opened 2020-06-26
Split out ResourceDefinition from StructDefinition

Porges:
This removes the bad-feeling `isResource` parameter from `CreateDefinitions` and cleans up some checks in the package/file generation. It also encodes that only resources can have `isStorageVersion` set which wasn’t enforced by the types before. (I also like that it gets us further away from Go-AST-specific naming.)

### DIFF
--- a/hack/generator/pkg/astmodel/resource.go
+++ b/hack/generator/pkg/astmodel/resource.go
@@ -56,7 +56,9 @@ func (definition *ResourceDefinition) Name() *TypeName {
 }
 
 func (definition *ResourceDefinition) Type() Type {
-	return definition.spec // TODO?????
+	return definition.spec // TODO BUG: the status is not considered here
+	// TO FIX: consider lifting up the two methods used on the result of this method
+	// (which are References/RequiredImports) into the TypeDefiner interface
 }
 
 func (definition *ResourceDefinition) MarkAsStorageVersion() *ResourceDefinition {
--- a/hack/generator/pkg/astmodel/resource.go
+++ b/hack/generator/pkg/astmodel/resource.go
@@ -1,0 +1,131 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ */
+
+package astmodel
+
+import (
+	"go/ast"
+	"go/token"
+)
+
+// CreateResourceDefinitions creates definitions for a resource
+func CreateResourceDefinitions(name *TypeName, specType *StructType, statusType *StructType, idFactory IdentifierFactory) (TypeDefiner, []TypeDefiner) {
+
+	var others []TypeDefiner
+
+	var specName *TypeName
+	if specType != nil {
+		specName = NewTypeName(name.PackageReference, name.Name()+"Spec")
+		specDef, specOthers := specType.CreateDefinitions(specName, idFactory)
+
+		others = append(append(others, specDef), specOthers...)
+	} else {
+		panic("spec must always be provided")
+	}
+
+	var statusName *TypeName
+	if statusType != nil {
+		statusName = NewTypeName(name.PackageReference, name.Name()+"Status")
+		statusDef, statusOthers := statusType.CreateDefinitions(statusName, idFactory)
+
+		others = append(append(others, statusDef), statusOthers...)
+	}
+
+	this := &ResourceDefinition{typeName: name, spec: specName, status: statusName, isStorageVersion: false}
+
+	return this, others
+}
+
+// ResourceDefinition represents an ARM resource
+type ResourceDefinition struct {
+	typeName         *TypeName
+	spec             *TypeName
+	status           *TypeName
+	isStorageVersion bool
+	description      *string
+}
+
+// assert that ResourceDefinition implements TypeDefiner
+var _ TypeDefiner = &ResourceDefinition{}
+
+func (definition *ResourceDefinition) Name() *TypeName {
+	return definition.typeName
+}
+
+func (definition *ResourceDefinition) Type() Type {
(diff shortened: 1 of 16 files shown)


hack/generator/pkg/astmodel/resource.go
```diff
@@ -0,0 +1,131 @@
...
+
+// assert that ResourceDefinition implements TypeDefiner
+var _ TypeDefiner = &ResourceDefinition{}
+
+func (definition *ResourceDefinition) Name() *TypeName {
+	return definition.typeName
+}
+
+func (definition *ResourceDefinition) Type() Type {
+	return definition.spec // TODO?????
```
Porges, line 58: Not an issue for now since we don’t yet set the `status` field but I will be doing so later.

One fix for this is that we remove `Type()` from `TypeDefiner` and instead lift up the two functions we need (`References(TypeName)` and `RequiredImports` into the `TypeDefiner` interface). These are then straightforward to implement by multiplexing over `spec`/`status` here.

hack/generator/pkg/astmodel/resource.go
```diff
@@ -0,0 +1,131 @@
...
+
+	var specName *TypeName
+	if specType != nil {
+		specName = NewTypeName(name.PackageReference, name.Name()+"Spec")
+		specDef, specOthers := specType.CreateDefinitions(specName, idFactory)
+
+		others = append(append(others, specDef), specOthers...)
+	} else {
+		panic("spec must always be provided")
+	}
```
theunrepentantgeek, line 26: Suggest inverting the condition, since `panic()` does not return ...
```suggestion
	if specType == nil {
		panic("spec must always be provided")
	}
	
	specName = NewTypeName(name.PackageReference, name.Name()+"Spec")
	specDef, specOthers := specType.CreateDefinitions(specName, idFactory)

	others = append(append(others, specDef), specOthers...)
```

hack/generator/pkg/astmodel/resource.go
theunrepentantgeek: Removing `Type()` from `TypeDefiner` _sounds_ like it reduces the number of abstractions we need to keep track of at once when working on the code, and therefore might be a good idea. #possiblymistaken

hack/generator/pkg/astmodel/resource.go
```diff
@@ -0,0 +1,131 @@
...
+
+func (definition *ResourceDefinition) Type() Type {
+	return definition.spec // TODO?????
+}
+
+func (definition *ResourceDefinition) WithIsStorageVersion(isStorageVersion bool) *ResourceDefinition {
+	result := *definition
+	result.isStorageVersion = isStorageVersion
+	return &result
+}
```
theunrepentantgeek, line 65: The `WithXYZ()` syntax is working well, but given this is a boolean parameter, elsewhere I've used `MarkAsXYZ` instead.

```suggestion
func (definition *ResourceDefinition) MarkAsStorageVersion() *ResourceDefinition {
	result := *definition
	result.isStorageVersion = true
	return &result
}
```

We might not even need `MarkAsNonStorageVersion()`.

hack/generator/pkg/astmodel/struct_definition.go
```diff
@@ -10,137 +10,65 @@ import (
...
-	return definition.isResource
+// NewStructDefinition creates a new StructDefinition
+func NewStructDefinition(typeName *TypeName, structType *StructType) *StructDefinition {
+	return &StructDefinition{typeName, structType, nil}
 }
 
-// Ensure StructDefinition implements TypeDefiner interface correctly
-var _ TypeDefiner = (*StructDefinition)(nil)
+// ensure StructDefinition is a TypeDefiner
+var _ TypeDefiner = &StructDefinition{}
```
theunrepentantgeek, line 26: Why are we changing our approach here? I thought the `(*Type)(nil)` approach was idiomatic for Go.

hack/generator/pkg/astmodel/type.go
```diff
@@ -29,7 +29,7 @@ type Type interface {
 	// that also must be included in the output.
 	//
 	// [isResource is only relevant to struct types and identifies if they are a root resource for Kubebuilder]
```
theunrepentantgeek, line 31: Comment needs updating - `isResource` is no longer a parameter.

hack/generator/pkg/astmodel/struct_definition.go
Porges: it’s harder to type!

hack/generator/pkg/astmodel/resource.go
Porges: Reworked method a bit more than this, factored out common code.

hack/generator/pkg/astmodel/resource.go
Porges: Yes, I won’t do it in this PR though.

Porges: Expanded comment to include this info.

hack/generator/pkg/astmodel/resource.go
```diff
@@ -15,22 +15,23 @@ func CreateResourceDefinitions(name *TypeName, specType *StructType, statusType
...
+	}
 
-		others = append(append(others, specDef), specOthers...)
-	} else {
+	var specName *TypeName
+	if specType == nil {
 		panic("spec must always be provided")
+	} else {
+		specName = defineStruct("Spec", specType)
 	}
```
theunrepentantgeek, line 30: Since the first branch of the **if** contains a panic, you could get rid of the **else** and move the variable declaration down below the test. #extremepedantmode #trivialnit

PR #160, opened 2020-06-28
Improve the IdentifierFactory to make it easy to deal with plurals

### LINKED ISSUE #141: Deal with plurality in resource names
For example, in `batch_accounts_types_gen.go` the type currently generated is 
```
// +kubebuilder:object:root=true
type BatchAccounts struct {
    ....
}
```

It has a `BatchAccountsSpec`, etc -- these should not be plural.

You can see a similar problem on the VMSS resources as well, for example: `virtual_machine_scale_sets_types_gen.go` has `VirtualMachineScaleSetsSpec` which should probably be `VirtualMachineScaleSetSpec` like it is in the handcrafted code

theunrepentantgeek:
Enhances the identifier factory with the ability to modify parts of an identifier when they appear separated by understores (`_`), allowing to rename, for example, `BatchAccounts` to `BatchAccount` throughout.

I tried a handful of other approaches, but they all over-triggered, so I went for a compromise that should allow us to get the naming we want without too many false positives.

Resolves #141

### ToDo
- [ ] Check for false positives

### Beyond Compare Screenshots

Filename changes (`master` on the left, this PR on the right):
image

Sample changes within files:
image

image

### DIFF
--- a/hack/generator/pkg/astmodel/identifier_factory.go
+++ b/hack/generator/pkg/astmodel/identifier_factory.go
@@ -228,5 +228,7 @@ func createWholeRenames() map[string]string {
 func createPartRenames() map[string]string {
 	return map[string]string{
 		"batchAccounts": "batchAccount",
+		"virtualMachineScaleSets": "virtualMachineScaleSet",
+		"virtualMachines": "virtualMachine",
 	}
 }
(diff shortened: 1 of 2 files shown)


theunrepentantgeek:
After discussion, we've agreed a different approach to solving this problem - will close this PR and put out a new one when the new code is ready.

PR #163, opened 2020-07-01
Resolve some build errors in generated code

matthchr:
Fix a few issues with enum value generation
  - Values with characters which are disallowed in Go identifiers ("*"
    in particular) were generating names which were identical to the
    enum type name, causing compiler errors.
  - Boolean enums (a somewhat weird concept) didn't correctly prune
    quotes around the values.
  - Add enum tests covering the above cases.

Generate DeepCopy functions in Makefile gen-arm


hack/generator/pkg/jsonast/jsonast.go
```diff
@@ -235,7 +235,17 @@ func enumHandler(ctx context.Context, scanner *SchemaScanner, schema *gojsonsche
...
+
+		// Some specs include boolean enums with quotes around the boolean literals
+		if baseType == astmodel.BoolType {
+			v = vTrimmed
+		}
+
+		// TODO: This is a bit of a hack as we don't have a way to handle this generically right now
+		// TODO: for an arbitrary non-renderable character
+		// use vTrimmed as seed for identifier as it doesn't have quotes surrounding it
+		id := scanner.idFactory.CreateIdentifier(vTrimmed, astmodel.Exported)
```
theunrepentantgeek, line 248: Odd that you need this - the `identifierFactory` includes a regex to strip out non-word characters:

```
var filterRegex = regexp.MustCompile(`[\W_]`)
```

We should probably simplify your change here by enhancing the regex; the goal of the factory is to hide away all that complexity.

matthchr: Removing the quotes here is to get the "replace" bit to be less awkward.

I could've said: `"\"*\""` is mapped to `star` but that seems icky too.

theunrepentantgeek: I've dug into it a little and now I understand the problem; it is a bit icky. I'm happy enough with what you've got here, not need to change it.

PR #167, opened 2020-07-06
Use github.com/pkg/errors to get stack traces from generation errors

babbageclunk:
Replaces all uses of `fmt.Errorf` with either `errors.New`/`errors.Errorf`, or `errors.Wrapf` where an existing error was being wrapped with %w. This means that any errors originating or wrapped in our code will be annotated with a stack trace that can be included when the error is logged at the top level, which will make it simpler to track down problems.

I checked for but didn't find any uses of `errors.Is` or `errors.As` to update to use `pkg/errors` instead.

As a driveby I also replaced uses of `'%s'` or `'%v'` with `%q`.

### DIFF
--- a/hack/generator/pkg/astmodel/package_definition.go
+++ b/hack/generator/pkg/astmodel/package_definition.go
@@ -73,7 +73,7 @@ func emitFiles(filesToGenerate map[string][]TypeDefiner, outputDir string) error
 		genFile := NewFileDefinition(&defs[0].Name().PackageReference, defs...)
 		outputFile := filepath.Join(outputDir, fullFileName)
 
-		klog.V(5).Infof("Writing '%s'\n", outputFile)
+		klog.V(5).Infof("Writing %q\n", outputFile)
 
 		err := genFile.SaveToFile(outputFile)
 		if err != nil {
--- a/hack/generator/pkg/astmodel/package_definition.go
+++ b/hack/generator/pkg/astmodel/package_definition.go
@@ -7,11 +7,11 @@ package astmodel
 
 import (
 	"bytes"
-	"fmt"
 	"io/ioutil"
 	"path/filepath"
 	"text/template"
 
+	"github.com/pkg/errors"
 	"k8s.io/klog/v2"
 )
 
@@ -77,7 +77,7 @@ func emitFiles(filesToGenerate map[string][]TypeDefiner, outputDir string) error
 
 		err := genFile.SaveToFile(outputFile)
 		if err != nil {
-			return fmt.Errorf("error saving definitions to file '%v'(%w)", outputFile, err)
+			return errors.Wrapf(err, "error saving definitions to file %q", outputFile)
 		}
 	}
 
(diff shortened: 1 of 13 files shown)


hack/generator/pkg/astmodel/package_definition.go
```diff
@@ -77,7 +77,7 @@ func emitFiles(filesToGenerate map[string][]TypeDefiner, outputDir string) error
 
 		err := genFile.SaveToFile(outputFile)
 		if err != nil {
-			return fmt.Errorf("error saving definitions to file '%v'(%w)", outputFile, err)
+			return errors.Wrapf(err, "error saving definitions to file %q", outputFile)
```
theunrepentantgeek, line 80: This doesn't include anything about the nested error in the main message; won't that mean that someone debugging an issue has to explicitly dig for the piece of information they need?

To illustrate, I think this is better:
```
error saving definitions to file "demo.txt" (no space on disk)
```
than
```
error saving definitions to file "demo.txt"
```

theunrepentantgeek: Same issue with your other changes, I think. (BTW - I didn't know about `%q`, that's nice.)

hack/generator/pkg/astmodel/resource.go
```diff
@@ -92,7 +94,7 @@ func (definition *ResourceDefinition) AsDeclarations(codeGenerationContext *Code
 
 	packageName, err := codeGenerationContext.GetImportedPackageName(MetaV1PackageReference)
 	if err != nil {
-		panic(fmt.Errorf("resource definition for %s failed to import package: %w", definition.typeName, err))
+		panic(errors.Wrapf(err, "resource definition for %s failed to import package", definition.typeName))
```
theunrepentantgeek, line 97: This error message seems to be an incomplete sentence, perhaps rephrase?

hack/generator/pkg/codegen/code_generator.go
```diff
@@ -39,12 +39,12 @@ type CodeGenerator struct {
 func NewCodeGenerator(configurationFile string) (*CodeGenerator, error) {
 	config, err := loadConfiguration(configurationFile)
 	if err != nil {
-		return nil, fmt.Errorf("failed to load configuration file '%v' (%w)", configurationFile, err)
+		return nil, errors.Wrapf(err, "failed to load configuration file '%v'", configurationFile)
```
theunrepentantgeek, line 42: Should this change to use `%q` as well?
(Applies to others in this file as well as this one)

hack/generator/pkg/astmodel/package_definition.go
babbageclunk: `Wrapf` doesn't need a formatting marker, it always includes the details of the wrapped error - the wrapped message coming from that example will be `error saving definitions to file "demo.txt": no space on disk`.

hack/generator/pkg/codegen/code_generator.go
babbageclunk: Ooh, yup - I'll do another pass to make sure I've got all the `'%v'` and `'%s'` cases.

hack/generator/pkg/astmodel/resource.go
babbageclunk: I think it reads alright with the error on the end? So maybe `resource definition for ThisType failed to import package: error reading file "blah.json"`?

hack/generator/pkg/astmodel/package_definition.go
theunrepentantgeek: That works for me. :+1:

PR #169, opened 2020-07-08
Use singular names for resources

theunrepentantgeek:
Uses the library [*inflection*](https://github.com/jinzhu/inflection) to ensure that the names of resources are given in singular form (as suits types) instead of plural forms (as used by ARM).

| Old Name | New Name |
| -- | -- |
| Accounts | Account |
| BatchAccounts | BatchAccount |
| BatchAccountsCertificates | BatchAccountsCertificate |
| VirtualMachineScaleSets | VirtualMachineScaleSet |
| Vaults | Vault |

### DIFF
--- a/hack/generator/pkg/jsonast/jsonast.go
+++ b/hack/generator/pkg/jsonast/jsonast.go
@@ -467,6 +467,10 @@ func generateDefinitionsFor(
 		return nil, err
 	}
 
+	if isResource {
+		typeName = typeName.Singular()
+	}
+
 	// see if we already generated something for this ref
 	if _, ok := scanner.findTypeDefinition(typeName); ok {
 		return typeName, nil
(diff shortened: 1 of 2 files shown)


hack/generator/pkg/jsonast/jsonast.go
```diff
@@ -51,6 +51,9 @@ type (
 // findTypeDefinition looks to see if we have seen the specified definition before, returning its definition if we have.
 func (scanner *SchemaScanner) findTypeDefinition(name *astmodel.TypeName) (astmodel.TypeDefiner, bool) {
 	result, ok := scanner.definitions[*name]
+	if !ok || result == nil {
+		result, ok = scanner.definitions[*name.Singular()]
```
Porges, line 55: I don’t much like this fallback, I’d rather wrap the call to `findTypeDefinition` with a check for `isResource`

Porges: e.g. something like this around line 466; it will also get rid of the extra removal added on 507

```golang
if isResource {
    typeName = toResourceTypeName(typeName)
}
```

theunrepentantgeek:
> Any chance you'd consider using https://github.com/gobuffalo/flect rather than https://github.com/jinzhu/inflection? Kubebuilder uses the former.

I'm not wedded to _inflect_, it just seemed a reasonable choice when I did some searching. Happy to switch over.

hack/generator/pkg/jsonast/jsonast.go
theunrepentantgeek: I'll give that a try.

theunrepentantgeek: Done; it's cleaner, and it works too. #alwaysimportant

theunrepentantgeek:
I've made the changes requested by @devigned , so will squash/merge this.

babbageclunk:
Shouldn't there be changes to `go.mod`/`go.sum` for the new dependency?

PR #170, opened 2020-07-09
Omit types that aren't referred to from a resource

### LINKED ISSUE #114: Generator can produce types which are not top level types and which also are not used
golden file which reproduces this issue is:

AllOf_generates_wrapper_type.json
```
{
    "$comment": "Check that the AllOf handler correctly generates a union type. Note that this will also generate the standalone types Foo, Bar, and Baz which might actually be a bug (see discussion: " ,
    "id": "https://test.test/schemas/2020-01-01/test.json",
    "$schema": "http://json-schema.org/draft-04/schema#",
    "title": "Test",
    "type": "object",
    "allOf": [
        {
            "$ref": "#/definitions/Foo"
        },
        {
            "$ref": "#/definitions/Bar"
        },
        {
            "$ref": "#/definitions/Baz"
        }
    ],
    "definitions": {
        "Foo": {
            "type": "object",
            "properties": {
                "name": {
                    "type": "string"
                }
            }
        },
        "Bar": {
            "type": "object",
            "properties": {
                "size": {
                    "type": "integer"
                }
            },
            "required": [
                "size"
            ]
        },
        "Baz": {
            "type": "object",
            "properties": {
                "enabled": {
                    "type": "boolean"
                }
            },
            "required": [
                "enabled"
            ]
        }
    }
}
```

The issue is, the generator doesn't make a distinction between types that are referenced or are not referenced. Once it visits a type in `definitions` it adds a definition for it, even if it visited that type as part of building an `allOf` or `anyOf` type which actually is going to use the types properties but not the type itself.

babbageclunk:
Not all of the type definitions that are generated end up being referred to (directly or indirectly) from a resource. This can be because they are combined into one type by a use of the `AllOf` operator or because types that referred to them have been excluded by filtering.

Add a pass to the generator to strip these out. `Type.References` is updated to return a set of types, then  `astmodel.StripUnusedTypes` collects all of the resource definitions as roots and reverses the references to make a map of referrers for each type. It uses that referrer map to exclude any type that can't be connected back to any of the roots.

Fixes #114


hack/generator/pkg/astmodel/strip_unused.go
```diff
@@ -0,0 +1,177 @@
...
+		result = result.Add(name)
+	}
+	return result
+}
+
+// Add includes the passed name in the set and returns the updated
+// set, so that adding can work for a nil set - this makes it more
+// convenient to add to sets kept in a map (in the way you might with
+// a map of slices).
+// TODO(babbageclunk): I like the idea of making the zero element
```
babbageclunk, line 30: I'm not sure about this - it makes the map update code pretty nice, but might be too confusing.

hack/generator/pkg/astmodel/optional_type.go
```diff
@@ -39,9 +39,9 @@ func (optional *OptionalType) RequiredImports() []*PackageReference {
 	return optional.element.RequiredImports()
 }
 
-// References is true if it is this type or the 'element' type references it
-func (optional *OptionalType) References(d *TypeName) bool {
-	return optional.element.References(d)
+// References is the type that this type refers to directly.
```
theunrepentantgeek, line 42:
```suggestion
// References returns the set of types that our underlying type refers to directly.
```

hack/generator/pkg/astmodel/strip_unused.go
```diff
@@ -0,0 +1,177 @@
...
+func (ts TypeNameSet) Contains(val TypeName) bool {
+	if ts == nil {
+		return false
+	}
+	_, found := ts[val]
+	return found
+}
+
+// SetUnion returns a new set with all of the names in s1 or s2.
+func SetUnion(s1, s2 TypeNameSet) TypeNameSet {
```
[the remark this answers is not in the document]
theunrepentantgeek: You could also do some shortcutting - assuming `TypeNameSet` becomes immutable (paraphrased):

``` go
if s1 == nil {
    return s2
}
```

hack/generator/pkg/astmodel/strip_unused.go
```diff
@@ -0,0 +1,177 @@
...
+	definitions []TypeDefiner,
+) ([]TypeDefiner, error) {
+	// Build a referrers map for each type.
+	referrers := make(map[TypeName]TypeNameSet)
+
+	for _, def := range definitions {
+		for reference := range def.Type().References() {
+			name := def.Name()
+			if name == nil {
+				klog.V(5).Infof("nil name for %#v", def)
```
theunrepentantgeek, line 88: Should you **continue** if this happens?

hack/generator/pkg/astmodel/strip_unused.go
```diff
@@ -0,0 +1,177 @@
...
+	return result
+}
+
+// StripUnusedDefinitions removes all types that aren't in roots or
+// referred to by the types in roots, for example types that are
+// generated as a byproduct of an allOf element.
+func StripUnusedDefinitions(
+	roots TypeNameSet,
+	definitions []TypeDefiner,
+) ([]TypeDefiner, error) {
```
theunrepentantgeek, line 80: Perfect method signature for turning into a `PipelineStage` (see #171)

hack/generator/pkg/astmodel/strip_unused.go
```diff
@@ -0,0 +1,177 @@
...
+		for reference := range def.Type().References() {
+			name := def.Name()
+			if name == nil {
+				klog.V(5).Infof("nil name for %#v", def)
+			}
+			referrers[reference] = referrers[reference].Add(*def.Name())
+		}
+	}
+
+	klog.V(0).Infof("definitions: %d, roots: %d, referrers: %d", len(definitions), len(roots), len(referrers))
```
theunrepentantgeek, line 94: Perhaps lower this to `V(4)` or `V(5)`

hack/generator/pkg/astmodel/strip_unused.go
```diff
@@ -0,0 +1,177 @@
...
+
+// CollectResourceDefinitions returns a TypeNameSet of all of the
+// resource definitions in the definitions passed in.
+func CollectResourceDefinitions(definitions []TypeDefiner) TypeNameSet {
+	resources := make(TypeNameSet)
+	for _, def := range definitions {
+		if _, ok := def.(*ResourceDefinition); ok {
+			if def.Name() == nil {
+				klog.V(5).Infof("nil name for %#v", def)
+			}
```
theunrepentantgeek, line 118: Does this ever happen? Perhaps we should modify `.Name()` to make the check and panic if it does.

hack/generator/pkg/astmodel/strip_unused.go
```diff
@@ -0,0 +1,177 @@
...
+		if c.checkWithPath(referrer, path) {
+			// If our referrer is connected to a root, then we are
+			// too - track that in memo. Parent callers will
+			// record for themselves, so we don't need to store
+			// all of path
+			c.memo = c.memo.Add(name)
+			return true
+		}
+	}
+	path.Remove(name)
```
theunrepentantgeek, line 168: If you make `TypeNameSet` immutable, you don't need to clean up - just use a different variable name above to capture the larger path.

hack/generator/pkg/astmodel/struct_type.go
```diff
@@ -80,18 +80,16 @@ func (structType *StructType) RequiredImports() []*PackageReference {
 	return result
 }
 
-// References this type has to the given type
-func (structType *StructType) References(d *TypeName) bool {
-
+// References returns the combination of all the types the fields refer to.
```
theunrepentantgeek, line 83:
```suggestion
// References returns the set of all the types referred to by any field.
```

hack/generator/pkg/astmodel/struct_type.go
```diff
@@ -80,18 +80,16 @@ func (structType *StructType) RequiredImports() []*PackageReference {
...
+func (structType *StructType) References() TypeNameSet {
+	var results TypeNameSet
 	for _, field := range structType.fields {
-		if field.FieldType().References(d) {
-			return true
+		for ref := range field.FieldType().References() {
+			results = results.Add(ref)
```
theunrepentantgeek, line 88: Since `References()` returns a `TypeNameSet`, you could make this just

```
var results TypeNameSet
for _, field := range structType.fields {
    results = results.Union(field.FieldType().References()
}
```

hack/generator/pkg/astmodel/struct_type.go
```diff
@@ -80,18 +80,16 @@ func (structType *StructType) RequiredImports() []*PackageReference {
...
 		}
 	}
-
-	// For now, not considering functions in references on purpose
-
-	return false
+	// Not collecting types from functions deliberately.
```
theunrepentantgeek, line 91: Ok for now, but when we start generation conversion functions, I think we'll need to do this.

hack/generator/pkg/astmodel/type_name.go
```diff
@@ -52,9 +52,12 @@ func (typeName *TypeName) AsType(codeGenerationContext *CodeGenerationContext) a
 	return ast.NewIdent(typeName.name)
 }
 
-// References indicates whether this Type includes any direct references to the given Type
-func (typeName *TypeName) References(d *TypeName) bool {
-	return typeName.Equals(d)
+// References returns a set containing this type name.
+func (typeName *TypeName) References() TypeNameSet {
```
theunrepentantgeek, line 56: A convention I like for these kinds of methods is to prefix with `As`  or `To` (usually _As_ 😁 )

```suggestion
func (typeName *TypeName) References() AsTypeNameSet {
```

hack/generator/pkg/astmodel/strip_unused.go
matthchr: It seems surprising to me that this is even possible?

matthchr: Actually from reading below it basically can't be? Since you do `*def.Name()` even if this if is hit, and that'd be a nil dereference?

hack/generator/pkg/astmodel/strip_unused.go
matthchr: Although as mentioned above I don't actually think this ever happens, because if it did the line below (`*def.Name()`) would panic?

hack/generator/pkg/astmodel/strip_unused.go
```diff
@@ -0,0 +1,177 @@
...
+			referrers[reference] = referrers[reference].Add(*def.Name())
+		}
+	}
+
+	klog.V(0).Infof("definitions: %d, roots: %d, referrers: %d", len(definitions), len(roots), len(referrers))
+
+	checker := newConnectionChecker(roots, referrers)
+	var newDefinitions []TypeDefiner
+	for _, def := range definitions {
+		if def.Name() == nil {
```
matthchr, line 99: Is this actually possible? (Same question as elsewhere)

hack/generator/pkg/astmodel/strip_unused.go
```diff
@@ -0,0 +1,177 @@
...
+			klog.V(5).Infof("nil name for definition %#v", def)
+			continue
+		}
+		if checker.connected(*def.Name()) {
+			newDefinitions = append(newDefinitions, def)
+		}
+	}
+	return newDefinitions, nil
```
matthchr, line 107: shouldn't this `newDefinitions` be called something like: `usedDefinitions` or `inUseDefinitions` or something?

It's a bit confusing as it is because from inside this function it almost reads like we're returning "new things to define" when in reality we've actually filtered stuff out of the list we started with. I think changing this varname would help with that.

hack/generator/pkg/astmodel/strip_unused.go
```diff
@@ -0,0 +1,177 @@
...
+		roots:     roots,
+		referrers: referrers,
+	}
+}
+
+type connectionChecker struct {
+	roots     TypeNameSet
+	referrers map[TypeName]TypeNameSet
+
+	// memo tracks results for typenames we've already seen.
```
matthchr, line 136: This comment is phrased a bit weirdly -- `memo` is tracking types which we have confirmed are rooted to the `roots` collection, but the way this is written makes it sound like we're caching both positive and negatives.

Part of me likes the name `memo` (since this set is what you're using to perform memoization of your results), but part of me wonders if a name that more closely describes _how a type gets in here_ would be clearer - `connectedTypes`, `knownConnectedTypes` or something along those lines?

hack/generator/pkg/astmodel/strip_unused.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+	path.Remove(name)
+
+	// Note: We can't memoise the negative result here because we
+	// don't know whether the search failed because it would have
+	// included a cycle. It's possible for this call to fail but a
+	// search taking a different path to this node still to succeed
+	// and prove that this is connected to a root. (It took me longer
+	// than I'd like to understand this.)
+	return false
```
matthchr, line 176: I'm not sure I follow this either -- might sync up with you offline on it.

hack/generator/pkg/astmodel/strip_unused.go
```diff
@@ -0,0 +1,177 @@
...
+// referred to by the types in roots, for example types that are
+// generated as a byproduct of an allOf element.
+func StripUnusedDefinitions(
+	roots TypeNameSet,
+	definitions []TypeDefiner,
+) ([]TypeDefiner, error) {
+	// Build a referrers map for each type.
+	referrers := make(map[TypeName]TypeNameSet)
+
+	for _, def := range definitions {
```
matthchr, line 84: I'm sorta surprised that you bother with this inversion (references -> referrers).

Is there a reason you don't just walk the roots, find the set of all types reachable from the roots, and then just return that? I _think_ what you're doing here and the "walk forward from roots" arrive at the same conclusion but walking forward from roots seems simpler to me.

hack/generator/pkg/astmodel/type_name.go
matthchr: Confused on this suggestion @theunrepentantgeek - are you suggesting to change the name of `TypeNameSet` to `AsTypeNameSet`? That seems weird to me.

Did you maybe mean `AsReferences()` or something? (Note I am not sure I love that either but I think it makes more sense than suggested change diff I'm seeing above)

hack/generator/pkg/jsonast/testdata/AllOf/AllOf_generates_wrapper_type.golden
```diff
@@ -3,25 +3,6 @@
 // Code generated by k8s-infra-gen. DO NOT EDIT.
 package v20200101
 
-//Generated from: https://test.test/schemas/2020-01-01/test.json#/definitions/Bar
```
matthchr, line 6: Nice

hack/generator/pkg/astmodel/strip_unused.go
```diff
@@ -0,0 +1,177 @@
...
+	return result
+}
+
+// Add includes the passed name in the set and returns the updated
+// set, so that adding can work for a nil set - this makes it more
+// convenient to add to sets kept in a map (in the way you might with
+// a map of slices).
+// TODO(babbageclunk): I like the idea of making the zero element
+// useful so a map works nicely, and the analogue with appending to a
+// slice, but the mutating and returning behaviour might be too weird.
```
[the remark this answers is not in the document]
babbageclunk: I did this and tried it out, unfortunately it makes generation on my machine go from 23s (mutable sets) to 1m31s (immutable sets) - the original with no stripping was 13s. 

Also, talking with Matt a bit more I think he's right about walking from the roots down rather than from the other types back to the root - it will get rid of the need to flip references and avoid any need for memoisation. But it would essentially end up making one big set with all the connected types, and building an immutable set of 22,000 items one at a time is going to really make the quadratic behaviour more apparent. I'll keep the changes in a branch for reference though.

hack/generator/pkg/astmodel/type_name.go
theunrepentantgeek: I did say I was tired. 😁 

I must have reverted to C style "type name" thinking ... 'cause I though the method was named `TypeNameSet` and I though `AsTypeNameSet` would be a better name. #sigh

hack/generator/pkg/astmodel/strip_unused.go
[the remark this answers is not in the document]
babbageclunk: I didn't do that because it's confusing when the sets are mutable but methods return the set - I think of union as a symmetrical operation but with mutation it's not. 
I agree, that confusion goes away when the sets are immutable

hack/generator/pkg/astmodel/strip_unused.go
babbageclunk: Oh, that was left over from my debugging - I've removed it.

hack/generator/pkg/astmodel/strip_unused.go
babbageclunk: It should have had a continue, but I guess the fact that I didn't notice it didn't means that it doesn't happen in any of the types.

hack/generator/pkg/astmodel/one_of_json_marshal_function.go
```diff
@@ -38,7 +38,10 @@ func (f *OneOfJSONMarshalFunction) Equals(other Function) bool {
 // References indicates whether this function includes any direct references to the given type
 func (f *OneOfJSONMarshalFunction) References(name *TypeName) bool {
```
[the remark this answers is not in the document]
babbageclunk: I didn't notice this (since nothing tries to put it into a Type) - I'll add it.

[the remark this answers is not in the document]
babbageclunk: Hmm, I updated the interface and nothing changed? I guess nothing calls `Function.References` (yet)?

hack/generator/pkg/astmodel/strip_unused.go
babbageclunk: I got rid of all of these checks - dereferencing *TypeNames happens everywhere. I think we should just be passing around TypeNames, but I won't try to squeeze that into this PR.

hack/generator/pkg/astmodel/strip_unused.go
babbageclunk: Goes away with your nice simpler way.

hack/generator/pkg/astmodel/strip_unused.go
babbageclunk: Oh man, that is so much nicer.

hack/generator/pkg/astmodel/one_of_json_marshal_function.go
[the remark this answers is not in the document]
theunrepentantgeek: StructType doesn't call `References()` on functions yet, there's a comment to that effect over there. We'll need it though once we start generating conversion methods - both the ownership and the versioning work will need it.

hack/generator/pkg/astmodel/strip_unused.go
babbageclunk: I've removed these checks now.

hack/generator/pkg/astmodel/strip_unused.go
babbageclunk: This is obviated by following references in the other direction.

hack/generator/pkg/astmodel/struct_type.go
babbageclunk: I didn't do this because it ends up being quadratic, although it's probably not a big deal in the structs we deal with. (Although it wouldn't be if I just made Unions an asymmetric method, hmmm.)

hack/generator/pkg/astmodel/one_of_json_marshal_function.go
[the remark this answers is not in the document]
babbageclunk: Ok, I've updated it for now anyway.

hack/generator/pkg/astmodel/map_type.go
```diff
@@ -44,9 +44,9 @@ func (m *MapType) RequiredImports() []*PackageReference {
 	return result
 }
 
-// References this type has to the given type
-func (m *MapType) References(d *TypeName) bool {
-	return m.key.References(d) || m.value.References(d)
+// References returns all of the types the key and value types refer to.
+func (m *MapType) References() TypeNameSet {
+	return SetUnion(m.key.References(), m.value.References())
```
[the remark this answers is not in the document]
babbageclunk: I haven't done this, since I haven't made sets immutable (for reasons below).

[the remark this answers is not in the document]
theunrepentantgeek: Fair enough.

Porges:
Looks good. We should also be able to use this infrastructure to drive the file-grouping in a cleaner manner 🙂

hack/generator/pkg/astmodel/one_of_json_marshal_function.go
```diff
@@ -35,10 +35,10 @@ func (f *OneOfJSONMarshalFunction) Equals(other Function) bool {
 	return false
 }
 
-// References indicates whether this function includes any direct references to the given type
-func (f *OneOfJSONMarshalFunction) References(name *TypeName) bool {
+// References returns the set of references for the underlying struct.
+func (f *OneOfJSONMarshalFunction) References() TypeNameSet {
 	// Defer this check to the owning struct as we only refer to its fields and it
-	return f.oneOfStruct.References(name)
+	return f.oneOfStruct.References()
```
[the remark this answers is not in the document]
theunrepentantgeek: After a chat with @Porges, we concluded that functions shouldn't include their receiver in the list of references. I'll document this separately to this PR.

PR #171, opened 2020-07-09
Create pipeline for transformations of generated types

theunrepentantgeek:
In preparation for more extensive transformations, this PR introduces a pipeline abstraction for composing the required steps in an extensible manner.

Each `PipeLineStage` is a tuple, containing a description (for logging) and a transforming **func** to invoke. 
Factory methods create Pipeline stages as closures, capturing required configuration in a declarative fashion:

``` go
pipeline := []PipelineStage{
    applyExportFilters(generator.configuration),
    deleteGeneratedCode(ctx, generator.configuration.OutputPath),
}
```

Log output:
``` text
I0709 18:31:20.967457   13912 code_generator.go:72] Pipeline stage 1/2: Filter generated types
I0709 18:31:20.994453   13912 code_generator.go:72] Pipeline stage 2/2: Delete generated code from apis
I0709 18:31:21.013462   13912 code_generator.go:93] Writing output files into apis
I0709 18:31:38.295428   13912 code_generator.go:118] Completed writing 3072 files containing 22898 definitions
```


hack/generator/pkg/codegen/code_generator.go
```diff
@@ -76,9 +63,17 @@ func (generator *CodeGenerator) Generate(ctx context.Context) error {
 		return errors.Wrapf(err, "failed to walk JSON schema")
 	}
 
-	defs, err = generator.FilterDefinitions(defs)
-	if err != nil {
-		return errors.Wrapf(err, "failed to filter generated definitions")
+	pipeline := []PipelineStage{
+		applyExportFilters(generator.configuration),
+		deleteGeneratedCode(ctx, generator.configuration.OutputPath),
```
devigned, line 68: Perhaps, `ctx` would be better served to be on the action rather than the constructor of the structure.

hack/generator/pkg/codegen/pipeline_apply_export_filters.go
```diff
@@ -0,0 +1,45 @@
+package codegen
+
+import (
+	"github.com/Azure/k8s-infra/hack/generator/pkg/astmodel"
+	"github.com/Azure/k8s-infra/hack/generator/pkg/config"
+	"k8s.io/klog/v2"
+)
+
+func applyExportFilters(configuration *config.Configuration) PipelineStage {
```
matthchr, line 9: minor: Since this is actually construction an object (and not "doing" anything), wonder if it would be better to phrase the method like:
`newApplyExportFiltersPipelineStage` or something, rather than as an action like it is now.

The same question applies below to the `deleteGeneratedCode` method

hack/generator/pkg/codegen/pipeline_delete_generated_code.go
```diff
@@ -0,0 +1,162 @@
+package codegen
```
matthchr, line 1: header check failed for a few of these files

hack/generator/pkg/codegen/pipeline_delete_generated_code.go
```diff
@@ -0,0 +1,162 @@
...
+	kerrors "k8s.io/apimachinery/pkg/util/errors"
+	"os"
+	"path"
+	"path/filepath"
+	"sort"
+	"strings"
+)
+
+// deleteGeneratedCode creates a pipeline stage for cleanup of our output folder prior to generating files
+func deleteGeneratedCode(ctx context.Context, outputFolder string) PipelineStage {
```
matthchr, line 20: I like that this code was moved to another file (as it was cluttering up code_gen), but I'm not sure that I like this as a pipeline stage.

On the one hand, it works just fine.
On the other hand, it doesn't actually do any transformation on the type collection (`types`), nor does it actually use `types` at all, which is sorta a red flag to me - This function is just one really just one big side-effect.

I was thinking of the pipeline as this sorta functional-programming style pipeline of pure functions (which `exportFilters` is, and I think @babbageclunk's PR would be too), and this violates that purity.

matthchr: Put maybe more succinctly, I was thinking about the pipeline as operating on `types`, like you have it here, and cleaning up the filesystem doesn't operate on types, it operates on the file system.

hack/generator/pkg/codegen/pipeline_apply_export_filters.go
theunrepentantgeek: I chose this method name to make the pipeline more declarative, so the naming does look a bit odd in this location; don't want to change it, but will add a comment to clarify.

hack/generator/pkg/codegen/code_generator.go
theunrepentantgeek: Agreed, will change.

hack/generator/pkg/codegen/pipeline_delete_generated_code.go
theunrepentantgeek: I confess that this is a bit klugy because it doesn't actually change the set of types. Partly did this to have more than one pipeline stage (as a demo), partly because moving the code out of `code_generator.go` made sense, and partly to defer cleanup until just prior to writing the new files.

hack/generator/pkg/codegen/code_generator.go
theunrepentantgeek: Done.

hack/generator/pkg/codegen/pipeline_apply_export_filters.go
theunrepentantgeek: Done.

hack/generator/pkg/codegen/pipeline_delete_generated_code.go
theunrepentantgeek: Fixed.

theunrepentantgeek:
I've also moved package exports into a pipeline stage, it's becoming nice and declarative:

``` go
pipeline := []PipelineStage{
	applyExportFilters(generator.configuration),
	deleteGeneratedCode(generator.configuration.OutputPath),
	exportPackages(generator.configuration.OutputPath),
}
```

hack/generator/pkg/codegen/pipeline_export_generated_code.go
```diff
@@ -0,0 +1,197 @@
+/*
```
matthchr, line 1: kicking ideas around here -- wondering if these pipeline stages should be in a sub-module (i.e. `pipeline` inside `codegen` or something)?

hack/generator/pkg/codegen/pipeline_export_generated_code.go
```diff
@@ -0,0 +1,197 @@
...
+			pkg.AddDefinition(def)
+			packages[pkgRef] = pkg
+		}
+	}
+
+	var pkgs []*astmodel.PackageDefinition
+	for _, pkg := range packages {
+		pkgs = append(pkgs, pkg)
+	}
+
```
matthchr, line 71: I think we had a "bug" in the original version of this code that this set of packages isn't sorted.

Since maps don't have a defined iteration order, this results in us generating different packages first.

I think we should add:
```
sort.Slice(pkgs, func(i int, j int) bool {
		return pkgs[i].GroupName < pkgs[j].GroupName || (pkgs[i].GroupName == pkgs[j].GroupName && pkgs[i].PackageName < pkgs[j].PackageName)
	})
```

matthchr:
I think you need to fix `jsonast_test` too by the way...

I'm currently fighting with it in my branch because it can't actually call methods from `code_generator` -- I'm wondering if we need to take the golden files tests and put them up at the `code_generator` level, and then update the `code_generator` so that it can be constructed by the tests in a way that lets us control how files are output but retains the rest of the pipeline phases... right now any new pipeline phases we add at that level aren't exposed in the golden file tests so the types we get aren't actually going to match what we do "for real", which is problematic.

hack/generator/pkg/codegen/pipeline_export_generated_code.go
theunrepentantgeek: Partitioning things into packages occurs earlier, and once that happens they're independent - so do we need to worry about which order they're written?

hack/generator/pkg/codegen/pipeline_export_generated_code.go
theunrepentantgeek: I tried moving them into a submodule, but ended up with a circular import error.

hack/generator/pkg/codegen/code_generator.go
```diff
@@ -76,212 +60,24 @@ func (generator *CodeGenerator) Generate(ctx context.Context) error {
...
-			}
-		}
-
-		count, err := pkg.EmitDefinitions(outputDir)
+	for i, stage := range pipeline {
+		klog.V(0).Infof("Pipeline stage %d/%d: %s", i+1, len(pipeline), stage.Name)
+		defs, err = stage.Action(ctx, defs)
 		if err != nil {
-			return errors.Wrapf(err, "error writing definitions into %q", outputDir)
+			return errors.Wrapf(err, "Failed during pipeline stage %s", stage.Name)
```
Porges, line 74: might as well include the index here as well if we're reporting it above

hack/generator/pkg/codegen/pipeline_export_generated_code.go
matthchr: The order which the packages are in in this collection impacts the order they're written to disk right? I think that would show up if something were say, causing a panic during package writing you'd end up with a different set of stuff written to disk every time.

I _think_ if this is ordered, you'd get determinism even there.

theunrepentantgeek: I'm not 100% convinced this is needed - but it certainly won't hurt. Done.

PR #174, opened 2020-07-14
Remove Type() from TypeDefiner

Porges:
Replace it with lift-up of `References()` into `TypeDefiner` interface.

This fixes an outstanding TODO/bug.


matthchr:
I think your submodule snuck in again

hack/generator/pkg/astmodel/file_definition.go
```diff
@@ -112,7 +112,7 @@ func assignRanks(definers []TypeDefiner, ranks map[TypeName]int) []TypeDefiner {
 	// Assign ranks
 	for _, d := range assignable {
 		rank := ranks[*d.Name()]
-		for ref := range d.Type().References() {
+		for ref := range d.References() {
 			//klog.V(0).Infof("%v: %v -> %v\n", rank, d.Name(), ref)
```
theunrepentantgeek, line 116: Should remove this log line since it's only needed for debugging.
```suggestion
```

PR #175, opened 2020-07-14
Fix issue with golden tests

matthchr:
- Add some asserts to ensure that we always have at least a few tests


hack/generator/pkg/codegen/codegen_test.go
```diff
@@ -58,14 +58,17 @@ func runGoldenTest(t *testing.T, path string) {
...
+	var pr *astmodel.PackageReference
 	var ds []astmodel.TypeDefiner
 	for _, def := range defs {
 		ds = append(ds, def)
+		if pr == nil {
+			pr = &def.Name().PackageReference
+		}
 	}
 
 	// put all definitions in one file, regardless.	// the package reference isn't really used here.
```
Porges, line 70: comment comment 🙂

matthchr: good catch

PR #177, opened 2020-07-19
Fix a bug in the type ranking algorithm 

theunrepentantgeek:
There was a bug in the type ranking algorithm that could result in different ranks depending on the iteration order of the types. Thanks @matthchr for finding this edge case. This PR fixes the bug and adds some test coverage.

### DIFF
--- a/hack/generator/pkg/astmodel/file_definition.go
+++ b/hack/generator/pkg/astmodel/file_definition.go
@@ -112,7 +112,7 @@ func assignRanks(definers []TypeDefiner, ranks map[TypeName]int) []TypeDefiner {
 	// Assign ranks
 	for _, d := range assignable {
 		rank := ranks[*d.Name()]
-		for ref := range d.Type().References() {
+		for ref := range d.References() {
 			if _, ok := ranks[ref]; !ok {
 				// Never overwrite an existing rank
 				ranks[ref] = rank + 1
--- a/hack/generator/pkg/astmodel/file_definition.go
+++ b/hack/generator/pkg/astmodel/file_definition.go
@@ -78,7 +78,7 @@ func calcRanks(definitions []TypeDefiner) map[TypeName]int {
 	for len(queue) > 0 {
 		queue = assignRanks(queue, ranks)
 		if len(queue) == lastLength {
-			// No progress made - give everything a fallback rank
+			// No progress made - give everything remaining a fallback rank
 			for _, d := range queue {
 				ranks[*d.Name()] = 10000
 			}
@@ -112,8 +112,11 @@ func assignRanks(definers []TypeDefiner, ranks map[TypeName]int) []TypeDefiner {
 	// Assign ranks
 	for _, d := range assignable {
 		rank := ranks[*d.Name()]
-		for ref := range d.References() {
-			ranks[ref] = rank + 1
+		for ref := range d.Type().References() {
+			if _, ok := ranks[ref]; !ok {
+				// Never overwrite an existing rank
+				ranks[ref] = rank + 1
+			}
 		}
 	}
 
--- a/hack/generator/pkg/astmodel/file_definition_test.go
+++ b/hack/generator/pkg/astmodel/file_definition_test.go
@@ -20,11 +20,11 @@ func Test_NewFileDefinition_GivenValues_InitializesFields(t *testing.T) {
 
 	person := NewTestStruct(
 		"Person",
-		NewStringFieldDefinition("fullName"),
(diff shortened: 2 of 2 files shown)


hack/generator/pkg/astmodel/file_definition_test.go
```diff
@@ -11,24 +11,144 @@ import (
...
+
+	ranks := calcRanks([]TypeDefiner{&top, &left, &right, &bottom})
+
+	g.Expect(ranks[*top.Name()]).To(Equal(0))
+	g.Expect(ranks[*right.Name()]).To(Equal(1))
+	g.Expect(ranks[*left.Name()]).To(Equal(1))
+	g.Expect(ranks[*bottom.Name()]).To(Equal(2))
+}
+
+func Test_CalcRanks_GivenDiamondWithBar_AssignRanksInOrder(t *testing.T) {
```
babbageclunk, line 97: I'm not sure what `DiamondWithBar` and `DiamondWithReverseBar` mean - might be worth a couple of ascii-art diagram comments?

theunrepentantgeek: Done.

hack/generator/pkg/astmodel/file_definition.go
```diff
@@ -113,7 +113,10 @@ func assignRanks(definers []TypeDefiner, ranks map[TypeName]int) []TypeDefiner {
 	for _, d := range assignable {
 		rank := ranks[*d.Name()]
 		for ref := range d.References() {
-			ranks[ref] = rank + 1
+			if _, ok := ranks[ref]; !ok {
```
matthchr, line 116: I'm late to the game but doesn't this break the topological sort-ness?

```
//         +---------+
//         |         |
//         |   top   |
//         |         |
//         ++-------++
//          |       |
//          |       |
//          v       v
// +--------++     ++--------+
// |         |     |         |
// |  left1  |     |  right  |
// |         |     |         |
// +--------++     ++--------+
//      |           |
//      |           |
//      |           |
//      v           |
// +--------++      |
// |         |      |
// |  left2  |      |
// |         |      |
// +--------++      |
//          |       |
//          |       |
//          v       v
//         ++-------++
//         |         |
//         | bottom  |
//         |         |
//         +---------+
```

First pass: `assignRanks` with just `[top]` sets `left1` and `right` to be rank 1 (top is rank0)
Second pass: `assignRanks` with `[left1, right]` will assign `left2` and `bottom` to be rank 2.
Third pass: `assignRanks` with just `left2` - which will want to assign `bottom` to rank 3 but won't because of the guard- this means we think `bottom` and `left2` are both rank 2 but really they aren't (`bottom` should be rank 3).

If you make this example bigger (say with `left3`, `left4`, etc), the topological ordering gets quite off from what it's supposed to be.

I _think_ we have two requirements:
1. The topological ordering is a valid topological ordering.
2. The topological ordering is consistent no matter the order of the inputs (because the order of the inputs can change since it's a map).

In order to maintain a true topological ordering we need to allow the last write to win I am pretty sure.

I did more investigation into my specific cases (the ones my test was reproing) and I realized that the issue in that case at least has more to do with this line:
```
               if _, ok := d.(*ResourceDefinition); ok {
			// Resources have rank 0
			ranks[*d.Name()] = 0
		}
```

The problem is that I define a spec where the resource _is_ referenced by something (this is to simulate the uber-spec which has references to all of the resources in Azure) - but in the way I did it, that entity referring to a resource also got generated, so it's actually a rank0 type and the resource is a rank1 type. I'm not sure if this is actually possible in the ARM schemas as well but it seems like since you already detect nonroots that that is sufficient and we can just remove the resource check entirely.

_With the resource check there_ I think what's happening is that the order that the rank0 types is processed in mattered, because one of the rank0 types (the resource itself) was actually supposed to be rank 1. If the resource type was processed first though (when it was at rank0), its dependencies were put with rank1, then it was moved to rank1 later.

I am pretty sure (but not 100% sure) that since order of node discovery matters for what ranks things end up at (basically it matters what path we take to get there), that in order to force goal 2 (topological sort order is consistent no matter the order of the inputs) we may need to just force sort the inputs alphabetically or something first?

theunrepentantgeek: Closing this off after a chat with @matthchr offline - technically his observations are correct, but in this case we're ordering types in a source file for improved readability, so a good enough _partial topological sort_ is sufficient. 

Independently, he may be right about **not** wanting to treat resources differently, so I'll test that.

PR #181, opened 2020-07-21
Fixes for allOfHandler

Porges:
Firstly I fixed an error-checking order problem in `generatePropertyDefinitions`; this revealed other issues that needed fixing in `allOfHandler` where we need to be able to “unpack” more `TypeDefiner` types.

I noticed this when using Ctrl-C cancellation as I would receive lots of errors about `nil` property types.


Porges:
Will pause this until after simplification PR has merged as it interferes.

Porges:
Resume!

PR #182, opened 2020-07-21
Fix TODO in generateDefinitionsFor

Porges:
Widen the argument types for `CreateResourceDefinitions` which allows removal of a TODO.

### DIFF
--- a/hack/generator/pkg/jsonast/jsonast.go
+++ b/hack/generator/pkg/jsonast/jsonast.go
@@ -488,14 +488,7 @@ func generateDefinitionsFor(
 
 	// Give the type a name:
 	if isResource {
-		if specType, ok := result.(*astmodel.StructType); ok {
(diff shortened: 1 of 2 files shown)


hack/generator/pkg/jsonast/jsonast.go
```diff
@@ -488,14 +488,7 @@ func generateDefinitionsFor(
 
 	// Give the type a name:
 	if isResource {
-		if specType, ok := result.(*astmodel.StructType); ok {
```
matthchr, line 491: I have a branch where I am introspecting resources based on their properties (and I have used this if statement to ensure that it's a struct and thus has properties).

Possibly that code needs to move -- what type(s) are the Kusto resources?

Porges: I'm going to hold off on this because my simplification PR removes this check entirely.

Porges: @matthchr take a look at #183 since this code/version is obsoleted

PR #185, opened 2020-07-22
Specification for CRD Versioning

theunrepentantgeek:
This is a specification for how to handle storage versioning with code generation.

In addition to the core specification, there are three case studies to compare and contrast the merits of alternative approaches, giving motivation to the decisions made in the spec itself.

Direct links for formatted versions for easier reading:

* [Versioning spec](https://github.com/Azure/k8s-infra/blob/spec/versioning/hack/generator/spec/versioning.md)
* [Case Study - Rolling Storage Versions](https://github.com/Azure/k8s-infra/blob/spec/versioning/hack/generator/spec/case-study-rolling-storage-versions.md)
* [Case Study - Fixed Storage Version](https://github.com/Azure/k8s-infra/blob/spec/versioning/hack/generator/spec/case-study-fixed-storage-version.md)
* [Case Study - Chained Storage Versions](https://github.com/Azure/k8s-infra/blob/spec/versioning/hack/generator/spec/case-study-chained-storage-versions.md)


hack/generator/spec/versioning.md
```diff
@@ -0,0 +1,308 @@
+# Versioning
+
+Specification for how storage versioning will operate for code generated CRD definitions.
+
+We're generating a large volume of CRD definitions based on the JSON schema definitions available for Azure Resource Manager use.
+
+## Goals
+
+**Principle of Least Surprise:** The goal of the service operator is to allow users to consumer Azure resources without having to leave the tooling they are familiar with. We therefore want to do things in the idiomatic Kubernetes fashion, so that they don't experience any nasty surprises.
```
jessemillar, line 9: Minor typo. Should be "allow users to consume Azure resources".

hack/generator/spec/versioning.md
```diff
@@ -0,0 +1,308 @@
...
+
+**Auto-generated conversions:** As far as practical, we want to autogenerating the schema for storage use along with the conversions to and from the actual ARM API versions. Hand coding all the required conversions doesn't scale across all the different Azure sevices, especially with the ongoing rate of change.
```
jessemillar, line 11: "autogenerate" instead of "autogenerating".

hack/generator/spec/versioning.md
```diff
@@ -0,0 +1,308 @@
...
+Unlike the typical situation with a hand written service operator, we don't have complete control over the schema we are publishing for custom resources - we're deriving the CRD schema from the ARM JSON schema published online. This somewhat paints us into a corner where some issues that would be easily avoided with a hand-coded schema have to be faced head on.
+
+
+## Case Studies
+
+There are three case studies that accompany this specification, each one walking through one possible solution and showing how it will perform over time.
+
+The [**Rolling Versions**](case-study-rolling-storage-versions.md) case study shows how the preferred solution adapts to changes as the Azure Resources evolve over time.
+
+The [**Fixed Version**](case-study-fixed-storage-versions.md) case study shows how the primary alternative would fare, calling out some specific problems that will occur.
```
jessemillar, line 38: This link isn't working for me. I get a 404.

hack/generator/spec/case-study-rolling-storage-versions.md
```diff
@@ -0,0 +1,936 @@
...
+    FirstName   *string
+    Id          *Guid
+    LastName    *string
+}
+
+// Hub marks this type as a conversion hub.
+func (*Person) Hub() {}
+```
+
+Every property is marked as optional. Optionality doesn't matter at this point, as we currently have only single version of the API. However, as we'll see with later versions, forward and backward compatibility issues would arise if they were not optional.
```
serbrech, line 39: that sparked something in my mind, as we are moving to protobuf to define our API with this particular benefit in mind.
I did a quick googling, and I found this : 
https://github.com/istio/operator/tree/master/common-protos

Which seem to generate the proto files that map pretty much all CRDs of k8s and istio.
just an interesting parrallel with what we are trying to achieve here.

hack/generator/spec/versioning.md
theunrepentantgeek: Fixed.

hack/generator/spec/versioning.md
theunrepentantgeek: Fixed.

hack/generator/spec/versioning.md
theunrepentantgeek: Broken link fixed.

PR #186, opened 2020-07-22
Handle MapType in allOfHandler

Porges:
At the moment we warn about not handling maps in `allOf`. We can handle them by turning them (back) into an `additionalProperties` field.

The `allOfHandler` now handles all types we expect in the schema so I have converted the unhandled cases to a hard failure.


hack/generator/pkg/codegen/testdata/AllOf/AllOf_turns_map_type_back_into_additionalProperties.golden
```diff
@@ -0,0 +1,12 @@
...
+// Licensed under the MIT license.
+// Code generated by k8s-infra-gen. DO NOT EDIT.
+package v20200101
+
+//Generated from: https://test.test/schemas/2020-01-01/test.json
+type Test struct {
+
+	// +kubebuilder:validation:Required
+	Size                 int             `json:"size"`
+	additionalProperties map[string]bool `json:"additionalProperties"`
```
babbageclunk, line 11: Should this be exported? As it stands client code couldn't construct one with a passed-in map.

hack/generator/pkg/jsonast/jsonast.go
```diff
@@ -539,35 +539,36 @@ func allOfHandler(ctx context.Context, scanner *SchemaScanner, schema *gojsonsch
...
 			}
 
-		default:
-			klog.Errorf("Unhandled type in allOf: %#v\n", concreteType)
+			return nil, errors.Errorf("couldn't find definition for: %v", concreteType)
+
+		case *astmodel.MapType:
+			if concreteType.KeyType().Equals(astmodel.StringType) {
+				// move map type into 'additionalProperties' property
+				// TODO: consider privileging this as a special property on ObjectType?
```
babbageclunk, line 561: Could you expand this a bit?

hack/generator/pkg/jsonast/jsonast.go
```diff
@@ -539,35 +539,36 @@ func allOfHandler(ctx context.Context, scanner *SchemaScanner, schema *gojsonsch
...
+					"additionalProperties",
+					"additionalProperties",
+					concreteType)
+
+				return append(properties, newProp), nil
+			}
 		}
 
-		return properties, nil
+		return nil, errors.Errorf("unable to handle type in allOf: %#v\n", st)
```
babbageclunk, line 571: Nice.

hack/generator/pkg/codegen/testdata/AllOf/AllOf_turns_map_type_back_into_additionalProperties.golden
Porges: The support for it is still incomplete. At the moment we don’t handle this properly at all and this needs to be addressed in the serialization layer, so I’ve created an issue for this: https://github.com/Azure/k8s-infra/issues/187

hack/generator/pkg/jsonast/jsonast.go
Porges, line 561:
```suggestion
				// TODO: consider privileging this as its own property on ObjectType,
				// since it has special behaviour and we need to handle it differently
				// for JSON serialization
```

hack/generator/pkg/codegen/testdata/AllOf/AllOf_turns_map_type_back_into_additionalProperties.json
```diff
@@ -0,0 +1,34 @@
...
+                "size": {
+                    "type": "integer"
+                }
+            },
+            "required": [
+                "size"
+            ]
+        }
+    }
+}
```
matthchr, line 34: minor: missing newline

hack/generator/pkg/codegen/testdata/AllOf/AllOf_turns_map_type_back_into_additionalProperties.golden
babbageclunk: Ah, ok - so we're only expecting something like this to be coming back from ARM as part of status, rather than expecting users to construct them?

hack/generator/pkg/jsonast/jsonast.go
Porges: done

hack/generator/pkg/codegen/testdata/AllOf/AllOf_turns_map_type_back_into_additionalProperties.golden
Porges: we will allow it in future but for now this doesn‘t work at all, so…

hack/generator/pkg/codegen/testdata/AllOf/AllOf_turns_map_type_back_into_additionalProperties.json
Porges, line 34:
```suggestion
}

```

PR #188, opened 2020-07-22
Make AllOf handle properties on “owning” node

### LINKED ISSUE #179: AllOf needs to handle additional properties &c
`AllOf` handling doesn’t currently handle other properties and things at the level of the current schema, it only looks at the `allOf` property. A fairly common pattern is to “inherit” a schema with `allOf` and then add more properties to it, so we need to handle this properly (or at least better).  [There’s already a `TODO` for this in the code.](https://github.com/Azure/k8s-infra/blob/cb72bcc404995c4d3415ca8c631d3cc2e63d9a71/hack/generator/pkg/jsonast/jsonast.go#L749-L750)

Porges:
This fixes #179, at least for the cases I’ve encountered so far.


hack/generator/pkg/codegen/testdata/AllOf/AllOf_handles_properties_on_owning_node.json
```diff
@@ -0,0 +1,28 @@
...
+        "Base": {
+            "type": "object",
+            "properties": {
+                "length": {
+                    "type": "integer"
+                }
+            }
+        }
+    }
+}
```
babbageclunk, line 28: Copying @matthchr's nitpick: newline

Porges, line 28:
```suggestion
}

```

Porges: Have reconfigured my VSCode now 😁

PR #189, opened 2020-07-23
Use pipeline in codegen_test

matthchr:


hack/generator/pkg/astmodel/function_test.go
```diff
@@ -30,11 +30,11 @@ func (fake *FakeFunction) AsFunc(codeGenerationContext *CodeGenerationContext, r
 }
```
matthchr, line 31: I think we missed a make all (`go fmt`) in some other PR which is where this came from

hack/generator/pkg/codegen/code_generator.go
```diff
@@ -145,23 +157,37 @@ func (loader *cancellableJSONLoader) LoaderFactory() gojsonschema.JSONLoaderFact
...
+
+	schema, err := sl.Compile(loader)
+	if err != nil {
+		return nil, errors.Wrapf(err, "error loading schema from %q", source)
+	}
+
+	return schema, nil
+}
+
+func loadSchemaIntoTypes(
```
Porges, line 177: while you’re here do you want to move this PipelineStage generator function and all its dependencies into another file? it has a lot of cruft around it for cancellation

matthchr: sure

Porges: @matthchr feels nicer 👍

PR #191, opened 2020-07-25
Update identifier_factory to handle reserved words

matthchr:
This solves an issue that we will end up having with certain words when we use them as either private variables (or function variables).

Uppercase identifiers with those same words don't have issues

### DIFF
--- a/hack/generator/pkg/astmodel/identifier_factory.go
+++ b/hack/generator/pkg/astmodel/identifier_factory.go
@@ -33,7 +33,8 @@ type IdentifierFactory interface {
 
 // identifierFactory is an implementation of the IdentifierFactory interface
 type identifierFactory struct {
-	renames map[string]string
+	renames       map[string]string
+	reservedWords map[string]string
 }
 
 // assert the implementation exists
@@ -42,7 +43,8 @@ var _ IdentifierFactory = (*identifierFactory)(nil)
 // NewIdentifierFactory creates an IdentifierFactory ready for use
 func NewIdentifierFactory() IdentifierFactory {
 	return &identifierFactory{
-		renames: createRenames(),
+		renames:       createRenames(),
+		reservedWords: createReservedWords(),
 	}
 }
 
@@ -73,6 +75,12 @@ func (factory *identifierFactory) CreateIdentifier(name string, visibility Visib
 	}
 
 	result := strings.Join(caseCorrectedWords, "")
+
+	if alternateWord, ok := factory.reservedWords[result]; ok {
+		// This is a reserved word, we need to use an alternate word
+		return alternateWord
+	}
+
 	return result
 }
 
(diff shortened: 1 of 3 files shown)


hack/generator/pkg/astmodel/identifier_factory.go
```diff
@@ -33,7 +33,8 @@ type IdentifierFactory interface {
 
 // identifierFactory is an implementation of the IdentifierFactory interface
 type identifierFactory struct {
-	renames map[string]string
+	renames       map[string]string
```
matthchr, line 36: I'm open to just using renames for this but that's case insensitive right now and in reality we only need these for the lowercase flavors.

This comes up when generating methods where I do something like:

```
func (c *TheType) ConvertToArm() *TheTypeArm {
    package := c.Package.ConvertToArm()
    ...
}
```

hack/generator/pkg/astmodel/identifier_factory.go
```diff
@@ -88,6 +96,14 @@ func createRenames() map[string]string {
 	}
 }
 
+// These are words reserved by go
+func createReservedWords() map[string]struct{} {
+	return map[string]struct{}{
+		"package": {},
```
matthchr, line 102: Should expand this list probably

hack/generator/pkg/astmodel/identifier_factory.go
```diff
@@ -73,6 +75,12 @@ func (factory *identifierFactory) CreateIdentifier(name string, visibility Visib
 	}
 
 	result := strings.Join(caseCorrectedWords, "")
+
+	if _, ok := factory.reservedWords[result]; ok {
+		// This is a reserved word, we need to create a different name
+		return result + "Var"
```
theunrepentantgeek, line 81: Can we come up with a better approach than just suffixing with `Var` ?

How about changing `reservedWords` to a `map[string]string` and hand crafting the replacement? 
Then `package` could become `pkg`, for example.

hack/generator/pkg/astmodel/identifier_factory.go
Porges: we might as well have this be a shared map as we will never mutate it

theunrepentantgeek: Identifier Factory is a singleton, so I think it's a moot point.

Porges: is it?

hack/generator/pkg/astmodel/identifier_factory.go
matthchr: Yeah, did that.

It's _really_ close to what the renames dict is already doing but not quite so I've left it it's own thing for now still (as it happens afterwards and is case sensitive).

hack/generator/pkg/astmodel/identifier_factory.go
matthchr: If by singleton you mean: `IdentifierFactory` is only constructed once, that is true.

I've left this as is as it matches the renames collection.

hack/generator/pkg/astmodel/identifier_factory.go
theunrepentantgeek: Given that there's a different context/timing, having a separate map makes sense.

hack/generator/pkg/astmodel/identifier_factory.go
```diff
@@ -88,6 +96,37 @@ func createRenames() map[string]string {
 	}
 }
 
+// These are words reserved by go
```
theunrepentantgeek, line 99:
```suggestion
// These are words reserved by go, along with our chosen substitutes
```

PR #192, opened 2020-07-27
Fix MakeRequired/MakeOptional usage with array/map

matthchr:
- Arrays and Maps are already optional - wrapping them in Optional
    causes them to end up rendered as *[] or *map which isn't actually
    what we want. This fixes the issue so that calling
    MakeRequired or MakeOptional on a property of type Array or Map will
    just modify the validation of the property and not the type.

### DIFF
--- a/hack/generator/pkg/astmodel/property_definition.go
+++ b/hack/generator/pkg/astmodel/property_definition.go
@@ -82,13 +82,13 @@ func (property *PropertyDefinition) WithValidation(validation Validation) *Prope
 
 // MakeRequired returns a new PropertyDefinition that is marked as required
 func (property *PropertyDefinition) MakeRequired() *PropertyDefinition {
-	if !property.HasOptionalType() && property.HasRequiredValidation() {
+	if !property.hasOptionalType() && property.HasRequiredValidation() {
 		return property
 	}
 
 	result := *property
 
-	if property.HasOptionalType() {
+	if property.hasOptionalType() {
 		// Need to remove the optionality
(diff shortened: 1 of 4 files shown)


hack/generator/pkg/astmodel/property_definition_test.go
```diff
@@ -231,6 +255,52 @@ func Test_PropertyDefinitionMakeOptional_WhenTypeMandatoryAndMissingValidation_R
 	g.Expect(field).NotTo(BeIdenticalTo(original))
 }
 
+func Test_PropertyDefinitionMakeOptional_WhenPropertyTypeIsArray_ReturnsSameReference(t *testing.T) {
```
devigned, line 258: These could be structured as test cases and dry up the tests a little.

hack/generator/pkg/astmodel/property_definition.go
```diff
@@ -143,13 +143,25 @@ func (property *PropertyDefinition) HasRequiredValidation() bool {
...
-// HasOptionalType returns true if the type of this property is an optioan reference to a value
+// hasOptionalType returns true if the type of this property is an optional reference to a value
 // (and might therefore be nil).
-func (property *PropertyDefinition) HasOptionalType() bool {
+func (property *PropertyDefinition) hasOptionalType() bool {
 	_, ok := property.propertyType.(*OptionalType)
 	return ok
 }
 
+// isReferenceType returns true if the type of this property is a struct
```
theunrepentantgeek, line 153:
```suggestion
// isReferenceType returns true if the type of this property is an array
```

PR #193, opened 2020-07-27
Property method WithDescription should take string

matthchr:


hack/generator/pkg/jsonast/jsonast.go
```diff
@@ -356,7 +356,13 @@ func getProperties(ctx context.Context, scanner *SchemaScanner, schema *gojsonsc
...
 
 		// add documentation
-		property = property.WithDescription(prop.Description)
+		var description string
+		if prop.Description == nil {
+			description = ""
+		} else {
+			description = *prop.Description
+		}
+		property = property.WithDescription(description)
```
babbageclunk, line 365: Maybe only do the assignment if it's non-nil? Sidesteps the (probably non-)issue of no longer being able to set description to nil.
```suggestion
		if prop.Description != nil {
			property = property.WithDescription(*prop.Description)
		}
```

matthchr: Much cleaner

PR #194, opened 2020-07-27
Make resource pluralization fixups its own pipeline stage

matthchr:
- Changing the names of resources is sensitive because other places
    (such as ownership mapping) may have a reference to the old type
    name. Promoting this step to a pipeline phase forces all the code
    updating resource names to be co-located, which is much easier to
    read than having it spread across a few different functions in
    jsonast.

### DIFF
--- a/hack/generator/pkg/codegen/code_generator.go
+++ b/hack/generator/pkg/codegen/code_generator.go
@@ -56,6 +56,7 @@ func NewCodeGeneratorFromConfig(configuration *config.Configuration, idFactory a
 func corePipelineStages(idFactory astmodel.IdentifierFactory, configuration *config.Configuration) []PipelineStage {
 	return []PipelineStage{
 		nameTypesForCRD(idFactory),
+		improveResourcePluralization(),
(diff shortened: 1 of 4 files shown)


hack/generator/pkg/codegen/code_generator.go
```diff
@@ -56,6 +56,7 @@ func NewCodeGeneratorFromConfig(configuration *config.Configuration, idFactory a
 func corePipelineStages(idFactory astmodel.IdentifierFactory, configuration *config.Configuration) []PipelineStage {
 	return []PipelineStage{
 		nameTypesForCRD(idFactory),
+		improveResourcePluralization(),
```
matthchr, line 59: will take suggestions for a better name =)

babbageclunk: How about `singularizeResourceNames`? (Or even better `singularise` although that's probably a bridge too far. ;)
Then my fix would include to changing it to `maybeSingularizeResourceNames`, which is quite a mouthful.

I quite like the vagueness of `improve` though.

PR #196, opened 2020-07-28
Make TypeName and PackageReference values

Porges:
... never pointers. This fixes a few annoyances and some gotchas with for-loops over the `Types` type.

### DIFF
--- a/hack/generator/pkg/astmodel/package_import.go
+++ b/hack/generator/pkg/astmodel/package_import.go
@@ -12,7 +12,7 @@ import (
 
 // PackageImport represents an import of a name from a package
 type PackageImport struct {
-	PackageReference PackageReference // This is used as the key in a map so can't be pointer
+	PackageReference PackageReference
 	name             *string
 }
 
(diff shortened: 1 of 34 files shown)


hack/generator/pkg/astmodel/package_import.go
```diff
@@ -57,5 +57,5 @@ func (pi *PackageImport) PackageName() string {
 
 // Equals returns true if the passed package reference references the same package, false otherwise
 func (pi *PackageImport) Equals(ref *PackageImport) bool {
-	return pi.PackageReference.Equals(&ref.PackageReference) && pi.name == ref.name
+	return pi.PackageReference.Equals(ref.PackageReference) && pi.name == ref.name
```
matthchr, line 60: Possible conflict with the PR I have out (just heads up).

Also remove the comment on line 15 of this file?

hack/generator/pkg/astmodel/type_name.go
```diff
@@ -19,27 +19,27 @@ type TypeName struct {
 }
 
 // NewTypeName is a factory method for creating a TypeName
-func NewTypeName(pr PackageReference, name string) *TypeName {
-	return &TypeName{pr, name}
+func NewTypeName(pr PackageReference, name string) TypeName {
```
matthchr, line 22: minor, in a few other places rather than using `NewX` and returning ptr I saw somebody (Bevan?) use `MakeX` -- might be worth making that distinction here too?

PR #198, opened 2020-07-28
Convert PackageImport into a value type

theunrepentantgeek:


hack/generator/pkg/astmodel/package_import.go
```diff
@@ -13,28 +13,26 @@ import (
...
 type PackageImport struct {
 	PackageReference PackageReference
-	name             *string
+	name             string
 }
 
 // NewPackageImport creates a new package import from a reference
-func NewPackageImport(packageReference PackageReference) *PackageImport {
-	return &PackageImport{
+func NewPackageImport(packageReference PackageReference) PackageImport {
```
matthchr, line 20: Use Make over New?

PR #199, opened 2020-07-28
Update makefile

matthchr:
- Add generate target to makefile
- CI build should fail if code is not fmt'ed
- Update golangci-lint version
- Add linter which checks references to for loop vars
- Fix a few doc comments

### DIFF
--- a/tools.mk
+++ b/tools.mk
@@ -38,7 +38,7 @@ $(CONVERSION_GEN): ## Build conversion-gen from tools folder.
 
 $(GOLANGCI_LINT): ## Build golangci-lint from tools folder.
 	echo $(ROOT_DIR)
-	GOBIN=$(TOOLS_BIN_DIR) $(GO_INSTALL) github.com/golangci/golangci-lint/cmd/golangci-lint@v1.26.0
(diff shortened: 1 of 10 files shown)


tools.mk
```diff
@@ -38,7 +38,7 @@ $(CONVERSION_GEN): ## Build conversion-gen from tools folder.
 
 $(GOLANGCI_LINT): ## Build golangci-lint from tools folder.
 	echo $(ROOT_DIR)
-	GOBIN=$(TOOLS_BIN_DIR) $(GO_INSTALL) github.com/golangci/golangci-lint/cmd/golangci-lint@v1.26.0
+	GOBIN=$(TOOLS_BIN_DIR) $(GO_INSTALL) github.com/golangci/golangci-lint/cmd/golangci-lint@v1.29.0
```
Porges, line 41: these version numbers should probably be in variables

matthchr: They _kindof_ are because `GOLANGCI_LINT` this is a target? They're not used outside of the context of this target.

Gonna leave this as is for now. Can always fix later.

PR #200, opened 2020-07-28
Bless "Types" by putting it in astmodel

Porges:
Also add an `Add` method that does sanity checks to ensure we don’t overwrite previously-generated types.

### DIFF
--- a/hack/generator/pkg/astmodel/type.go
+++ b/hack/generator/pkg/astmodel/type.go
@@ -34,7 +34,7 @@ type Type interface {
 // Types is the set of all types being generated
 type Types map[TypeName]TypeDefinition
 
-// Add adds a type to the set, with sanity check
+// Add adds a type to the set, with safety check that it has not already been defined
 func (types Types) Add(def TypeDefinition) {
 	key := def.Name()
 	if _, ok := types[key]; ok {
--- a/hack/generator/pkg/astmodel/type.go
+++ b/hack/generator/pkg/astmodel/type.go
@@ -31,6 +31,19 @@ type Type interface {
 	Equals(t Type) bool
 }
 
+// Types is the set of all types being generated
+type Types map[TypeName]TypeDefinition
+
+// Add adds a type to the set, with sanity check
+func (types Types) Add(def TypeDefinition) {
+	key := def.Name()
+	if _, ok := types[key]; ok {
+		panic(fmt.Sprintf("type already defined: %v", key))
+	}
+
+	types[key] = def
+}
+
 // TypeEquals decides if the types are the same and handles the `nil` case
 func TypeEquals(left, right Type) bool {
 	if left == nil {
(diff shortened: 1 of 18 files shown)


hack/generator/pkg/astmodel/type.go
```diff
@@ -31,6 +31,19 @@ type Type interface {
 	Equals(t Type) bool
 }
 
+// Types is the set of all types being generated
+type Types map[TypeName]TypeDefinition
+
+// Add adds a type to the set, with sanity check
```
theunrepentantgeek, line 37: The term "sanity check" has some baggage, best to avoid ...

```suggestion
// Add adds a type to the set, with safety check that we haven't already added that type
```

Porges: good point

Porges: updated all instances

hack/generator/pkg/astmodel/type.go
```diff
@@ -31,6 +31,19 @@ type Type interface {
...
 }
 
+// Types is the set of all types being generated
+type Types map[TypeName]TypeDefinition
+
+// Add adds a type to the set, with safety check that it has not already been defined
+func (types Types) Add(def TypeDefinition) {
+	key := def.Name()
+	if _, ok := types[key]; ok {
+		panic(fmt.Sprintf("type already defined: %v", key))
```
matthchr, line 41: Do we prefer this over err return?

hack/generator/pkg/codegen/pipeline_improve_resource_pluralization.go
```diff
@@ -16,16 +17,31 @@ func improveResourcePluralization() PipelineStage {
...
+
+			// now sort resources so that plurals will come after any singulars that
+			// already exist, so we don't conflict
+			sort.Slice(resources, func(i, j int) bool {
+				return resources[i].Name().Name() < resources[j].Name().Name()
+			})
+
+			for _, resource := range resources {
+				newTypeName := resource.Name().Singular()
+				if _, ok := result[newTypeName]; !ok {
```
matthchr, line 40: Rather than do this sorting can't you just check the provided `types`? If you `Singular()` something and it's in `types` then skip singularizing it?

I guess that would miss something weird like`Fooss` and `Foos` both being mapped to `Foo`? I'm not actually sure what happens when calling `Singular()` though so not sure that's even possible?

hack/generator/pkg/astmodel/type.go
Porges: it's not something that can be dealt with by the caller, it's a code-correctness issue

hack/generator/pkg/codegen/pipeline_improve_resource_pluralization.go
Porges: hmm, I'm not sure why I didn't think of that

PR #202, opened 2020-07-28
Base type-to-file allocation on the ReferenceGraph code

Porges:
This simplifies it greatly and it no longer gets “stuck” on my WIP branch.

### DIFF
--- a/hack/generator/pkg/astmodel/package_definition.go
+++ b/hack/generator/pkg/astmodel/package_definition.go
@@ -20,35 +20,28 @@ type PackageDefinition struct {
 	GroupName        string
 	PackageName      string
 	GeneratorVersion string
-
-	definitions []TypeDefinition
+	definitions      Types
 }
 
 // NewPackageDefinition constructs a new package definition
 func NewPackageDefinition(groupName string, packageName string, genVersion string) *PackageDefinition {
-	return &PackageDefinition{groupName, packageName, genVersion, nil}
+	return &PackageDefinition{groupName, packageName, genVersion, make(Types)}
 }
 
-func (pkgDef *PackageDefinition) Definitions() []TypeDefinition {
+func (pkgDef *PackageDefinition) Definitions() Types {
 	return pkgDef.definitions
 }
 
 // AddDefinition adds a Definition to the PackageDefinition
 func (pkgDef *PackageDefinition) AddDefinition(def TypeDefinition) {
-	pkgDef.definitions = append(pkgDef.definitions, def)
+	pkgDef.definitions.Add(def)
 }
 
 // EmitDefinitions emits the PackageDefinition to an output directory
 func (pkgDef *PackageDefinition) EmitDefinitions(outputDir string) (int, error) {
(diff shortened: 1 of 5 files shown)


hack/generator/pkg/astmodel/package_definition.go
```diff
@@ -84,131 +77,37 @@ func emitFiles(filesToGenerate map[string][]TypeDefinition, outputDir string) er
...
 
-		allocatedFileName := allocateFileName(filesReferencingType, typeToAllocate, hasPendingReferences)
-		if allocatedFileName == "" && skippedTypes > skipLimit {
-			// We've stalled during allocation of types, implying we have a cycle of mutually referencing types
-			allocatedFileName = allocateFilenameWhenStalled(typesToAllocate, filesToGenerate, len(filesReferencingType), typeToAllocate)
+	// rootsFor maps a type name to the roots that can reach it
+	rootsFor := make(map[TypeName][]TypeName)
+	for root := range graph.roots {
+		collected := make(TypeNameSet)
+		graph.CollectTypes(root, collected)
```
matthchr, line 87: That method is public here? So maybe it does need to be public?

Porges: it’s in the same package

hack/generator/pkg/astmodel/reference_graph.go
```diff
@@ -44,27 +44,39 @@ func NewReferenceGraphWithResourcesAsRoots(types Types) ReferenceGraph {
...
 // collectTypes returns the set of types that are reachable from the given
 // 'node' (and places them in 'collected').
-func (c ReferenceGraph) collectTypes(node TypeName, collected TypeNameSet) {
-	if collected.Contains(node) {
+func (c ReferenceGraph) collectTypes(depth int, node TypeName, collected ReachableTypes) {
+	if currentDepth, ok := collected[node]; ok {
 		// We can stop here - we've already visited this node.
+		// But first, see if we are at a lower depth:
+		if depth < currentDepth {
+			collected[node] = depth
```
theunrepentantgeek, line 72: If you lower the depth of `node` then you need to keep processing, and the depths of the things reachable from `node` need to be lowered too.

hack/generator/pkg/astmodel/type_definition.go
```diff
@@ -90,8 +90,8 @@ func (std *TypeDefinition) RequiredImports() []PackageReference {
 	return std.theType.RequiredImports()
 }
 
-// FileNameHint returns what a file that contains this definition (if any) should be called
+// FileNameHint returns what a file that contains this name (if any) should be called
 // this is not always used as we might combine multiple definitions into one file
```
theunrepentantgeek, line 94:
```suggestion
// this is not always used as we often combine multiple definitions into one file
```

hack/generator/pkg/astmodel/reference_graph.go
Porges: 🤔

Porges: added a test case and fix

PR #203, opened 2020-07-30
Add a pipeline stage to check for remaining AnyType instances

babbageclunk:
Raises an error with the containing packages (so they can be added to the filtering) if any are found.

Added exclusions in azure-arm.yaml for all groups/versions that fail at the moment. This gives us a list of the packages that need to be worked through.

If you want to see the specific types containing AnyType run with verbosity level 2 or higher. (Adjusted the type filtering logging
level up to 3 to avoid drowning in skipped types at the same time.)

### DIFF
--- a/hack/generator/azure-arm.yaml
+++ b/hack/generator/azure-arm.yaml
@@ -23,6 +23,117 @@ exportFilters:
     group: definitions
     name: ResourceBase
     because: this type is not used
+
+  # Exclusions for packages that currently produce types including AnyType.
+  # TODO: get rid of these, either by chasing the teams generating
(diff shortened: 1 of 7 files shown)


hack/generator/azure-arm.yaml
```diff
@@ -23,6 +23,115 @@ exportFilters:
     group: definitions
     name: ResourceBase
     because: this type is not used
+
+  # Exclusions for packages that currently produce types including AnyType
```
Porges, line 27: big TODO 🙂

babbageclunk: Good call - adding that!

Issue #162, opened 2020-06-29
We should avoid having Spec/Status in the type names where possible


Porges:
With changes for Spec/Status generation we will get "Spec" and "Status" in the struct type names as we essentially have two "hierarchies". However at the leaves we will end up on primitive/small types like enums and so on which will not differ between the two versions. We should try to unify these back to a single type and remove "Spec"/"Status" from their names.

PR #205, opened 2020-07-30
Import status information from Swagger

Porges:
Augment existing types with status information derived from Azure Swagger specs.

This information is derived from the Azure Swagger specifications. We parse the Swagger specs and look for
any actions that appear to be ARM resources (have PUT methods with types we can use and appropriate names in the
action path). Then for each resource, we use the existing JSON AST parser to extract the status type
(the type-definition part of swagger is the same as JSON Schema).

Next, we walk over all the resources we are currently generating CRDs for and attempt to locate
a match for the resource in the status information we have parsed. If we locate a match, it is
added to the Status field of the Resource type (after we have renamed all the status types to
avoid any conflicts with existing Spec types that have already been defined).

Outstanding items to be completed in separate PRs are: https://github.com/Azure/k8s-infra/issues/219


hack/generator/pkg/codegen/pipeline_augment_status.go
```diff
@@ -0,0 +1,421 @@
...
+			}
+
+			klog.Errorf("unable to produce type for %v (in %s): %v", resourceName, filePath, err)
+			continue // TODO: make fatal
+		}
+
+		if resourceType != nil {
+			if existingResource, ok := resources[resourceName]; ok {
+				if !astmodel.TypeEquals(existingResource.Type(), resourceType) {
+					klog.Errorf("RESOURCE already defined differently 😱: %v (%s)", resourceName, filePath)
```
theunrepentantgeek, line 325: 😱 😱 😱 

😉 😉 😉

hack/generator/pkg/codegen/pipeline_augment_status.go
```diff
@@ -0,0 +1,421 @@
...
+	}
+
+	if sharedErr != nil {
+		return nil, sharedErr
+	}
+
+	return schemas, nil
+}
+
+type typeExtractor struct {
```
theunrepentantgeek, line 280: Move to `type_extractor.go` or maybe `swagger_type_extractor.go`

hack/generator/pkg/jsonast/schema_abstraction_gojson.go
```diff
@@ -0,0 +1,181 @@
...
+func transformGoJSONSlice(slice []*gojsonschema.SubSchema) []Schema {
+	result := make([]Schema, len(slice))
+	for i := range slice {
+		result[i] = GoJSONSchema{slice[i]}
+	}
+
+	return result
+}
+
+func (it GoJSONSchema) URL() *url.URL {
```
theunrepentantgeek, line 38: I think idiomatic Go would suggest a more informative name for the receiver than `it` - how about `schema` ?

hack/generator/azure-arm.yaml
```diff
@@ -1,4 +1,4 @@
-schemaUrl: https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json
+schemaUrl: file:///home/george/k8s-infra/hack/generator/deploymentTemplate.json
```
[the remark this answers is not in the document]
Porges: ... once they fix it 👍

hack/generator/azure-arm.yaml
```diff
@@ -29,3 +29,567 @@ typeTransformers:
     because: NumberOrExpression is an ARM template artifact that doesn't make sense in CRDs
     target:
       name: float
+
+status:
+  schemaRoot: "./azure-rest-api-specs/specification"
+
+  # the configuration below is copied from the ARM schema repo:
+  # https://github.com/Azure/azure-resource-manager-schemas/blob/master/generator/autogenlist.ts
```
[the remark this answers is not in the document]
Porges: I'm going to be removing all the ones that don't have any customization,

hack/generator/deploymentTemplate.json
```diff
@@ -0,0 +1,6654 @@
+{
```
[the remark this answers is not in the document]
Porges: because the current one is broken; I won't include it in the real PR

hack/generator/pkg/codegen/pipeline_augment_status.go
```diff
@@ -0,0 +1,421 @@
...
+			for _, otherType := range otherTypes {
+				newTypes.Add(otherType)
+			}
+
+			found := 0
+			notFound := 0
+			for typeName, typeDef := range types {
+				if resource, ok := typeDef.Type().(*astmodel.ResourceType); ok {
+					if statusDef, ok := resourceLookup[typeName]; ok {
+						//klog.Infof("Found swagger information for %v", typeName)
```
matthchr, line 58: commented code

hack/generator/pkg/codegen/pipeline_augment_status.go
```diff
@@ -0,0 +1,421 @@
...
+			extractor := typeExtractor{
+				outputVersion: version,
+				outputGroup:   outputGroup,
+				idFactory:     idFactory,
+				cache:         cache,
+				config:        config,
+			}
+
+			for filePath, schema := range schemas {
+				err := extractor.extractTypes(ctx, filePath, schema, result.resources, result.types)
```
matthchr, line 176: I'm sometimes thrown by functions like this where the parameter is also being modified by the function and contains the output/result.

I'm not sure there's really a good alternative though for maps (for arrays you could return the stuff to append and then append it on the next line with I usually prefer)

matthchr: Also unlike C/C++ (where I see this pattern quite frequently), go doesn't have a way to annotate the function signature/parameters with `const` (or in this case, _not_) to inform callers that actually that parameter will change.

hack/generator/pkg/codegen/pipeline_name_types_for_crd.go
```diff
@@ -93,11 +94,11 @@ func nameInnerTypes(
 	visitor.VisitResourceType = func(this *astmodel.TypeVisitor, it *astmodel.ResourceType, ctx interface{}) astmodel.Type {
 		nameHint := ctx.(string)
 
-		spec := this.Visit(it.SpecType(), nameHint+"Spec")
+		spec := this.Visit(it.SpecType(), nameHint+"_Spec")
```
matthchr, line 97: The names of the types only show up in the type graph which isn't exposed in the CRDs so this isn't that big of a deal, but: why `_` + `X`? Isn't the standard idiom in Go to avoid `_` in type names?

hack/generator/pkg/config/configuration.go
```diff
@@ -149,3 +162,30 @@ func (config *Configuration) TransformType(name astmodel.TypeName) (astmodel.Typ
...
+type StatusConfiguration struct {
+	// The root URL of the status (Swagger) files (relative to this file)
+	SchemaRoot string `yaml:"schemaRoot"`
+
+	// The per-namespace configuration
+	Schemas []StatusNamespace `yaml:"schemas"`
+}
+
+// StatusNamespace provides per-namespace configuration
+type StatusNamespace struct {
```
matthchr, line 178: oh, it _is_ a namespace (I had made a comment earlier related to this)...

Maybe there's a better name for `StatusConfiguration.Schemas` then? Does `StatusConfiguration.Namespaces` work better...?

hack/generator/pkg/jsonast/schema_abstraction.go
```diff
@@ -0,0 +1,73 @@
...
+// 'go-openapi' for everything because it could handle
+// both JSON Schema and Swagger; but this is not yet done.
+type Schema interface {
+	URL() *url.URL
+	Title() *string
+	Description() *string
+
+	HasType(schemaType SchemaType) bool
+
+	// complex things
```
matthchr, line 25: Do these comments get interpreted as documentation comments for the field which they are directly above?

Not sure if there's a way to avoid that but worth considering.

tools.mk
```diff
@@ -52,4 +52,7 @@ $(GOX): ## Build gox
 	GOBIN=$(TOOLS_BIN_DIR) $(GO_INSTALL)  github.com/mitchellh/gox@v1.0.1
 
 $(GCOV2LCOV): ## Build gcov2lcov
-	GOBIN=$(TOOLS_BIN_DIR) $(GO_INSTALL) github.com/jandelgado/gcov2lcov@v1.0.2
\ No newline at end of file
+	GOBIN=$(TOOLS_BIN_DIR) $(GO_INSTALL) github.com/jandelgado/gcov2lcov@v1.0.2
+
+.PHONY: install-tools
+install-tools: $(KIND) $(KUSTOMIZE) $(CONTROLLER_GEN) $(CONVERSION_GEN) $(GOLANGCI_LINT) $(GOLINT) $(GOX) $(GCOV2LCOV)
```
matthchr, line 58: minor: missing newline

hack/generator/pkg/jsonast/schema_abstraction_openapi.go
```diff
@@ -0,0 +1,337 @@
...
+	err = swagger.UnmarshalJSON(fileContent)
+	if err != nil {
+		return swagger, errors.Wrap(err, "unable to parse swagger file")
+	}
+
+	return swagger, err
+}
+
+// PreloadCache loads the specified schema into the cache. it does not attempt to deduplicate work.
+func (fileCache *OpenAPISchemaCache) PreloadCache(filePath string) (spec.Swagger, error) {
```
matthchr, line 209: what's the advantage of having this and `fetchFileAbsolute`? I know this is exported and `fetchFileAbsolute` isn't, and `fetchFileAbsolute` has double-checked locking/is threadsafe, but that just makes `fetchFileAbsolute` a superset of this?

Would it make more sense to just expose `fetchFileAbsolute` and remove `PreloadCache`?

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "hack/generator/azure-rest-api-specs"]
+	path = hack/generator/azure-rest-api-specs
+	url = https://github.com/Azure/azure-rest-api-specs
```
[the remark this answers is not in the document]
Porges: I think it would be useful for repeatability (and situations like the one we are currently in WRT broken refs).

hack/generator/pkg/codegen/pipeline_augment_status.go
```diff
@@ -0,0 +1,421 @@
...
+
+		versionDirs, err := findVersionDirectories(rootPath)
+		if err != nil {
+			return swaggerTypes{}, errors.Wrapf(err, "unable to find version directories")
+		}
+
+		for _, versionDir := range versionDirs {
+			version := path.Base(versionDir)
+
+			schemas, err := loadSchemas(ctx, versionDir, cache)
```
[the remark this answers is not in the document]
Porges: This is a bit wonky since I reworked it with the namespacing information from the config. Originally it was loading all the files in one pass; I'm going to restore it back to that model.

hack/generator/pkg/codegen/pipeline_name_types_for_crd.go
Porges: This is something to be resolved later (#162), at the moment just using `Status` as a suffix without underscore causes some typenames to clash so this was a quick fix for that.

hack/generator/pkg/config/configuration.go
Porges: Right so we have a bit of a conflict with the ARM and Kubernetes naming here. I think I'm going to drop the ARM naming because I'm not going to end up using much of their config file.

hack/generator/pkg/jsonast/schema_abstraction_openapi.go
Porges: `fetchFileAbsolute` only lets one file be loaded at any one time, this will load as many in parallel as you want (but would duplicate work if you load duplicate filenames).

The threadsafety here is not actually needed any more since I pulled the type-extraction out as its own pass over the schema types, so I can simplify this type greatly.

hack/generator/pkg/astmodel/enum_type.go
```diff
@@ -26,6 +26,10 @@ var _ Type = (*EnumType)(nil)
 
 // NewEnumType defines a new enumeration including the legal values
 func NewEnumType(baseType *PrimitiveType, options []EnumValue) *EnumType {
+	if baseType == nil {
+		panic("baseType must be provided")
+	}
+
```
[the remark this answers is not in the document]
Porges: But we don't do this anywhere else. Maybe we should make `PrimitiveType` a value always, and do that change in another PR?

hack/generator/pkg/codegen/pipeline_augment_status.go
```diff
@@ -0,0 +1,421 @@
...
+				return filepath.SkipDir // this is a magic error
+			}
+		}
+
+		if filepath.Ext(path) != ".json" {
+			return nil
+		}
+
+		wg.Add(1)
+		go func() {
```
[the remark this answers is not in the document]
Porges: Explain on the other comment, so resolving this.

hack/generator/pkg/codegen/pipeline_augment_status.go
```diff
@@ -0,0 +1,421 @@
...
+			mutex.Lock()
+			if err != nil {
+				// first error wins
+				if sharedErr == nil {
+					sharedErr = err
+				}
+			} else {
+				schemas[path] = swagger
+			}
+			mutex.Unlock()
```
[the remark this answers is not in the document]
Porges: I think I was just being paranoid after defer scoping bit me earlier.

hack/generator/pkg/codegen/pipeline_augment_status.go
```diff
@@ -0,0 +1,421 @@
...
+
+			wg.Done()
+		}()
+
+		return nil
+	})
+
+	wg.Wait()
+
+	if err != nil {
```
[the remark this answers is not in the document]
Porges: `filePath.Walk` can return an error; I wait for all goroutines to finish before returning

hack/generator/pkg/astmodel/enum_type.go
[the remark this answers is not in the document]
Porges: Punting to a different change.

hack/generator/pkg/jsonast/schema_abstraction_openapi.go
Porges: Stripped out most of this.

hack/generator/pkg/jsonast/schema_abstraction.go
Porges: I think you have to start them with the name of the field.

hack/generator/azure-arm.yaml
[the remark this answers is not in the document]
Porges: Stripped out most of this; there's some information in there we might need later (not sure yet). If we don't it will be removed later 🙂

hack/generator/deploymentTemplate.json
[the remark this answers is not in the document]
Porges: Gone now.

hack/generator/pkg/codegen/pipeline_augment_status.go
```diff
@@ -0,0 +1,421 @@
...
+)
+
+// augmentResourcesWithStatus creates a PipelineStage to add Swagger information into the 'status' part of CRD
+func augmentResourcesWithStatus(idFactory astmodel.IdentifierFactory, config *config.Configuration) PipelineStage {
+	return PipelineStage{
+		"Add information from Swagger specs for 'status' fields",
+		func(ctx context.Context, types astmodel.Types) (astmodel.Types, error) {
+
+			if config.Status.SchemaRoot == "" {
+				klog.Warningf("no status schema root specified, not generating status types")
```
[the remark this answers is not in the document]
Porges: The CRDs could still be useful without status information?

hack/generator/pkg/codegen/pipeline_augment_status.go
```diff
@@ -0,0 +1,421 @@
...
+		}
+
+		resourceType, err := extractor.resourceTypeFromOperation(ctx, scanner, swagger, filePath, put)
+		if err != nil {
+			if err == context.Canceled {
+				return err
+			}
+
+			klog.Errorf("unable to produce type for %v (in %s): %v", resourceName, filePath, err)
+			continue // TODO: make fatal
```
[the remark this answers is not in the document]
Porges: No, still some work to do to improve this.

hack/generator/pkg/codegen/pipeline_augment_status.go
Porges: Going to leave as-is.

hack/generator/pkg/config/configuration.go
Porges: Renamed this area.

hack/generator/pkg/jsonast/schema_abstraction_openapi.go
```diff
@@ -0,0 +1,293 @@
...
+	}
+
+	return result
+}
+
+func (schema *OpenAPISchema) isRef() bool {
+	return schema.inner.Ref.GetURL() != nil
+}
+
+type filePathAndSwagger struct {
```
babbageclunk, line 191: How about `resolvedSwagger`? Or you could just return `(string, spec.Swagger, error)` from `fetchFileRelative`.

hack/generator/pkg/jsonast/jsonast.go
```diff
@@ -442,28 +448,33 @@ func refHandler(ctx context.Context, scanner *SchemaScanner, schema *gojsonschem
...
-	url := schema.ID.GetUrl()
+	url := schema.url()
 	isResource := isResource(url)
 
 	// see if we already generated something for this ref
 	if _, ok := scanner.findTypeDefinition(typeName); ok {
 		return typeName, nil
 	}
 
+	if typeName.Name() == "SwaggerCustomDynamicProperties" {
```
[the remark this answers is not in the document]
Porges: yikes! thanks

hack/generator/pkg/codegen/swagger_type_extractor.go
```diff
@@ -0,0 +1,168 @@
...
+// in the name "ResourceType".
+func inferNameFromURLPath(operationPath string) (string, string, error) {
+
+	group := ""
+	name := ""
+
+	urlParts := strings.Split(operationPath, "/")
+	reading := false
+	skippedLast := false
+	for _, urlPart := range urlParts {
```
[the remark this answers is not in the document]
Porges: that's intentional, in this case we generate `ResourceTypeDifferentTypeSomething`. This is needed for nested/child resources.

hack/generator/pkg/codegen/pipeline_augment_status.go
```diff
@@ -0,0 +1,270 @@
...
+				mutex.Lock()
+				if err != nil {
+					// first error wins
+					if sharedErr == nil {
+						sharedErr = err
+					}
+				} else {
+					schemas[filePath] = swagger
+				}
+				mutex.Unlock() // not using defer as mutex must be unlocked before wg.Done() is invoked
```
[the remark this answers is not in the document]
Porges: Nice! This made it a lot cleaner.

hack/generator/pkg/jsonast/schema_abstraction.go
```diff
@@ -0,0 +1,72 @@
...
+	"net/url"
+)
+
+// Schema abstracts over the exact implementation of
+// JSON schema we are using as we need to be able to
+// process inputs provided by both 'gojsonschema' and
+// 'go-openapi'. It is possible we could switch to
+// 'go-openapi' for everything because it could handle
+// both JSON Schema and Swagger; but this is not yet done.
+type Schema interface {
```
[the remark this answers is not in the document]
Porges: It needs to be exported to be able to invoke `scanner.GenerateDefinitions` with a `Schema`. You make `Schema` instances with the two factory methods `MakeGoJSONSchema` and `MakeOpenAPISchema`.

hack/generator/pkg/jsonast/jsonast.go
```diff
@@ -779,27 +790,14 @@ func getPrimitiveType(name SchemaType) (*astmodel.PrimitiveType, error) {
...
-
-// Extract the 'group' (here filename) of an object from the supplied schemaURL
-func groupOf(url *url.URL) (string, error) {
-	pathParts := strings.FieldsFunc(url.Path, isURLPathSeparator)
-
-	file := pathParts[len(pathParts)-1]
-	if !strings.HasSuffix(file, ".json") {
-		return "", errors.Errorf("Unexpected URL format (doesn't point to .json file)")
+func appendIfUniqueType(slice []astmodel.Type, item astmodel.Type) []astmodel.Type {
+	for _, r := range slice {
```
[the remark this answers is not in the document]
Porges: Not new on my branch, just a misleading diff 🙂

hack/generator/pkg/codegen/pipeline_augment_status.go
```diff
@@ -0,0 +1,270 @@
...
+			outputVersion: outputVersion,
+			outputGroup:   outputGroup,
+			idFactory:     idFactory,
+			cache:         cache,
+			config:        config,
+		}
+
+		err := extractor.extractTypes(ctx, schemaPath, schema, result.resources, result.otherTypes)
+		if err != nil {
+			return swaggerTypes{}, err
```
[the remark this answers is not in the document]
Porges: Yes, makes it nicer 👍

hack/generator/pkg/codegen/pipeline_augment_status.go
```diff
@@ -0,0 +1,262 @@
...
+			swaggerGroupRegex.MatchString(filePath) &&
+			swaggerVersionRegex.MatchString(filePath) {
+
+			// all files are loaded in parallel to speed this up
+			eg.Go(func() error {
+				var swagger spec.Swagger
+
+				fileContent, err := ioutil.ReadFile(filePath)
+				if err != nil {
+					return errors.Wrap(err, "unable to read swagger file")
```
matthchr, line 232: Do we need to include the path of the file in these errors?

hack/generator/pkg/jsonast/schema_abstraction_openapi.go
```diff
@@ -0,0 +1,288 @@
...
+
+// OpenAPISchemaCache is a cache of schema that have been loaded,
+// identified by file path
+type OpenAPISchemaCache struct {
+	files map[string]spec.Swagger
+}
+
+// MakeOpenAPISchemaCache creates an OpenAPISchemaCache with the initial
+// file path → spec mapping
+func MakeOpenAPISchemaCache(specs map[string]spec.Swagger) *OpenAPISchemaCache {
```
matthchr, line 37: minor: Other places I think we had `NewXxx` returns a ptr, whereas `MakeXxx` returns a non-ptr.

Here `MakeXxx` is returning a ptr.

matthchr: Same comment on a few other functions in this file

hack/generator/pkg/jsonast/schema_abstraction_openapi.go
```diff
@@ -0,0 +1,288 @@
...
+
+	path = fileURL.ResolveReference(url).Path
+	swagger, err = fileCache.fetchFileAbsolute(path)
+
+	return path, swagger, err
+}
+
+// fetchFileAbsolute fetches the schema for the absolute path specified
+func (fileCache *OpenAPISchemaCache) fetchFileAbsolute(filePath string) (spec.Swagger, error) {
+	if swagger, ok := fileCache.files[filePath]; ok {
```
matthchr, line 213: You preload the cache will all of the Swagger files, right? So in reality do we take this path (the path is in the cache) basically all the time?

hack/generator/pkg/jsonast/schema_abstraction_openapi.go
```diff
@@ -0,0 +1,288 @@
...
+		return swagger, nil
+	}
+
+	var swagger spec.Swagger
+
+	fileContent, err := ioutil.ReadFile(filePath)
+	if err != nil {
+		return swagger, errors.Wrap(err, "unable to read swagger file")
```
matthchr, line 221: Same question as before about needing file paths here in the errors?

Also, worth factoring out to a little helper function since there seem to be 2 places that do the same thing w/ the same errors?

hack/generator/pkg/jsonast/schema_abstraction_openapi.go
```diff
@@ -0,0 +1,288 @@
...
+	}
+
+	return result
+}
+
+func (schema *OpenAPISchema) isRef() bool {
+	return schema.inner.Ref.GetURL() != nil
+}
+
+// fetchFileRelative fetches the schema for the relative path created by combining 'baseFileName' and 'url'
```
matthchr, line 191: It's sorta confusing that you interleave the schema and cache stuff in the same file. It would help I think if they were in two different files, or were written one after the other (so `MakeCache` + `Cache` + `Cache` methods and then the schema methods)

tools.mk
matthchr, line 58: What calls this?

hack/generator/pkg/codegen/swagger_type_extractor.go
```diff
@@ -0,0 +1,170 @@
...
+
+	for operationPath, op := range swagger.Paths.Paths {
+		put := op.Put
+		if put == nil {
+			continue
+		}
+
+		resourceName, err := extractor.resourceNameFromOperationPath(packageName, operationPath)
+		if err != nil {
+			klog.Error(err)
```
matthchr, line 53: Should this be fatal?

hack/generator/pkg/codegen/swagger_type_extractor.go
```diff
@@ -0,0 +1,170 @@
...
+		}
+
+		resourceType, err := extractor.resourceTypeFromOperation(ctx, scanner, swagger, filePath, put)
+		if err != nil {
+			if err == context.Canceled {
+				return err
+			}
+
+			klog.Errorf("unable to produce type for %v: %v", resourceName, err)
+			continue // TODO: make fatal
```
matthchr, line 64: Is this TODO for later or for before you merge?

hack/generator/pkg/codegen/swagger_type_extractor.go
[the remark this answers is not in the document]
matthchr: Include this in the doc comment for clarity?

hack/generator/pkg/codegen/swagger_type_extractor_test.go
```diff
@@ -0,0 +1,35 @@
...
+
+import (
+	"fmt"
+	"testing"
+
+	. "github.com/onsi/gomega"
+)
+
+func Example_inferNameFromURLPath() {
+	group, name, _ := inferNameFromURLPath("/Microsoft.GroupName/resourceName/{resourceId}")
```
matthchr, line 16: Is this a test?

Sorta weird casing `_infer` vs `_Infer`.

Also... no assert?

hack/generator/pkg/codegen/swagger_type_extractor_test.go
```diff
@@ -0,0 +1,35 @@
...
+	fmt.Printf("%s: %s", group, name)
+	// Output: Microsoft.GroupName: ResourceName
+}
+
+func TestInferNameFromURLPath_FailsWithMultipleParametersInARow(t *testing.T) {
```
matthchr, line 21: minor: I think a lot (most?) of the test do `Test_`

hack/generator/pkg/config/configuration.go
```diff
@@ -62,7 +66,7 @@ func (config *Configuration) WithExportFilters(filters ...*ExportFilter) *Config
 
 // Initialize checks for common errors and initializes structures inside the configuration
 // which need additional setup after json deserialization
-func (config *Configuration) Initialize() error {
+func (config *Configuration) Initialize(myLocation string) error {
```
matthchr, line 69: `path` rather than `myLocation`?

matthchr: And add doc comment that `path` can be either local or absolute?

hack/generator/pkg/codegen/pipeline_augment_status.go
```diff
@@ -0,0 +1,262 @@
...
+			klog.V(1).Infof("Input %v types, output %v types", len(types), len(newTypes))
+
+			return newTypes, nil
+		},
+	}
+}
+
+type resourceLookup map[astmodel.TypeName]astmodel.Type
+
+// makeStatusLookup makes a map of old name (non-status name) to status type
```
matthchr, line 97: Comment (and possibly function name) seems out of date with signature -- there are two return types here but comment only talks about one. What is the other one?

Also, since this seemingly is making two things (and also altering some types in the process) unclear if `makeStatusLookup` is the best name.

Wondering about something like: `renameStatusDefinitions` or something?

hack/generator/Makefile
```diff
@@ -14,10 +14,14 @@ V = 0
...
-all: header-check fmt build test gen-arm make-generated-code
+all: header-check fmt test gen-arm make-generated-code
+
+.PHONY: checkout-arm-submodule
+checkout-arm-submodule: $(info $(M) checking out azure-rest-api-specs...)
+	$(Q) git submodule update --init
 
 .PHONY: gen-arm
-gen-arm: ; $(info $(M) running k8sinfra-gen gen…) @ ## Running gen
+gen-arm: build checkout-arm-submodule ; $(info $(M) running k8sinfra-gen gen…) @ ## Running gen
```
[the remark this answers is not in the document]
Porges: I didn't think it was possible with Actions but have now found it 🙂

hack/generator/pkg/codegen/swagger_type_extractor_test.go
Porges: for an Example the specified magic Output comment is tested against what is what is written on stdout (and the Example goes in the docs, I think): see https://blog.golang.org/examples

Porges: and re casing, if it doesn't match then a linter fires 😛

hack/generator/pkg/jsonast/schema_abstraction_openapi.go
Porges: I walk all the .json files for each service and each version of the service, but there are some files that live outside the normal tree structure that don't get preloaded, such as files containing common types.

hack/generator/pkg/codegen/pipeline_augment_status.go
```diff
@@ -0,0 +1,262 @@
...
+
+		if err != nil {
+			return err
+		}
+
+		if ctx.Err() != nil {
+			return ctx.Err()
+		}
+
+		for _, skipDir := range skipDirectories {
```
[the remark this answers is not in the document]
Porges: Done.

hack/generator/pkg/jsonast/schema_abstraction_openapi.go
Porges: Really I want a module of the OpenApiSchema stuff (exporting just the schema/cache factories) and a module of the JSONSchema stuff (exporting just the schema factory) with only 1/2 files in each module, but Go feels like it doesn't encourage such small modules.

Porges: But agree they were not meant to be interleaved, reordered to group all cache bits together.

tools.mk
Porges: humans!

Porges: This target used to exist a while ago but disappeared.

hack/generator/pkg/jsonast/schema_abstraction_openapi.go
Porges: Done.

Porges: Actually this can be a value type again now that there are not mutexes involved. I will rerefactor.

hack/generator/pkg/codegen/swagger_type_extractor.go
Porges: later; need to address some items in #219

hack/generator/pkg/jsonast/schema_abstraction_openapi.go
Porges: Examples are things like:
* `azure-rest-api-specs/specification/automation/resource-manager/Microsoft.Automation/common/v1/definitions.json`
* `azure-rest-api-specs/specification/guestconfiguration/resource-manager/Microsoft.GuestConfiguration/common/v1/definitions.json`

hack/generator/pkg/jsonast/schema_abstraction_openapi.go
Porges: Done ×2.

hack/generator/pkg/config/configuration.go
Porges: Done and refactored to make it internal to `config`, an implementation detail.

hack/generator/pkg/codegen/swagger_type_extractor.go
Porges: No, at the moment this covers various operations that don't map to resources, mostly around the resource group stuff, and some other cases in #219

Porges: I'll add a specific item about this.

hack/generator/pkg/codegen/swagger_type_extractor.go
Porges: I'll add a specific item about this.

hack/generator/pkg/codegen/swagger_type_extractor.go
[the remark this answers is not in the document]
Porges: I added a test for this scenario as well to codify the behaviour.

hack/generator/pkg/codegen/pipeline_augment_status.go
Porges: Renamed to `generateStatusTypes` and added clarifying comment.

Porges: Reworked this greatly, actually. Should be more understandable now…

hack/generator/pkg/jsonast/schema_abstraction_openapi.go
Porges: I had thought about a common function but they live in different modules and the "sharing" would be one module reaching into another's internal helper functions. I don't think the duplication is a big deal here.

hack/generator/pkg/codegen/pipeline_augment_status.go
Porges: I was rather hopeful that the stdlib would already :)

Porges: Done.

hack/generator/pkg/jsonast/schema_abstraction_openapi.go
matthchr: Ah, I missed that they were in different modules. Carry on then, fine as is.

tools.mk
matthchr: Ah, probably accidentally removed in some refactor. Sounds good.

hack/generator/pkg/codegen/swagger_type_extractor_test.go
matthchr: TIL about Examples.

hack/generator/pkg/codegen/swagger_type_extractor.go
```diff
@@ -0,0 +1,168 @@
...
+			reading = true
+		}
+	}
+
+	if !reading {
+		return "", "", errors.Errorf("no group name (‘Microsoft…’ = %q) found", group)
+	}
+
+	if name == "" {
+		return "", "", errors.Errorf("couldn’t infer name")
```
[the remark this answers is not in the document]
matthchr: This was marked resolved but I had the same question and I don't see an operation path in the error?

hack/generator/pkg/jsonast/schema_abstraction.go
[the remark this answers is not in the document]
matthchr: Reactivated this because it feels weird to have an exported interface but no expected methods (I had the same double-take as @babbageclunk)

Is the _intent_ to hide the shape of this thing from callers? (i.e. they can make on using `MakeGoJSONSchema` or `MakeOpenAPISchema` but they aren't really supposed to be able introspect the type they get back?)

Not saying this needs to change, just want to double check that we're in a situation where we are hiding the internals because its "the right thing to do" and not just because this so happens to have been defined in the same module where it's used.

hack/generator/pkg/jsonast/jsonast.go
```diff
@@ -183,18 +168,27 @@ func (scanner *SchemaScanner) GenerateDefinitions(ctx context.Context, schema *g
...
+
 	defs := make(astmodel.Types)
 	for defName, def := range scanner.definitions {
 		if def == nil {
 			// safety check/assert:
 			panic(fmt.Sprintf("%v was nil", defName))
 		}
 
+		if defName != def.Name() {
+			panic("mismatched typenames!")
```
matthchr, line 185: minor: panic with the actual typenames that mismatch?

hack/generator/pkg/jsonast/jsonast.go
```diff
@@ -368,58 +363,69 @@ func getProperties(
...
 	// see: https://json-schema.org/understanding-json-schema/reference/object.html#properties
-	if schema.AdditionalProperties == nil {
-		// if not specified, any additional properties are allowed (TODO: tell all Azure teams this fact and get them to update their API definitions)
-		// for now we aren't following the spec 100% as it pollutes the generated code
-		// only generate this property if there are no other properties:
-		if len(properties) == 0 {
+	if schema.additionalPropertiesAllowed() {
+		additionalPropSchema := schema.additionalPropertiesSchema()
+		if additionalPropSchema == nil {
+			// if not specified, any additional properties are allowed (TODO: tell all Azure teams this fact and get them to update their API definitions)
```
matthchr, line 369: extra minor: We're linebreaking here slightly oddly -- first line is really long and the subsequent ones aren't.

IMO, break the first line either in the parenthesis statement, or maybe just break before it starts and put that whole statement onto it's own line?

hack/generator/pkg/jsonast/jsonast.go
```diff
@@ -368,58 +363,69 @@ func getProperties(
...
+			}
+		} else {
+			// otherwise, it is a type for all additional fields
 			// TODO: for JSON serialization this needs to be unpacked into "parent"
+			additionalPropsType, err := scanner.RunHandlerForSchema(ctx, additionalPropSchema)
+			if err != nil {
+				return nil, err
+			}
+
+			// This can happen if the property type was prune away by a type filter.
```
matthchr, line 389:
```suggestion
			// This can happen if the property type was pruned away by a type filter.
```

hack/generator/pkg/jsonast/schema_abstraction.go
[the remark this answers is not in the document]
Porges: yes, that is the intent, the internals are not useful for the caller since they’re just for `jsonast` to use (and exposing them means documenting them all as well!)

hack/generator/pkg/codegen/swagger_type_extractor.go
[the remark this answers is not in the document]
Porges: @matthchr the calling method (`resourceNameFromOperationPath`) adds the operation path to the error information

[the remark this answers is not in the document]
matthchr: Missed that! Resolving this again.

hack/generator/pkg/jsonast/schema_abstraction.go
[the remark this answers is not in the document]
matthchr: As long as that's the intent I think it's fine.

PR #206, opened 2020-07-30
Remove type aliases

matthchr:
- They can cause trouble with Kubebuilder and also
   makes our life more difficult when attaching methods to types.
 - Since the type isn't visible to the customer in the CRD anyway
   there's no customer impact doing this.


hack/generator/pkg/astmodel/optional_type.go
```diff
@@ -15,8 +15,15 @@ type OptionalType struct {
 }
 
 // NewOptionalType creates a new optional type that may or may not have the specified 'element' type
-func NewOptionalType(element Type) *OptionalType {
-	return &OptionalType{element}
+func NewOptionalType(element Type) Type {
+	switch element.(type) {
+	case *ArrayType:
```
Porges, line 20: probably have a comment here explaining it

hack/generator/pkg/astmodel/property_definition_test.go
```diff
@@ -280,7 +280,7 @@ func Test_PropertyDefinitionMakeOptional_PropertyTypeArrayAndMap(t *testing.T) {
 				original := NewPropertyDefinition(fieldName, fieldJsonName, c.propertyType)
 				field := original.MakeOptional()
 
-				g.Expect(field).To(BeIdenticalTo(original))
+				g.Expect(field).To(Equal(original))
```
theunrepentantgeek, line 283: This change isn't right - the test is checking that the exact same field is returned, avoiding memory churn; merely returning a field that's equal to the original isn't a strong enough assertion.

hack/generator/pkg/codegen/pipeline_remove_type_aliases.go
```diff
@@ -0,0 +1,67 @@
...
+	switch concreteType := def.Type().(type) {
+	case *astmodel.ObjectType:
+		return def.Name()
+	case *astmodel.EnumType:
+		return def.Name()
+	case *astmodel.ResourceType:
+		return def.Name()
+	case astmodel.TypeName:
+		// We need to resolve further because this type is an alias
+		klog.V(0).Infof("Found type alias %s, replacing it with %s", name, concreteType)
```
theunrepentantgeek, line 54: Probably need this at `V(3)`

hack/generator/pkg/astmodel/property_definition_test.go
matthchr: I've fixed this, although:
1. We allocate types quite freely basically everywhere (when modifying immutable stuff, using the visitor, etc) and I don't think this particular location is really even a blip on the radar in terms of how many allocations it drives (because technically speaking it only returns a new copy if the type is an optional Map or optional Array).
2. Including the fix makes the code a bit more awkward IMO because `MakeRequired` and `MakeOptional` end up diverging in how they check for optionality -- `MakeRequired` just needs to know if the type is exactly `Optional` (to unroll it) whereas `MakeOptional` needs to check if it's `Optional`, `Map`, or `Array` (to return early).

I originally hadn't fixed this because it felt like the extra level of awkwardness in the code wasn't worth the negligible perf improvement (especially given how freely we allocate memory elsewhere). I've included the fix in the updated PR though so folks can take a look and let me know if they have a better way to do the checks that's cleaner. If not I left a comment saying why in `MakeOptional`.

hack/generator/pkg/codegen/pipeline_remove_type_aliases.go
```diff
@@ -0,0 +1,67 @@
...
+		return def.Name()
+	case *astmodel.ResourceType:
+		return def.Name()
+	case astmodel.TypeName:
+		// We need to resolve further because this type is an alias
+		klog.V(3).Infof("Found type alias %s, replacing it with %s", name, concreteType)
+		return resolveTypeName(concreteType, types)
+	case *astmodel.PrimitiveType:
+		return concreteType
+	case *astmodel.OptionalType:
```
Porges, line 58: I think for Optional, Array, and Map you want to Visit the concrete types here, or else nested TypeNames inside them will not be resolved. (This should be easy enough to write a golden test for.)

hack/generator/pkg/astmodel/property_definition_test.go
theunrepentantgeek: My motivation is to keep things tidy so that later changes are straightforward. 

For example, in a situation where we're allocating too much memory and we need to increase out sharing of immutable instances, fixes will be easier if our leaf types (like this one) are already optimal with instance reuse.

hack/generator/pkg/codegen/pipeline_remove_type_aliases.go
matthchr: Doesn't the default TypeVisitor also have this problem?

From what I can tell of the usage of TypeVisitor, it's usually called once for each def, and the default impl stops at a `TypeName` (it doesn't recurse in), which I think makes sense because anyway `Visit` will be called for the other types in the `Types` collection since the visitor is used in a loop?

I think this means we won't drop any types?

Porges: but here you are inlining named types, so essentially cloning them from the collection. Depending on the processing order of the `Types` collection they may or may not have already been visited, so in some cases you will get unprocessed types inlined.

hack/generator/pkg/codegen/pipeline_remove_type_aliases.go
```diff
@@ -0,0 +1,67 @@
...
+func resolveTypeName(name astmodel.TypeName, types astmodel.Types) astmodel.Type {
+	def, ok := types[name]
+	if !ok {
+		panic(fmt.Sprintf("Couldn't find type for type name %s", name))
+	}
+
+	// If this typeName definition has a type of object, enum, or resource
+	// it's okay. Everything else we want to pull up one level to remove the alias
+	switch concreteType := def.Type().(type) {
+	case *astmodel.ObjectType:
```
Porges, line 46: BTW go does let you combine cases in a type-switch

hack/generator/pkg/codegen/pipeline_remove_type_aliases.go
matthchr: but getting an unprocessed type inlined doesn't matter does it? That type will get processed afterwards anyway since bypassing it here won't remove it from the `Types` collection? As a matter of fact this pipeline stage shouldn't be removing any types (`len(types) == length(result)`). The types that aren't referenced anymore (because they're just a type alias and so were bypassed by this pipeline phase) will be removed by the later phase that removes unreferenced types?

hack/generator/pkg/codegen/pipeline_remove_type_aliases.go
matthchr: I don't think it does =)

hack/generator/pkg/codegen/pipeline_remove_type_aliases.go
theunrepentantgeek: I think this calls for a unit test to prove (or disprove) the need to recurse within Optional, Array, and Map.

matthchr: George wrote one and showed me what the issue was

hack/generator/pkg/codegen/pipeline_remove_type_aliases.go
Porges: boo!

PR #207, opened 2020-07-31
jsonast cleanup

matthchr:
Just some cleanup taken from my work in jsonast for upcoming type ownership PR

### DIFF
--- a/hack/generator/pkg/jsonast/jsonast.go
+++ b/hack/generator/pkg/jsonast/jsonast.go
@@ -36,9 +36,6 @@ type (
 		Filters []string
 	}
 
-	// A BuilderOption is used to provide custom configuration for our scanner
-	BuilderOption func(scanner *SchemaScanner) error
-
 	// A SchemaScanner is used to scan a JSON Schema extracting and collecting type definitions
 	SchemaScanner struct {
 		definitions   map[astmodel.TypeName]*astmodel.TypeDefinition
@@ -100,7 +97,7 @@ func (use *UnknownSchemaError) Error() string {
 func NewSchemaScanner(idFactory astmodel.IdentifierFactory, configuration *config.Configuration) *SchemaScanner {
 	return &SchemaScanner{
 		definitions:   make(map[astmodel.TypeName]*astmodel.TypeDefinition),
-		TypeHandlers:  DefaultTypeHandlers(),
+		TypeHandlers:  defaultTypeHandlers(),
 		configuration: configuration,
 		idFactory:     idFactory,
 	}
@@ -156,19 +153,9 @@ func (scanner *SchemaScanner) RunHandlerForSchema(ctx context.Context, schema *g
 // 							- ARM specific resources. I'm not 100% sure why...
 //
 // 		allOf acts like composition which composites each schema from the child oneOf with the base reference from allOf.
-func (scanner *SchemaScanner) GenerateDefinitions(
-	ctx context.Context,
-	schema *gojsonschema.SubSchema,
-	opts ...BuilderOption) (astmodel.Types, error) {
-
+func (scanner *SchemaScanner) GenerateDefinitions(ctx context.Context, schema *gojsonschema.SubSchema) (map[astmodel.TypeName]astmodel.TypeDefinition, error) {
 	ctx, span := tab.StartSpan(ctx, "GenerateDefinitions")
 	defer span.End()
-
-	for _, opt := range opts {
-		if err := opt(scanner); err != nil {
-			return nil, err
-		}
-	}
 
 	// get initial topic from ID and Title:
 	url := schema.ID.GetUrl()
@@ -194,7 +181,7 @@ func (scanner *SchemaScanner) GenerateDefinitions(
 
 	rootTypeName := astmodel.MakeTypeName(rootPackage, rootName)
 
-	_, err = generateDefinitionsFor(ctx, scanner, rootTypeName, false, url, schema)
+	_, err = generateDefinitionsFor(ctx, scanner, rootTypeName, schema)
 	if err != nil {
 		return nil, err
 	}
@@ -213,8 +200,8 @@ func (scanner *SchemaScanner) GenerateDefinitions(
 	return defs, nil
 }
 
-// DefaultTypeHandlers will create a default map of JSONType to AST transformers
-func DefaultTypeHandlers() map[SchemaType]TypeHandler {
+// defaultTypeHandlers will create a default map of JSONType to AST transformers
+func defaultTypeHandlers() map[SchemaType]TypeHandler {
 	return map[SchemaType]TypeHandler{
 		Array:  arrayHandler,
 		OneOf:  oneOfHandler,
@@ -332,8 +319,13 @@ func generatePropertyDefinitions(ctx context.Context, scanner *SchemaScanner, pr
 	return property, nil
 }
 
-func getProperties(ctx context.Context, scanner *SchemaScanner, schema *gojsonschema.SubSchema) ([]*astmodel.PropertyDefinition, error) {
+func getProperties(
+	ctx context.Context,
+	scanner *SchemaScanner,
+	schema *gojsonschema.SubSchema) ([]*astmodel.PropertyDefinition, error) {
+
 	ctx, span := tab.StartSpan(ctx, "getProperties")
+
 	defer span.End()
 
 	var properties []*astmodel.PropertyDefinition
@@ -440,8 +432,6 @@ func refHandler(ctx context.Context, scanner *SchemaScanner, schema *gojsonschem
 		return nil, err
 	}
 
-	isResource := isResource(url)
-
 	// produce a usable name:
 	typeName := astmodel.MakeTypeName(
 		astmodel.MakeLocalPackageReference(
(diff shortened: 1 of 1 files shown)


hack/generator/pkg/jsonast/jsonast.go
```diff
@@ -463,22 +453,23 @@ func refHandler(ctx context.Context, scanner *SchemaScanner, schema *gojsonschem
...
 	schema *gojsonschema.SubSchema) (astmodel.Type, error) {
 
 	schemaType, err := getSubSchemaType(schema)
 	if err != nil {
 		return nil, err
 	}
 
+	url := schema.ID.GetUrl()
+	isResource := isResource(url)
+
```
Porges, line 472: are you sure this change is okay? It will be getting the URL for the ref’d schema and not the schema itself

hack/generator/pkg/jsonast/jsonast.go
```diff
@@ -552,7 +543,7 @@ func allOfHandler(ctx context.Context, scanner *SchemaScanner, schema *gojsonsch
 		case *astmodel.ResourceType:
 			// it is a little strange to merge one resource into another with allOf,
 			// but it is done and therefore we have to support it.
-			// (an example is Microsoft.VisualStudio’s Project type)
+			// (an example is Microsoft.VisualStudio's Project type)
```
Porges, line 546: boo 🙂

theunrepentantgeek, line 546: I can get rid of the disagreement ...
```suggestion
			// (an example is the Microsoft.VisualStudio Project type)
```

hack/generator/pkg/jsonast/jsonast.go
```diff
@@ -97,10 +94,10 @@ func (use *UnknownSchemaError) Error() string {
 }
 
 // NewSchemaScanner constructs a new scanner, ready for use
-func NewSchemaScanner(idFactory astmodel.IdentifierFactory, configuration *config.Configuration) *SchemaScanner {
+func NewSchemaScanner(idFactory astmodel.IdentifierFactory, configuration *config.Configuration, typeHandlers map[SchemaType]TypeHandler) *SchemaScanner {
```
theunrepentantgeek, line 97: Why inject the type handlers from outside the type? It feels as though an internal (implementation) issue is being pushed outside, violating encapsulation.

hack/generator/pkg/jsonast/jsonast.go
matthchr: It should be.

`generateDefinitionsFor` was called from two places:
`refHandler`, which does:
```
func refHandler... {
    url := schema.Ref.GetUrl()
    isResource := isResource(url)
    ...
    return generateDefinitionsFor(ctx, scanner, typeName, isResource, url, schema.RefSchema)
}
```

and

```
func (scanner *SchemaScanner) GenerateDefinitions ... {
    url := schema.ID.GetUrl()
    ...
    _, err = generateDefinitionsFor(ctx, scanner, rootTypeName, false, url, schema)
}
```

In both cases, the passed schema should be the same entity which the `url` was gotten from in the original caller. The second one is obviously correct. In the first one, `schema.Ref.GetUrl()` and `schema.RefSchema.ID.GetUrl()` return the same thing.

I think this change additionally makes sense -- it doesn't make sense to call `generateDefinitionsFor` with a url that _doesn't_ match the schema provided (or with `isResource` true when the schema in question is not actually pointing to a resource). Really the existing 3 parameters (`isResource`, `url`, and `schema`) are all tightly coupled and the first 2 have to have been derived from the 3rd for the call to be correct - this just makes that requirement official in the signature.

I added some sanity logging and ran the full generator just to triple check and here's a snippet of what I got:
```
 RefHandler url: https://schema.management.azure.com/schemas/2015-01-01/Microsoft.Resources.json#/resourceDefinitions/links
 generateDefinitionsFor url: https://schema.management.azure.com/schemas/2015-01-01/Microsoft.Resources.json#/resourceDefinitions/links
 RefHandler url: https://schema.management.azure.com/schemas/2015-01-01/Microsoft.Authorization.json#/resourceDefinitions/locks
 generateDefinitionsFor url: https://schema.management.azure.com/schemas/2015-01-01/Microsoft.Authorization.json#/resourceDefinitions/locks
 RefHandler url: https://schema.management.azure.com/schemas/2017-09-01/Microsoft.Authorization.json#/resourceDefinitions/roleAssignments
 generateDefinitionsFor url: https://schema.management.azure.com/schemas/2017-09-01/Microsoft.Authorization.json#/resourceDefinitions/roleAssignments
 RefHandler url: https://schema.management.azure.com/schemas/2017-09-01/Microsoft.Authorization.json#/definitions/RoleAssignmentProperties
 generateDefinitionsFor url: https://schema.management.azure.com/schemas/2017-09-01/Microsoft.Authorization.json#/definitions/RoleAssignmentProperties
 RefHandler url: https://schema.management.azure.com/schemas/2017-09-01/Microsoft.Authorization.json#/resourceDefinitions/roleDefinitions
 generateDefinitionsFor url: https://schema.management.azure.com/schemas/2017-09-01/Microsoft.Authorization.json#/resourceDefinitions/roleDefinitions
 RefHandler url: https://schema.management.azure.com/schemas/2017-09-01/Microsoft.Authorization.json#/definitions/RoleDefinitionProperties
 generateDefinitionsFor url: https://schema.management.azure.com/schemas/2017-09-01/Microsoft.Authorization.json#/definitions/RoleDefinitionProperties
 RefHandler url: https://schema.management.azure.com/schemas/2017-09-01/Microsoft.Authorization.json#/definitions/Permission
 generateDefinitionsFor url: https://schema.management.azure.com/schemas/2017-09-01/Microsoft.Authorization.json#/definitions/Permission
 RefHandler url: https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#/definitions/output
 generateDefinitionsFor url: https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#/definitions/output
 RefHandler url: https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#/definitions/parameterTypes
 generateDefinitionsFor url: https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#/definitions/parameterTypes
 RefHandler url: https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#/definitions/parameterValueTypes
 generateDefinitionsFor url: https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#/definitions/parameterValueTypes
 RefHandler url: https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#/definitions/outputCopy
 generateDefinitionsFor url: https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#/definitions/outputCopy
 RefHandler url: https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#/definitions/parameter
 generateDefinitionsFor url: https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#/definitions/parameter
W0803 10:11:24.928653   28608 jsonast.go:714] Interface assumption unproven for https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#/definitions/parameter
 RefHandler url: https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#/definitions/parameterTypes
 generateDefinitionsFor url: https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#/definitions/parameterTypes
 RefHandler url: https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#/definitions/parameterValueTypes
 generateDefinitionsFor url: https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#/definitions/parameterValueTypes
```

hack/generator/pkg/jsonast/jsonast.go
matthchr: I think this actually arose from my copying this file into vscode and back (possibly across my VM boundary?) which I had to do because of some shenanigans unrelated to this change.

I'll put this back =)

matthchr: Actually I just reformatted it to what Bevan suggested below. I am not philisophically opposed to the original character though it just got mangled in some horrible copy operation I did and I replaced it manually thinking that's what it had been when it reality it was ’ not '

hack/generator/pkg/jsonast/jsonast.go
matthchr: 1. Originally in my PR you haven't yet seen, I needed to supply different handlers for the "ownership discovery" bits vs the normal bits. Now that this is happening in a pipeline stage this is no longer needed, so technically speaking I can revert this change (it's not required for anything that is coming).
2. I don't think that this necessarily violates encapsulation though - `TypeHandlers` as a property on `SchemaScanner` is already public, as is the function `DefaultTypeHandlers` - even the name `DefaultTypeHandlers` suggests that a consumer of the schema scanner understands that there are `TypeHandlers` and they can be overridden - this is just providing an easier way to do it all on one line as opposed to constructing a `SchemaScanner` and then going `scanner.TypeHandlers[refHandler] = ...`.

With that said -- since this change isn't actually required for what is coming, I've made `DefaultTypeHandlers` package scope rather than exported and removed this parameter since there's no caller who needs to make changes to the type handlers collection right now.

hack/generator/pkg/jsonast/jsonast.go
Porges: Looks good then. I think I have to rollback this slightly for #205 though since the openapi types don't work in the same way, but that's something I can deal with 🙂

PR #210, opened 2020-07-31
Add astbuilder

matthchr:
- Simple helper module for building asts.

### DIFF
--- a/hack/generator/pkg/astbuilder/builder.go
+++ b/hack/generator/pkg/astbuilder/builder.go
@@ -1,0 +1,197 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ */
+
+package astbuilder
+
+import (
+	"fmt"
+	"go/ast"
+	"go/token"
+)
+
+// CheckErrorAndReturn checks if the err is non-nil, and if it is returns. For example:
+// 	if err != nil {
+// 		return <otherReturns...>, err
+//	}
+func CheckErrorAndReturn(otherReturns ...ast.Expr) ast.Stmt {
+
+	returnValues := append([]ast.Expr{}, otherReturns...)
+	returnValues = append(returnValues, ast.NewIdent("err"))
+
+	return &ast.IfStmt{
+		Cond: &ast.BinaryExpr{
+			X:  ast.NewIdent("err"),
+			Op: token.NEQ,
+			Y:  ast.NewIdent("nil"),
+		},
+		Body: &ast.BlockStmt{
+			List: []ast.Stmt{
+				&ast.ReturnStmt{
+					Results: returnValues,
+				},
+			},
+		},
+	}
+}
+
+// NewStruct creates a new assignment statement where a struct is constructed and stored in a variable of the given name. For example:
+// 	<varName> := <structName>{}
+func NewStruct(varName *ast.Ident, structName *ast.Ident) ast.Stmt {
+	return SimpleAssignment(
+		varName,
+		token.DEFINE,
+		&ast.CompositeLit{
+			Type: structName,
+		})
+}
+
+// NilGuardAndReturn checks if a variable is nil and if it is returns, like:
+// 	if <toCheck> == nil {
+// 		return <returns...>
+//	}
+func NilGuardAndReturn(toCheck *ast.Ident, returns ...ast.Expr) ast.Stmt {
+	return &ast.IfStmt{
+		Cond: &ast.BinaryExpr{
+			X:  toCheck,
+			Op: token.EQL,
+			Y:  ast.NewIdent("nil"),
+		},
+		Body: &ast.BlockStmt{
+			List: []ast.Stmt{
+				&ast.ReturnStmt{
+					Results: returns,
+				},
+			},
+		},
+	}
+}
+
+type FuncDetails struct {
+	ReceiverIdent *ast.Ident
+	ReceiverType  ast.Expr
+	Name          *ast.Ident
+	Comment       string
+	Params        []*ast.Field
+	Returns       []*ast.Field
+	Body          []ast.Stmt
+}
+
+// DefineFunc defines a function (header, body, etc), like:
+// 	<comment>
+//	func (<receiverIdent> <receiverType>) <name>(<params...>) (<returns...>) {
+// 		<body...>
+//	}
+func DefineFunc(funcDetails FuncDetails) *ast.FuncDecl {
+
+	var comment []*ast.Comment
+	if funcDetails.Comment != "" {
+		comment = []*ast.Comment{
+			{
+				Text: fmt.Sprintf("// %s %s", funcDetails.Name, funcDetails.Comment),
+			},
+		}
+	}
+
+	return &ast.FuncDecl{
+		Recv: &ast.FieldList{
+			List: []*ast.Field{
+				{
+					Names: []*ast.Ident{
+						funcDetails.ReceiverIdent,
+					},
+					Type: funcDetails.ReceiverType,
+				},
+			},
+		},
+		Name: funcDetails.Name,
+		Doc: &ast.CommentGroup{
+			List: comment,
+		},
+		Type: &ast.FuncType{
+			Params: &ast.FieldList{
+				List: funcDetails.Params,
+			},
+			Results: &ast.FieldList{
+				List: funcDetails.Returns,
+			},
+		},
+		Body: &ast.BlockStmt{
+			List: funcDetails.Body,
+		},
+	}
+}
+
+// SimpleAssignment performs a simple assignment like:
+// 	<lhs> <tok> <rhs>
+func SimpleAssignment(lhs ast.Expr, tok token.Token, rhs ast.Expr) *ast.AssignStmt {
+	return &ast.AssignStmt{
+		Lhs: []ast.Expr{
+			lhs,
+		},
+		Tok: tok,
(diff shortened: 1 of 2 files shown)


matthchr:
This is quite useful in reducing verbosity a bit... building asts is still super verbose but it reads a bit better with some of these helpers.

Actual use of this module will come in a follow up PR.

Porges:
Seems fine, hard to judge without usage.

hack/generator/pkg/astbuilder/builder.go
```diff
@@ -0,0 +1,190 @@
...
+
+	var returnValues []ast.Expr
+	for _, ret := range otherReturns {
+		returnValues = append(returnValues, ret)
+	}
+	returnValues = append(returnValues, ast.NewIdent("err"))
+
+	return &ast.IfStmt{
+		Cond: &ast.BinaryExpr{
+			X:  ast.NewIdent("err"),
```
theunrepentantgeek, line 25: Is the assumption that it's always called `err` safe?

hack/generator/pkg/astbuilder/builder.go
```diff
@@ -0,0 +1,190 @@
...
+	receiverIdent *ast.Ident,
+	receiverType ast.Expr,
+	name *ast.Ident,
+	comment string,
+	params []*ast.Field,
+	returns []*ast.Field,
+	body []ast.Stmt) *ast.FuncDecl {
+
+	// TODO: I'm not actually sure this is better than just declaring it all inline...?
+	// TODO: Maybe if this took a struct param instead of so many individual items?
```
theunrepentantgeek, line 89: A parameter struct could be constructed with a fluent-lite chain of method calls, might pan out as easier to read.

hack/generator/pkg/astbuilder/builder.go
```diff
@@ -0,0 +1,190 @@
...
+			},
+		},
+		Body: &ast.BlockStmt{
+			List: body,
+		},
+	}
+}
+
+// SimpleAssignment performs a simple assignment
+func SimpleAssignment(lhs ast.Expr, tok token.Token, rhs ast.Expr) *ast.AssignStmt {
```
theunrepentantgeek, line 125: Why does this need `tok`? Won't it always be assignment (by definition)?

hack/generator/pkg/astbuilder/builder.go
```diff
@@ -0,0 +1,190 @@
...
+	"go/token"
+)
+
+// CheckErrorAndReturn checks if the err is non-nil, and if it is returns.
+func CheckErrorAndReturn(otherReturns ...*ast.Ident) ast.Stmt {
+
+	var returnValues []ast.Expr
+	for _, ret := range otherReturns {
+		returnValues = append(returnValues, ret)
+	}
```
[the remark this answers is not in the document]
babbageclunk: `otherReturns[:]` won't create new storage, just a new slice object (https://play.golang.org/p/zmyDBLe9vGu) - you'd need to use `copy` to get that:

https://play.golang.org/p/NWN7LRtiLIQ

Another option is `returnValues := append([]ast.Expr, otherReturns...)`.

hack/generator/pkg/astbuilder/builder.go
```diff
@@ -0,0 +1,190 @@
...
+
+package astbuilder
+
+import (
+	"fmt"
+	"go/ast"
+	"go/token"
+)
+
+// CheckErrorAndReturn checks if the err is non-nil, and if it is returns.
```
babbageclunk, line 14: It would be useful to have the output syntax that these correspond to in the doc comments.

hack/generator/pkg/astbuilder/builder.go
babbageclunk, line 89: I think a parameter struct would definitely be good here, doubly so if some of the args are optional.

hack/generator/pkg/astbuilder/builder.go
```diff
@@ -0,0 +1,190 @@
...
+		},
+		Tok: tok,
+		Rhs: []ast.Expr{
+			rhs,
+		},
+	}
+}
+
+// SimpleAssignment performs a simple assignment
+func SimpleVariableDecalaration(ident *ast.Ident, typ ast.Expr) *ast.DeclStmt {
```
babbageclunk, line 138: Typo SimpleVariableDeclaration

hack/generator/pkg/astbuilder/builder.go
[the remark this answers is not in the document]
theunrepentantgeek: Huh. I tried it out in the Playground before making the comment, so _sometimes_ the storage isn't shared. But I can't now reproduce it. 😁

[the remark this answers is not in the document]
babbageclunk: Well, it creates a new slice object (pointer, length, capacity), but the pointer will be shared. It might not look like they're shared if you do an append when length = capacity, because in that case it needs to allocate new storage and copy the existing items across and it will only update the pointer the new slice b.
https://play.golang.org/p/jorKrpM7vb2

But if length < capacity then the append won't reallocate, so a and b will both be operating on the same storage.

https://play.golang.org/p/Xcb5YvQMxE8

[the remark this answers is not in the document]
matthchr: I think the original issue doing what @babbageclunk had suggested was that I had the parameter type here as `[]*ast.Ident` but the collection to supply to `ast.ReturnStmt` needed to be an `[]ast.Expr`, and you can't "reinterpret" a collection type like that in Go (the generics they have are not covariant).

In retrospect I should've just made the parameter `[]ast.Expr` instead as that's more flexible anyway (and still works for `*ast.Ident` since `*ast.Ident` impls `ast.Expr`.

hack/generator/pkg/astbuilder/builder.go
matthchr: It's _pretty_ safe and these are intended to be helpers for common patterns not replace the `ast` library entirely so I think as long as they cater for the 80+% case they've achieved their objective.

Adding support for other configuration increases the complexity of the signature and my objective was to keep the signatures as simple as I could (in contrast to ast which has quite complex signatures for most of the struct constructions because they need to support all the cases).

I think if/when we run into a situation where it can't be `err` (not sure what that would be off the top of my head) we can at that point either:
1. Just use raw `ast` for that uncommon case.
2. Update this method to allow an `errIdent` parameter and find/update callers.
3. Add a new method/override with more configuration options.

hack/generator/pkg/astbuilder/builder.go
matthchr: Added

hack/generator/pkg/astbuilder/builder.go
matthchr: Not necessarily. This supports:
`<lhs> = <rhs>` and `<lhs> := <rhs>`

Possibly the language (assignment) is throwing a wrench here - but that's what the go ast calls the statement (and it supports both "definition" and "assignment"). I was using that terminology here.

If you're getting at the fact that this is _too_ flexible (because a tok could be anything), I sorta agree although again the ast lib doesn't make that distinction and allows asts that don't actually compile like assignment statement `foo } bar`. We could throw a `panic` in if `tok` is anything other than `DEFINE` or `ASSIGN`

hack/generator/pkg/astbuilder/builder.go
matthchr: With the way this is written right now, all of the args are required. I've updated it to where the comment is optional though and moved to a struct paramter.

hack/generator/pkg/astbuilder/builder.go
theunrepentantgeek: Works for me.

hack/generator/pkg/astbuilder/builder.go
theunrepentantgeek: What about splitting this into two methods, one for each legit method:
``` go
func SimpleAssignment(lhs ast.Expr, rhs ast.Expr) *ast.AssignStmt {
```
and
```
func DeclarationAssignment(lhs ast.Expr, rhs ast.Expr) *ast.AssignStmt {
```
Thoughts?

hack/generator/pkg/astbuilder/builder.go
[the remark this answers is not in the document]
babbageclunk: Oh right, I hadn't noticed this was a type conversion as well.

hack/generator/pkg/astbuilder/builder.go
matthchr: I thought about this and initially liked it, but then realized there are actually quite a few tokens which are valid here:
`+=`, `-=`, `&=`, `=`, `:=`, etc -- there are a total of 13 which will work so I think actually we should keep it as is.

I'm open to changing the name of the function though to be more generic (although I am not really sure what we can change it to) - the only thing I came up with was `SimpleAssignmentStatement` but I didn't really like that because none of the other helpers really refer to the "ast bits" directly (i.e. it's `NilGuardAndReturn` not `NilGuardAndReturnStatement`)

hack/generator/pkg/astbuilder/builder.go
matthchr: Added struct -- no fluent like helpers for now but I think just the struct is probably good enough initially. We can always add more helpers later if we want.

PR #214, opened 2020-08-05
Put generated code into its own module, build during CI

matthchr:
- It is built during CI now (and by make all for local testing).


hack/generator/Makefile
```diff
@@ -87,5 +62,9 @@ gox: install-tools
 	$(Q) tar -czvf ./bin/$(SHORT_VERSION)/$(APP)_linux_amd64.tar.gz -C ./bin/$(SHORT_VERSION)/ $(APP)_linux_amd64
 	$(Q) tar -czvf ./bin/$(SHORT_VERSION)/$(APP)_windows_amd64.tar.gz -C ./bin/$(SHORT_VERSION)/ $(APP)_windows_amd64.exe
 
+.PHONY: make-generated-code
+make-generated-code:
+	$(Q) cd ../generated && make all
```
matthchr, line 67: I'm far from a make expert - this works pretty well but wouldn't mind a second set of eyes if there's some other way we should be doing this.

Longer term this probably won't be how we do it...
(@devigned maybe?)

Porges: https://www.gnu.org/software/make/manual/html_node/Recursion.html

PR #218, opened 2020-08-13
Add reference ids to PipelineStage

theunrepentantgeek:
To avoid the need to identify pipeline stages by doing string comparisons on their descriptions (which is brittle), I've restructured `PipeLineStage` to have an explicit `id`, along with a helper method `HasId()`.


theunrepentantgeek:
Gah! I ran `make ci` locally and it passed just fine. #investigating

PR #224, opened 2020-08-18
Move Types into a separate file and test

theunrepentantgeek:
Move `Types` into a separate file, then adds additional functionality:

* `AddAll(otherTypes Types)` - adds all the types into this set, with the same check as for the existing `Add()`
* `Where(predicate func(definition TypeDefinition) bool) Types` - returns a new set of filtered types
* `Except(otherTypes Types) Types` - returns a new set containing only types not in `otherTypes`
* `Contains(def TypeDefinition) bool` - what it says on the tin. 😁 

Full unit test coverage.

### DIFF
--- a/hack/generator/pkg/astmodel/types.go
+++ b/hack/generator/pkg/astmodel/types.go
@@ -39,15 +39,15 @@ func (types Types) Where(predicate func(definition TypeDefinition) bool) Types {
 	return result
 }
 
-// Except returns a new set of types including only those not present in otherTypes
+// Except returns a new set of types including only those not defined in otherTypes
 func (types Types) Except(otherTypes Types) Types {
 	return types.Where(func(def TypeDefinition) bool {
-		return !otherTypes.Contains(def)
+		return !otherTypes.Contains(def.Name())
 	})
 }
 
-// Contains returns true if the set contains the passed definition
-func (types Types) Contains(def TypeDefinition) bool {
-	_, ok := types[def.Name()]
+// Contains returns true if the set contains a definition for the specified name
+func (types Types) Contains(name TypeName) bool {
+	_, ok := types[name]
 	return ok
 }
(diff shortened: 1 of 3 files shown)


hack/generator/pkg/astmodel/types.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+// Except returns a new set of types including only those not present in otherTypes
+func (types Types) Except(otherTypes Types) Types {
+	return types.Where(func(def TypeDefinition) bool {
+		return !otherTypes.Contains(def)
+	})
+}
+
+// Contains returns true if the set contains the passed definition
```
Porges, line 49: comment is untrue, it checks if the set contains something with the same _name_ 🙂

hack/generator/pkg/astmodel/types.go
```diff
@@ -0,0 +1,53 @@
...
+// AddAll adds multiple types to the set, with the same safety check as Add() to panic if a duplicate is included
+func (types Types) AddAll(otherTypes Types) {
+	for _, t := range otherTypes {
+		types.Add(t)
+	}
+}
+
+// Where returns a new set of types including only those that satisfy the predicate
+func (types Types) Where(predicate func(definition TypeDefinition) bool) Types {
+	result := make(Types, len(types))
```
Porges, line 32: allocating the full size might be a pessimization if we are filtering out lots of types

hack/generator/pkg/astmodel/types.go
```diff
@@ -0,0 +1,53 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ */
+
+package astmodel
+
+import "fmt"
+
+// Types is the set of all types being generated
```
matthchr, line 10: minor: This comment supposes a particular usage.

```suggestion
// Types is a map of TypeName to TypeDefinition, representing a set of types.
```

hack/generator/pkg/astmodel/types.go
theunrepentantgeek: Changed to accept a name, and updated the comment.

hack/generator/pkg/astmodel/types.go
theunrepentantgeek: Agreed; removed the length parameter.

PR #226, opened 2020-08-18
Add Flags to Resource and Object types to aid internal flow

theunrepentantgeek:
Adds the ability to add a post-it note style flag to a type in one place so that you can recognize it elsewhere.

### DIFF
--- a/hack/generator/pkg/astmodel/flags.go
+++ b/hack/generator/pkg/astmodel/flags.go
@@ -1,0 +1,9 @@
+package astmodel
+
+type Flag string
+
+const (
+	OneOfFlag          Flag = "OneOf"
+	AnyOfFlag          Flag = "AnyOf"
+	StorageVariantFlag Flag = "Storage"
+)
--- a/hack/generator/pkg/astmodel/resource.go
+++ b/hack/generator/pkg/astmodel/resource.go
@@ -132,15 +132,30 @@ func (resource *ResourceType) AsType(_ *CodeGenerationContext) ast.Expr {
 }
 
 // Equals returns true if the other type is also a ResourceType and has Equal fields
-func (definition *ResourceType) Equals(other Type) bool {
-	if definition == other {
+func (resource *ResourceType) Equals(other Type) bool {
+	if resource == other {
 		return true
 	}
 
+	if resource == nil || other == nil {
+		return false
+	}
+
 	if otherResource, ok := other.(*ResourceType); ok {
-		return TypeEquals(definition.spec, otherResource.spec) &&
-			TypeEquals(definition.status, otherResource.status) &&
-			definition.isStorageVersion == otherResource.isStorageVersion
+		if !TypeEquals(resource.spec, otherResource.spec) ||
+			!TypeEquals(resource.status, otherResource.status) ||
+			!resource.isStorageVersion == otherResource.isStorageVersion ||
+			len(resource.flags) != len(otherResource.flags) {
+			return false
+		}
+
+		for f := range resource.flags {
+			if _, ok := otherResource.flags[f]; !ok {
+				return false
+			}
+		}
+
+		return true
 	}
 
 	return false
--- a/hack/generator/pkg/astmodel/resource.go
+++ b/hack/generator/pkg/astmodel/resource.go
@@ -254,3 +254,14 @@ func (resource *ResourceType) AsDeclarations(codeGenerationContext *CodeGenerati
 
 	return declarations
 }
+
+func (resource *ResourceType) copy() *ResourceType {
+	result := *resource
+
+	result.flags = make(map[string]struct{})
+	for k, v := range resource.flags {
+		result.flags[k] = v
+	}
+
+	return &result
+}
--- a/hack/generator/pkg/astmodel/resource.go
+++ b/hack/generator/pkg/astmodel/resource.go
@@ -21,11 +21,16 @@ type ResourceType struct {
 	status           Type
 	isStorageVersion bool
 	owner            *TypeName
+
+	// flags are used to identify resources during processing;
+	// they have no direct representation in the generated code.
+	flags map[Flag]struct{}
 }
 
 // NewResourceType defines a new resource type
 func NewResourceType(specType Type, statusType Type) *ResourceType {
-	return &ResourceType{specType, statusType, false, nil}
+	return &ResourceType{
+		specType, statusType, false, nil, make(map[Flag]struct{})}
 }
 
 // NewAzureResourceType defines a new resource type for Azure. It ensures that
@@ -137,11 +142,16 @@ func (resource *ResourceType) Equals(other Type) bool {
 		return true
 	}
 
-	if resource == nil || other == nil {
+	if resource == nil {
 		return false
 	}
 
 	if otherResource, ok := other.(*ResourceType); ok {
+
+		if otherResource == nil {
+			return false
+		}
+
 		if !TypeEquals(resource.spec, otherResource.spec) ||
 			!TypeEquals(resource.status, otherResource.status) ||
 			!resource.isStorageVersion == otherResource.isStorageVersion ||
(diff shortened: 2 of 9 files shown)


hack/generator/pkg/astmodel/flags.go
```diff
@@ -0,0 +1,9 @@
+package astmodel
+
+type Flag string
```
matthchr, line 3: I feel like `Flag` is so generic - these have a targeted use case and wondering if we need a more targeted name as a result?

I guess they are `astmodel` `Flag`'s but still, there are a lot of possible types in `astmodel` which _could_ be flagged?

@Porges had mentioned `Metadata` as an alternative and I think I like something like `ObjectMetadata` better?

hack/generator/pkg/astmodel/resource.go
```diff
@@ -19,71 +19,109 @@ type ResourceType struct {
...
-	result := *definition
+func (resource *ResourceType) MarkAsStorageVersion() *ResourceType {
+	result := resource.copy()
 	result.isStorageVersion = true
-	return &result
+	return result
+}
+
+// AddFlag includes the specified flag on the resource
+func (resource *ResourceType) AddFlag(flag Flag) *ResourceType {
```
matthchr, line 107: `WithFlag` maybe? (I think we use `WithXxx` elsewhere) `AddFlag` makes it sound more like a mutating operation to me.

hack/generator/pkg/astmodel/type_definition.go
```diff
@@ -15,10 +15,18 @@ type TypeDefinition struct {
 	name        TypeName
 	description []string
 	theType     Type
+
+	// flags are used to identify resources during processing;
+	// they have no direct representation in the generated code.
+	flags map[Flag]struct{}
```
matthchr, line 21: You have flags on both `Resource` and `TypeDefinition` - but `TypeDefinition` can be for a `Resource`.

Are `flags` for `Type` implementations, or for `TypeDefinitions`, or for both?

What does it mean to have a `TypeDefinition` whose `Type` _also_ has a flag?

Would it make more sense to put some of the `Flag` methods on the `Type` interface? Or do we only allow them on resources?

hack/generator/pkg/astmodel/type_definition.go
```diff
@@ -68,6 +76,19 @@ func (def TypeDefinition) AsDeclarations(codeGenerationContext *CodeGenerationCo
...
+// AddFlag includes the specified flag on the resource
+func (def TypeDefinition) AddFlag(flag Flag) TypeDefinition {
+	result := def.copy()
+	result.flags[flag] = struct{}{}
+	return result
+}
+
+// HasFlag returns true if the specified flag is present on this resource
+func (def TypeDefinition) HasFlag(flag Flag) bool {
+	_, ok := def.flags[flag]
```
matthchr, line 88: Slightly awkward here as I mentioned above since this `TypeDefinition` can be of a `Type` (`Resource`) that can itself have Flags.

As a caller that feels awkward because then doesn't my check end up something like:
```
func someFunc(types astmodel.Types) (astmodel.Types, error) {
    for typeName, def := range types {
        hasFlag := false
        asResource, ok := def.Type().(*astmodel.Resource)
        if ok {
            hasFlag = asResource.HasFlag("foo")
        }
        hasFlag |= def.HasFlag("foo")
}
```

hack/generator/pkg/astmodel/flags.go
theunrepentantgeek: They're supposed to be generic - and usable across a bunch of different scenarios.

Having `ObjectMetadata` implies fields like `IsAnyType`, `IsOneOfType` and `IsStorageType` which gets away from the idea of a generic mechanism.

hack/generator/pkg/astmodel/resource.go
theunrepentantgeek: `WithFlag()` implies that any other flags are discarded, which we don't want - flags accumulate - which is why I went with `AddFlag()`. 

Agree that it's not ideal, do you have a better suggestion?

hack/generator/pkg/astmodel/type_definition.go
theunrepentantgeek: They're now on `TypeDefinition`, `ObjectType`, and `Resource`, with a level of code duplication that I dislike intensely.

Can't put the methods on the `Type` interface because the return type needs to be the implementation type, not `Type` - otherwise the fluent interface breaks down.

Aside: C# is finally going to address this by allowing covariant return types - as long as the implementation type is more specific, it will match the interface.

hack/generator/pkg/astmodel/type_definition.go
theunrepentantgeek: No, because you know where you added the flag originally. If you added the flag to an `ObjectType`, that's where you check; if you added it to a `Resource`, that's where  you check. 

Flags are a coordination primitive - I'd expect the creator and consumer of a flag to be written by the same person at the same time.

hack/generator/pkg/astmodel/resource.go
matthchr: We use `With` in a number of other places in the code where it's not discarded.

`PropertyDefinition`: `WithValidation`, `WithTag`.
`ObjectType`: `WithProperty` `WithProperties`, `WithFunction`, `WithInterface`
(There's probably more but I didn't bother looking further)

I don't think `With` implies _replacement of the whole property_ - it only implies full replacement if the thing you're operating on is a single value. At least to me, `With` implies that it will return an object with the field requested. If that's in the context of a single value (a string, a type, etc) then yes, it's replacement, but if it's in the context of a map or slice, it's just addition (other elements of the slice/map are preserved, not destroyed).

hack/generator/pkg/astmodel/flags.go
matthchr: Yeah I'm cooling on metadata somewhat just because I don't know what "one metadata" is... whereas `Flag` is clearly the singular of the `Flags` collection.

matthchr: In terms of the level of genericism here, the part that feels awkward to me is that it's not clear at the flag definition point what places the author of the flag intended it to be valid for - but in practice they certainly intend it to be valid on some types but often not all. In order to understand flag usage you'd have to go find all of the locations it's checked in the code (which in practice may not be very difficult?). There's a disconnect between author intent and actual code rendered though which has to be made up by comments or something.

For example, putting `OneOfFlag` on a `TypeName` makes no sense (and wouldn't do anything).
Other flags (such at the storage one) make sense everywhere.

So it feels like flags have a `scope` or something, and we could limit flags to only be used in the scope they're allowed at.

```
type FlagScope int

// TODO: not sure if this is how you do a [flags] enum in Go...
const (
    Object = 0x1,
    Property = 0x2,
    All = Object | Property
    ...
)

type Flag struct {
    Scope FlagScope
    Name string 
}
```

Then individual types ensure they can't be flagged by flags that don't make sense

object_type AddFlag (for example)
```
func (o *ObjectType) AddFlag(flag Flag) *ObjectType {
    if !flag.Scope.HasFlag(Object) { // Here's the trouble with the generic flag name as well I think, because it collides with other contexts that may possibly show up in our code at a later date (such as a bitfield/flags enum).
        panic(fmt.Sprintf("Can't use a flag with %s scope on an Object", flag.Scope))
    }
    result := o.copy()
    result.flags[flag.Name] = flag
    return &result
}
```

This would allow an author to encode expected usage into the definition of the flag. It's still easy to broaden its usage in the future if so desired but it's explicit then.

theunrepentantgeek: You raise some good points - but I fear this is veering into territory where it will just get over complicated. Maybe we need a different solution to "Flags".

hack/generator/pkg/astmodel/resource.go
theunrepentantgeek: You've convinced me. I'll rename it to `WithFlag()`.

hack/generator/pkg/astmodel/flags.go
Porges: > I don't know what "one metadata" is...
A metadatum, obviously.

theunrepentantgeek:
As discussed offline, going to abandon this PR in favour of using decorator types.

PR #227, opened 2020-08-18
Add ResourceListType

matthchr:
- controller-runtime expects we have this type. For an existing
    example look at VirtualMachineList.


matthchr:
It's not 100% clear that a new type was the best way to do this but all the other ways I came up with seemed sorta complicated (is a resourceList a resource? Probably not. Is it just an Object? Maybe but there's no way currently to embed types in our objects so would have to add that).

PR #231, opened 2020-08-19
Improve status info coverage

Porges:
With these changes there are ~~199~~ 24 types with missing status info:

* Make resource name lookup case-insensitive; Swagger has `Mediaservices` but the ARM Schema has `MediaServices` (I’m not sure how this name is being produced).
* Tighten up how we detect ARM resources; we should have been looking at the `PUT` _responses_, not the `PUT` body. 
* Exclude `/default/` part of operation path from resource name.
* Exclude some old versions of services with issues with Swagger specs (missing information or missing version).

`SwaggerTypeExtractor` has moved into `jsonast` as it seems to make more sense there; it uses the `Schema` abstraction types now.

This also fixes the 😱 log message (sadly).

### DIFF
--- a/hack/generator/pkg/jsonast/swagger_type_extractor.go
+++ b/hack/generator/pkg/jsonast/swagger_type_extractor.go
@@ -311,16 +311,22 @@ func inferNameFromURLPath(operationPath string) (string, string, error) {
 		}
 
 		if reading {
-			if urlPart[0] != '{' {
-				name += strings.ToUpper(urlPart[0:1]) + urlPart[1:]
-				skippedLast = false
-			} else {
+			if urlPart == "default" {
+				// skip; shouldn’t be part of name
+				// TODO: I haven’t yet found where this is done in autorest/autorest.armresource to document this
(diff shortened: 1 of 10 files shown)


hack/generator/pkg/jsonast/swagger_type_extractor.go
```diff
@@ -93,6 +117,122 @@ func (extractor *typeExtractor) extractTypes(
...
+
+// a Schema represents an ARM resource if it (or anything reachable via $ref or AllOf)
+// is marked with x-ms-azure-resource, or if it (or anything reachable via $ref or AllOf)
+// has each of the properties: id,name,type.
+func isMarkedAsARMResource(schema Schema) bool {
+	hasID := false
+	hasName := false
+	hasType := false
+
+	var recurse func(schema Schema) bool
```
babbageclunk, line 187: It's weird that this is necessary - I don't understand why they couldn't have made recurse visible inside the function if it was defined with `:=`. :shrug:

babbageclunk (approved):
Looks good!

Issue #211, opened 2020-08-03
Resources should all inherit from ResourceBase


matthchr:
In the [main deployment JSON schema](https://raw.githubusercontent.com/Azure/azure-resource-manager-schemas/master/schemas/2015-01-01/deploymentTemplate.json) there is the following:
```
            "allOf": [
              {
                "$ref": "#/definitions/resourceBase"
              },
              {
                "oneOf": [
                  { "$ref": "https://schema.management.azure.com/schemas/2019-09-16/Microsoft.HealthcareApis.json#/resourceDefinitions/services" },
                  { "$ref": "https://schema.management.azure.com/schemas/2019-02-01-preview/Microsoft.AppConfiguration.json#/resourceDefinitions/configurationStores" },
                  { "$ref": "https://schema.management.azure.com/schemas/2018-08-20-preview/Microsoft.HealthcareApis.json#/resourceDefinitions/services" }
...
```

We aren't honoring the `allOf`:
```
            "allOf": [
              {
                "$ref": "#/definitions/resourceBase"
              },
```

This results in us generating resources which are missing properties which they should have (`name`, `dependsOn`, etc).

PR #233, opened 2020-08-20
Add ability to transform types of properties

Porges:
This is useful to fixup some types that have properties with types that are problematic in CRDs, like `ResourceBase` in the main deployment template.

At the moment `deploymenttemplate` group is excluded by a filter but this will be included soon to address #211.

### DIFF
--- a/hack/generator/pkg/astmodel/array_type.go
+++ b/hack/generator/pkg/astmodel/array_type.go
@@ -6,6 +6,7 @@
 package astmodel
 
 import (
+	"fmt"
 	"go/ast"
 )
 
@@ -59,4 +60,9 @@ func (array *ArrayType) Equals(t Type) bool {
 	}
 
 	return false
+}
+
+// String implements fmt.Stringer
+func (array *ArrayType) String() string {
+	return fmt.Sprintf("[]%s", array.element.String())
 }
--- a/hack/generator/pkg/config/configuration.go
+++ b/hack/generator/pkg/config/configuration.go
@@ -29,7 +29,11 @@ type Configuration struct {
 	// Filters used to control which types are created from the JSON schema
 	TypeFilters []*TypeFilter `yaml:"typeFilters"`
 	// Transformers used to remap types
-	TypeTransformers []*TypeTransformer `yaml:"typeTransformers"`
+	Transformers []*TypeTransformer `yaml:"typeTransformers"`
+
+	// after init TypeTransformers is split into property and non-property transformers
+	typeTransformers     []*TypeTransformer
+	propertyTransformers []*TypeTransformer
 }
 
 // NewConfiguration returns a new empty Configuration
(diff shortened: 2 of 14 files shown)


hack/generator/pkg/astmodel/array_type.go
```diff
@@ -60,3 +61,8 @@ func (array *ArrayType) Equals(t Type) bool {
 
 	return false
 }
+
+// String implements fmt.Stringer
```
matthchr, line 65: Minor: Do we care to add a type assert that we're impling this interface?

matthchr: Same for all below, won't repeat comment

matthchr: Ah, you embedded it into `Type` -- even better.

hack/generator/pkg/config/type_transformer.go
```diff
@@ -17,65 +18,96 @@ import (
 type TransformTarget struct {
 	PackagePath string `yaml:",omitempty"`
 	Name        string `yaml:",omitempty"`
+	Map         *MapType
```
matthchr, line 21: I assume that `Map` is mutually exclusive with `Name` -- do we need to check that somewhere?

hack/generator/pkg/config/configuration.go
```diff
@@ -176,16 +176,63 @@ func (config *Configuration) ShouldPrune(typeName astmodel.TypeName) (result Sho
...
+	Property        astmodel.PropertyName
+	NewPropertyType astmodel.Type
+	Because         string
+}
+
+// TransformTypeProperties applies any property transformers to the type
+func (config *Configuration) TransformTypeProperties(name astmodel.TypeName, objectType *astmodel.ObjectType) *PropertyTransformResult {
+
+	for _, transformer := range config.TypeTransformers {
+		if transformer.propertyRegex != nil { // exclude non-property transformers
```
matthchr, line 203: Are these really two different transformers hidden in one?

hack/generator/pkg/config/type_transformer.go
Porges: Added check for this.

hack/generator/pkg/config/configuration.go
Porges: Maybe in initialize I split them into two sets?

Porges: Done that.

hack/generator/pkg/config/type_transformer.go
```diff
@@ -17,65 +18,110 @@ import (
...
+			result = primitive
+		} else {
+			result = astmodel.MakeTypeName(
+				astmodel.MakePackageReference(target.PackagePath),
+				target.Name)
+		}
+	}
+
+	if target.Map != nil {
+		if result != nil {
```
matthchr, line 62: minor: It might be clearer if rather than the result + check here, you just did:
```
if target.Map != nil && target.Name !=nil {
     return errors.Errorf()
}
```

up at the top?

It saves you a `result` (you can just return inline like you were), and it moves one of the base failure cases to the top?

Porges: yeah this is so much clearer, ugh 😛

PR #235, opened 2020-08-20
Status should be serialized as "status", not "spec"

matthchr:


matthchr:
We should probably make it so that the integration tests can do stuff with Status too. We might have to fiddle with the golden-files infrastructure so that we can do that though (@Porges might have some ideas...).

What I'm thinking of is, an optional `foo_status.json` that pairs with `foo.json` and together can produce `foo.golden`?

PR #238, opened 2020-08-23
Filter out AnyType packages in the check stage

babbageclunk:
This allows the pipeline stage to check for unnecessarily excluded packages so we know to remove a filter when we fix a case that generates AnyTypes in the output.

Move the list of packages with AnyTypes out of the filters and into
a separate configuration item.


Porges (approved):
nice! 🚀

PR #239, opened 2020-08-24
Fix path handling to run on Windows

theunrepentantgeek:
Convert Windows file paths (containing `\`) to the canonical Go form (containing `/`) where necessary.

### DIFF
--- a/hack/generator/pkg/codegen/pipeline_augment_status.go
+++ b/hack/generator/pkg/codegen/pipeline_augment_status.go
@@ -218,8 +218,10 @@ var skipDirectories = []string{
 }
 
 func shouldSkipDir(filePath string) bool {
+	p := filepath.ToSlash(filePath)
+
 	for _, skipDir := range skipDirectories {
-		if strings.Contains(filePath, skipDir) {
+		if strings.Contains(p, skipDir) {
 			return true
 		}
 	}
--- a/hack/generator/pkg/codegen/pipeline_augment_status_test.go
+++ b/hack/generator/pkg/codegen/pipeline_augment_status_test.go
@@ -1,3 +1,8 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ */
+
 package codegen
 
 import (
--- a/hack/generator/pkg/codegen/pipeline_augment_status_test.go
+++ b/hack/generator/pkg/codegen/pipeline_augment_status_test.go
@@ -7,30 +7,47 @@ package codegen
 
 import (
 	. "github.com/onsi/gomega"
-
+	"runtime"
 	"testing"
 )
 
 func Test_ShouldSkipDir_GivenPath_HasExpectedResult(t *testing.T) {
-	cases := []struct {
+	linuxCases := []struct {
 		name       string
 		path       string
 		shouldSkip bool
 	}{
 		// Simple paths
 		{"Root", "/", false},
-		{"Drive", "D:\\", false},
 		{"Top level", "/foo/", false},
-		{"Top level, Windows", "D:\\foo\\", false},
 		{"Nested", "/foo/bar/", false},
-		{"Nested, Windows", "D:\\foo\\bar\\", false},
 		// Paths to skip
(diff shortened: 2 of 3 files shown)


hack/generator/pkg/jsonast/schema_abstraction_openapi.go
```diff
@@ -7,12 +7,12 @@ package jsonast
 
 import (
 	"fmt"
-	"io/ioutil"
-	"net/url"
-
 	"github.com/go-openapi/jsonpointer"
 	"github.com/go-openapi/spec"
 	"github.com/pkg/errors"
+	"io/ioutil"
```
babbageclunk, line 13: These imports should be at the top - the general pattern is one block for standard lib packages, one for external packages, one for packages in the current project. Might be an IDE setup issue?

hack/generator/pkg/codegen/pipeline_augment_status_test.go
```diff
@@ -0,0 +1,36 @@
...
+	}{
+		// Simple paths
+		{"Root", "/", false},
+		{"Top level", "/foo/", false},
+		{"Nested", "/foo/bar/", false},
+		// Paths to skip
+		{"Skip top level", "/examples/", true},
+		{"Skip nested", "/foo/examples/", true},
+		{"Skip nested, trailing directory", "/foo/examples/bar/", true},
+	}
```
babbageclunk, line 23: Could you add some test cases for Windows-y paths? Or would those fail on non-Windows machines?

babbageclunk (approved):
Nice fix! Couple of minor comments.

hack/generator/pkg/codegen/pipeline_augment_status_test.go
theunrepentantgeek: Good idea, added.

hack/generator/pkg/jsonast/schema_abstraction_openapi.go
theunrepentantgeek: I didn't change them manually. I'll put them back. GoLand uses `go fmt` so I assume the changes were done by that.

hack/generator/pkg/codegen/pipeline_augment_status.go
```diff
@@ -218,8 +218,10 @@ var skipDirectories = []string{
 }
 
 func shouldSkipDir(filePath string) bool {
+	p := filepath.ToSlash(filePath)
```
matthchr, line 221: minor: Is this worth a comment? Something like...
`Normalize path to use "/" as directory separator to avoid issues with skipped directories differing on Windows vs Linux`?

hack/generator/pkg/codegen/pipeline_augment_status_test.go
theunrepentantgeek: I've added them - but they fail when run on Linux, so I've gated them to only run on Windows. #sigh

hack/generator/pkg/jsonast/schema_abstraction_openapi.go
Porges: my formatter always puts stdlib packages at the top. is there some kind of config option?

hack/generator/pkg/codegen/pipeline_augment_status.go
Porges: an alternate fix would be to call `filepath.ToSlash` on each case in `skipDirectories`; less work overall

hack/generator/pkg/jsonast/schema_abstraction_openapi.go
Porges: looks like the VSCode formatter is `goimports` not just `go fmt`

Porges: GoLand configuration here: https://www.jetbrains.com/help/go/integration-with-go-tools.html#goimports

Porges: and automatic run setting: https://stackoverflow.com/a/45590498/10311

PR #240, opened 2020-08-25
Remove ResourceListType

matthchr:
- It's generated in ResourceType instead.


matthchr:
Much better - thanks for the suggestion @Porges

Porges (approved):
Looks good 👍

PR #242, opened 2020-08-26
Resources should be able to implement interfaces

matthchr:
- Added InterfaceImplementer, for sharing interface implementation code between ResourceType and ObjectType (and possibly others in the future).
- Add KubernetesResource interface and implementation. This will be needed by the controller.
- A few minor cleanups for generated ConvertToArm methods.

### DIFF
--- a/hack/generator/pkg/astmodel/interface_implementer.go
+++ b/hack/generator/pkg/astmodel/interface_implementer.go
@@ -1,0 +1,168 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ */
+
+package astmodel
+
+import (
+	"go/ast"
+	"go/token"
+	"sort"
+)
+
+// TODO: need to add tests for this probably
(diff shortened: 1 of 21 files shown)


hack/generator/pkg/astmodel/interface_implementer.go
```diff
@@ -0,0 +1,176 @@
...
+	return true
+}
+
+// TODO: Do we actually want to impl this? I kinda feel like we actually don't?
+// TODO: maybe we want an interface with _most_ of these methods but not AsType?
+// TODO: my primary objective with implementing this here is not necessarily to allow InterfaceImplementer
+// TODO: to be able to be used as a Type but rather to ensure that it has a common look/feel
+// TODO: with other astmodel types in terms of how imports are managed, etc. It probably shouldn't
+// TODO: actually be possible to use it as a type though...
+var _ Type = InterfaceImplementer{}
```
Porges, line 119: agree with not implementing this 👍

Porges (approved):

PR #243, opened 2020-08-27
Add generation of ArmResourceSpec interface

matthchr:
- Fix up a few package exclusions to happen a bit earlier (cleaning up some hacky code we had a TODO about).

### DIFF
--- a/hack/generated/pkg/genruntime/base_types.go
+++ b/hack/generated/pkg/genruntime/base_types.go
@@ -26,3 +26,12 @@ type KubernetesResource interface {
 	// AzureName returns the Azure name of the resource
 	AzureName() string
 }
+
+// ArmResourceSpec is an ARM resource specification
+type ArmResourceSpec interface {
+	GetApiVersion() string
+
+	GetType() string
+
+	GetName() string
+}
--- a/hack/generated/pkg/genruntime/base_types.go
+++ b/hack/generated/pkg/genruntime/base_types.go
@@ -27,7 +27,9 @@ type KubernetesResource interface {
 	AzureName() string
 }
 
-// ArmResourceSpec is an ARM resource specification
+// ArmResourceSpec is an ARM resource specification. This interface contains
+// methods to access properties common to all ARM Resource Specs. An Azure
+// Deployment is made of these.
 type ArmResourceSpec interface {
 	GetApiVersion() string
 
--- a/hack/generator/pkg/astmodel/resource.go
+++ b/hack/generator/pkg/astmodel/resource.go
@@ -47,7 +47,6 @@ func NewAzureResourceType(specType Type, statusType Type, typeName TypeName) *Re
 		var apiVersionProperty *PropertyDefinition
 		isNameOptional := false
 		isTypeOptional := false
-		apiVersionOptional := false
 		for _, property := range objectType.Properties() {
 			if property.HasName(NameProperty) {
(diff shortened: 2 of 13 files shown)


hack/generator/pkg/astmodel/resource.go
```diff
@@ -81,12 +83,25 @@ func NewAzureResourceType(specType Type, statusType Type, typeName TypeName) *Re
...
+			panic(fmt.Sprintf("Resource %s is missing apiVersion property", typeName))
+		}
+
 		if isNameOptional {
 			// Fix name to be required -- again this is an artifact of bad spec more than anything
 			nameProperty = nameProperty.MakeRequired()
 			objectType = objectType.WithProperty(nameProperty)
 		}
 
+		// Fix APIVersion to be optional. Technically this isn't due to a bad specification, but in our
```
Porges, line 96:
```suggestion
		// Fix APIVersion to be required. Technically this isn't due to a bad specification, but in our
```

hack/generator/pkg/codegen/pipeline_create_arm_types.go
```diff
@@ -48,15 +48,17 @@ func createArmTypesAndCleanKubernetesTypes(idFactory astmodel.IdentifierFactory)
 	}
 }
 
-func createArmTypes(definitions astmodel.Types) (astmodel.Types, astmodel.Types, error) {
+func createArmTypes(
+	definitions astmodel.Types,
+	idFactory astmodel.IdentifierFactory) (astmodel.Types, astmodel.Types, error) {
```
Porges, line 53: it might be nice to definer a receiver type with `definitions` & `idFactory` on it.

Porges (approved):

hack/generated/pkg/genruntime/base_types.go
```diff
@@ -26,3 +26,12 @@ type KubernetesResource interface {
 	// AzureName returns the Azure name of the resource
 	AzureName() string
 }
+
+// ArmResourceSpec is an ARM resource specification
```
babbageclunk, line 30: I feel like this could be fleshed out - maybe say how it's used?

hack/generator/pkg/astmodel/resource.go
```diff
@@ -81,12 +83,25 @@ func NewAzureResourceType(specType Type, statusType Type, typeName TypeName) *Re
...
+		// case forcing it to required makes our lives simpler (and the vast majority of resources specify
+		// it as required anyway). The only time it's allowed to be optional is if you set apiProfile on
+		// the ARM template instead, which we never do.
+		if apiVersionOptional {
```
babbageclunk, line 100: If we always want `apiVersion` to be required it would be clearer to do it unconditionally, rather than checking to see if it's optional and only updating it when it is.

babbageclunk (approved):
Nice!

hack/generator/pkg/astmodel/resource.go
matthchr: Yes, yes it would.

Will fix!

hack/generator/pkg/codegen/pipeline_create_arm_types.go
matthchr: Agree - did this and reorganized this file -- will send it in a separate PR once I coordinate with @theunrepentantgeek

PR #247, opened 2020-09-02
Introduce ArmType as a wrapper to make ARM types easily identifiable

theunrepentantgeek:
The new `ArmType` can be used to wrap any existing `ObjectType` to flag it as specifically for ARM use.

As the upcoming work for Storage types will need to do a similar thing, I've taken an extensible approach instead of hard coding each case. `TypeVisitor` instances are used as they already know how to deeply introspect a complex type and make precise changes if needed. This makes code more flexible as it can handle both `TypeDefinition{ObjectType}` and `TypeDefinition{ArmType{ObjectType}}` without modification.

Testing: Output into `\apis\` is identical, save for a change I made in `type_visitor.go` where `TypeDefinition` descriptions are now preserved.


hack/generator/pkg/astmodel/type_visitor.go
```diff
@@ -155,5 +158,18 @@ func MakeTypeVisitor() TypeVisitor {
...
 		},
+		VisitArmType: func(this *TypeVisitor, at *ArmType, ctx interface{}) (Type, error) {
+			newType, err := this.Visit(&at.objectType, ctx)
+			if err != nil {
+				return nil, errors.Wrapf(err, "failed to visit ARM underlying type %v", at.objectType)
+			}
+
+			ot, ok := newType.(*ObjectType)
+			if !ok {
+				return nil, errors.Wrapf(err, "expected transformation of ARM underlying type %s to return ObjectType, not %v", at.objectType, ot)
```
babbageclunk, line 169: This should be `errors.Errorf` rather than `errors.Wrapf` - `Wrapf` returns nil if the error  being wrapped is nil, which it always will be here.

hack/generator/pkg/codegen/pipeline_create_arm_types.go
```diff
@@ -123,17 +118,18 @@ func modifyKubeTypes(
...
+			}
+
+			return *modifiedDef, nil
 		})
 }
 
 func iterDefs(
 	definitions astmodel.Types,
-	resourceHandler func(name astmodel.TypeName, resourceType *astmodel.ResourceType) (astmodel.TypeName, astmodel.TypeDefinition, error),
+	resourceHandler func(name astmodel.TypeName, resourceType *astmodel.ResourceType) (*astmodel.TypeName, *astmodel.TypeDefinition, error),
```
babbageclunk, line 134: Why change the name and definition returns to pointers here? It doesn't seem like you use it for anything (the place that calls `resourceHandler` just derefs them immediately), and the difference between the `resourceHandler` and `otherDefsHandler` is confusing. 

If it's just to simplify the error return lines, another way could be to add
```go
var noName astmodel.TypeName
var noDef astmodel.TypeDefinition
```
...at the top of the func and then use `return noName, noDef, errors.Wrapf(....)` on the error lines.

hack/generator/pkg/codegen/pipeline_create_arm_types.go
```diff
@@ -196,22 +192,25 @@ func transformTypeDefinition(
 	def astmodel.TypeDefinition,
 	handlers []conversionHandler) (astmodel.TypeDefinition, error) {
 
-	originalType, ok := def.Type().(*astmodel.ObjectType)
-	if !ok {
-		return astmodel.TypeDefinition{}, errors.Errorf("input type %q (%T) was not of expected type Object", def.Name(), def.Type())
+	visitor := astmodel.MakeTypeVisitor()
+	visitor.VisitObjectType = func(this *astmodel.TypeVisitor, it *astmodel.ObjectType, ctx interface{}) (astmodel.Type, error) {
```
babbageclunk, line 198: Should this keep track of whether it actually found any object type in the definition? The old version of the code would return an error if called on a non-object type definition but I think this would quietly succeed without applying the handlers. (Which might be fine, not sure!)

hack/generator/pkg/codegen/pipeline_create_arm_types.go
```diff
@@ -320,24 +319,26 @@ func addArmConversionInterface(
 	kubeDef astmodel.TypeDefinition,
 	armDef astmodel.TypeDefinition,
 	idFactory astmodel.IdentifierFactory,
-	isResource bool) (astmodel.TypeDefinition, error) {
+	isResource bool) (*astmodel.TypeDefinition, error) {
```
babbageclunk, line 340: Same question here? The pointer return type just seems to make the return here more awkward (need to assign to a local to take a reference) and add a dereference at the call site.

babbageclunk (approved):
This looks good - I like the use of the visitor. There's one place that needs to create a new error rather than wrapping since the wrapped error is nil. Also a couple of questions about why you're changing value returns to pointer returns - maybe I'm missing something about why they should be pointers?

hack/generator/pkg/astmodel/arm_type.go
```diff
@@ -0,0 +1,78 @@
...
+	return fmt.Sprintf("ARM(%v)", at.objectType)
+}
+
+// ObjectType returns the underlying object type of the arm type
+func (at ArmType) ObjectType() ObjectType {
+	return at.objectType
+}
+
+// IsArmType returns true if the passed type is a Arm type; false otherwise.
+// We need to handle cases where it might be wrapped, so this isn't just a simple type check, we need to walk the top
```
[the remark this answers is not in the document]
theunrepentantgeek: Whoops. Comment is from an over complex solution I backed away from. Removed.

hack/generator/pkg/astmodel/arm_type.go
```diff
@@ -0,0 +1,78 @@
...
+ */
+
+package astmodel
+
+import (
+	"fmt"
+	"go/ast"
+)
+
+// ArmType wraps an existing type to indicate that it is an ARM targetted variation
```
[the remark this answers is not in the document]
theunrepentantgeek: Fixed.

hack/generator/pkg/astmodel/type.go
```diff
@@ -43,3 +44,69 @@ func TypeEquals(left, right Type) bool {
...
+
+	ot, ok := rt.(*ObjectType)
+	if !ok {
+		return nil, fmt.Errorf("Type %v is not an object", rt)
+	}
+
+	return ot, nil
+}
+
+var cachedAsObjectTypeVisitor = createAsObjectTypeVisitor()
```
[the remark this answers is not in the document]
theunrepentantgeek: I'm expecting `AsObjectType()` to be called a lot - and while the construction isn't expensive in time, it is expensive in memory; caching it should help prevent GC churn (though, to be fair, I didn't measure this)

hack/generator/pkg/astmodel/type_visitor.go
theunrepentantgeek: Good catch, thanks.

hack/generator/pkg/codegen/pipeline_create_arm_types.go
theunrepentantgeek: Yeah, mostly a style issue - returning empty structs just feels as though the type system is forcing us to return _something_ when we really want to return _nothing_. 

At the risk of arguing by analogy, it's like going to the supermarket to buy a packet of chips and when your card is declined, you end up walking out with an empty packet instead of nothing at all.

hack/generator/pkg/codegen/pipeline_create_arm_types.go
theunrepentantgeek: It should error if no change is made; I'll add in a check - thanks for catching this.

hack/generator/pkg/codegen/pipeline_create_arm_types.go
theunrepentantgeek: Reverted.

hack/generator/pkg/astmodel/arm_type.go
```diff
@@ -0,0 +1,76 @@
...
+	"go/ast"
+)
+
+// ArmType wraps an existing type to indicate that it is an ARM targeted variation
+type ArmType struct {
+	objectType ObjectType
+}
+
+// ArmType is a Type
+var _ Type = &ArmType{}
```
matthchr, line 19: Is there a mismatch here? We are using `ArmType` as a value most places it seems yet here we assert its `ptr` impls the interface.

I think technically this is fine (actually maybe this is correct). Do any of you know if we should be asserting this matching the receiver type (which would suggest this should then be: `var _ Type = ArmType{}`), or should it always be ptr?

I observe that other types that use a value for receiver (i.e. `TypeName`) do: `var _ Type = TypeName{}`

hack/generator/pkg/astmodel/arm_type.go
```diff
@@ -0,0 +1,76 @@
...
+	return at.objectType
+}
+
+// IsArmType returns true if the passed type is a Arm type; false otherwise.
+func IsArmType(t Type) bool {
+	_, ok := t.(ArmType)
+	return ok
+}
+
+// IsArmDefinition returns true if the passed definition is for a Arm type; false otherwise.
```
matthchr, line 73:
```suggestion
// IsArmDefinition returns true if the passed definition is for an Arm type; false otherwise.
```

hack/generator/pkg/codegen/pipeline_create_arm_types.go
```diff
@@ -60,12 +61,12 @@ func createArmTypes(
 		func(name astmodel.TypeName, resourceType *astmodel.ResourceType) (astmodel.TypeName, astmodel.TypeDefinition, error) {
 			armSpecDef, kubeSpecName, err := createArmResourceSpecDefinition(definitions, resourceType, idFactory)
 			if err != nil {
-				return astmodel.TypeName{},
-					astmodel.TypeDefinition{},
-					errors.Wrapf(err, "unable to create arm resource spec definition for resource %s", name)
+				nilName := astmodel.TypeName{}
```
matthchr, line 64: minor: `emptyName` + `emptyDef` might be more technically accurate names?

hack/generator/pkg/astmodel/type.go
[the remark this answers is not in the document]
matthchr: FWIW I had the same thought -- not sure the global state is worth the perf improvement

hack/generator/pkg/codegen/pipeline_create_arm_types.go
```diff
@@ -198,22 +199,29 @@ func transformTypeDefinition(
 	def astmodel.TypeDefinition,
 	handlers []conversionHandler) (astmodel.TypeDefinition, error) {
 
-	originalType, ok := def.Type().(*astmodel.ObjectType)
-	if !ok {
-		return astmodel.TypeDefinition{}, errors.Errorf("input type %q (%T) was not of expected type Object", def.Name(), def.Type())
+	if !astmodel.IsObjectType(def.Type()) {
```
matthchr, line 202: This allows ARM types, which seems wrong since we don't want to allow ARM -> ARM transformations?

hack/generator/pkg/codegen/pipeline_create_arm_types.go
```diff
@@ -198,22 +199,29 @@ func transformTypeDefinition(
...
 	}
 
-	resultType := originalType
-	var err error
-	for _, handler := range handlers {
-		resultType, err = handler(resultType)
-		if err != nil {
-			return astmodel.TypeDefinition{}, err
+	visitor := astmodel.MakeTypeVisitor()
+	visitor.VisitObjectType = func(this *astmodel.TypeVisitor, it *astmodel.ObjectType, ctx interface{}) (astmodel.Type, error) {
```
matthchr, line 207: I am not sure I agree that it makes sense to use a visitor here?

Effectively what you've done (I think) is:

1. Asserted we have an object or ARM type (`astmodel.IsObjectType`) above.
2. It's practically speaking always an `astmodel.ObjectType`
3. Created a visitor and overridden the visit func for object type so that it doesn't call `visit`... effectively meaning we aren't traversing anything with the visitor (I think?)

So it ends up feeling like this visitor is adding complexity but not really actually doing anything - My understanding (and I could be missing something here) is that the visitor is for managing an arbitrary `Type`. Since in this case we know what we have (and we have defined custom handlers for them which don't use the visitor framework) the visitor isn't really helping us?

hack/generator/pkg/codegen/pipeline_create_arm_types.go
```diff
@@ -340,22 +350,25 @@ func addArmConversionInterface(
 	idFactory astmodel.IdentifierFactory,
 	isResource bool) (astmodel.TypeDefinition, error) {
 
-	armObjectType, ok := armDef.Type().(*astmodel.ObjectType)
-	if !ok {
-		return astmodel.TypeDefinition{}, errors.Errorf("arm def %q was not of type object", armDef.Name())
+	objectType, err := astmodel.TypeAsObjectType(armDef.Type())
+	if err != nil {
+		nilDef := astmodel.TypeDefinition{}
```
matthchr, line 355: minor: same comment on empty vs nil

hack/generator/pkg/codegen/pipeline_create_arm_types.go
```diff
@@ -340,22 +350,25 @@ func addArmConversionInterface(
...
 		return t.WithInterface(armconversion.NewArmTransformerImpl(
 			armDef.Name(),
-			armObjectType,
+			objectType,
 			idFactory,
 			isResource)), nil
 	}
 
-	return transformTypeDefinition(
+	transformedDef, err := transformTypeDefinition(
```
matthchr, line 367: minor: Why assign and then return here rather than just return the result of the call like before?

hack/generator/pkg/codegen/pipeline_create_arm_types.go
matthchr: Might be better to have an `IsObjectType` (used here) and an `IsObjectOrArmType` (used elsewhere where you want to treat them the same - not exactly sure where that is).

matthchr: Actually on further inspection I only see this used here - why have the duality at all then (Where an ARMType `IsObjectType`?)

hack/generator/pkg/codegen/pipeline_create_arm_types.go
matthchr: When I wrote this bit I think I played around with trying to use visitor for the handlers but couldn't get it to work cleanly... I recall having the argument in my head for why it didn't make sense, but of course I didn't write it down and now I don't remember it =P

theunrepentantgeek: The content of the `TypeDefinition` may be either an `ObjectType` _or_ an `ArmType`. I tried writing the code to handle both situations in this method and it got tremendously convoluted. The `TypeVisitor` already knows how to unwrap/rewrap a definition (and an `ArmType`) so reusing that to make the transformation allowed the code to stay relatively easy to read.

PR #249, opened 2020-09-03
Improve TypeVisitor and other cleanup

theunrepentantgeek:
A number of minor code gardening changes

* `MakeTypeVisitor()` now references named functions, giving better clarity of code than having them inline
* Changed naming of empty values to use `empty` prefix instead of `nil` (followup on PR #247)
* Simplified `IsObjectType()` to have a single responsibility and updated usage where an `ArmType` may reasonably appear
* Fixed bug in `IsObjectDefinition()`


matthchr (approved):

matthchr:
LGTM

PR #253, opened 2020-09-04
Initial thoughts on generating Crossplane provider

matthchr:


babbageclunk (approved):
This all sounds very sensible to me, thanks for writing it up Matt! I'm going to read some more Crossplane docs to try to get a clearer picture of how the higher-level parts work.

hack/generator/spec/crossplane.md
```diff
@@ -0,0 +1,71 @@
+# Thoughts on performing provider-azure code generation for Crossplane
+
+The patterns identified below were extrapolated from [the Crossplane VNET spec](https://github.com/crossplane/provider-azure/blob/master/config/crd/network.azure.crossplane.io_virtualnetworks.yaml).
+
+## General notes
+
+### Crossplane has the following static "special" properties
+  1. ClaimReference
```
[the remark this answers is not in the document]
matthchr: Yeah that's probably worth noting -- I assume we'd avoid using it in the generator then since for new resources my bet is that those fields shouldn't be there (I think that they had confirmed this in a meeting some time back?)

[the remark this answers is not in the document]
matthchr: I see where you see `ClaimRef`, `ClassRef`, and `ReclaimPolicy` are all deprecated. Where did you see `ProviderRef` was deprecated... and if it's deprecated is `ProviderConfigRef` as well?

hack/generator/spec/crossplane.md
```diff
@@ -0,0 +1,71 @@
...
+
+### VNET
+As mentioned above, our VNET today has `subnets` and `virtualNetworkPeerings`, which are really subresources. It also has `provisioningState`.
+
+**If we excluded these properties in the configuration file, we could definitely hack towards generating a VNET that looks like Crossplane's now**
+
+### Microsoft SQL
+
+The strange thing with SQL is somehow some of their APIs don't have the full API surface area... For example I think `2019-06-01-preview` exists but it's not in the JSON schema we're using. Other bits are split across API versions but then unioned together in their SDK:
+[2015-05-01-preview](https://schema.management.azure.com/schemas/2015-05-01-preview/Microsoft.Sql.json#/resourceDefinitions) has servers, and [2017-03-01-preview](https://schema.management.azure.com/schemas/2017-03-01-preview/Microsoft.Sql.json#/resourceDefinitions) has databases.
```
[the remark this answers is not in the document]
matthchr: The `v3` version of the SQL SDK uses the `2019-06-01-preview` API for server create, and the `2017-10-01-preview` API for database create.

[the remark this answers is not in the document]
matthchr: I've asked our SQL contact about this, as it does seem a bit weird. Although my bet is it has to do with the fact that technically Server and DB are different RPs whose APIs are versioned separately?

hack/generator/spec/crossplane.md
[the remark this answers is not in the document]
matthchr: Found it

hack/generator/spec/crossplane.md
```diff
@@ -0,0 +1,71 @@
+# Thoughts on performing provider-azure code generation for Crossplane
+
+The patterns identified below were extrapolated from [the Crossplane VNET spec](https://github.com/crossplane/provider-azure/blob/master/config/crd/network.azure.crossplane.io_virtualnetworks.yaml).
```
negz, line 3: I'm afraid VNets might not be the best representation of our current CR schema patterns, which are best represented by v1beta1 resources. That said, I _think_ the key difference is probably the use of `spec.forProvider` and `spec.atProvider` as containers for the objects that map to the Azure API in v1beta1 resources.

https://github.com/crossplane/provider-azure/blob/master/config/crd/cache.azure.crossplane.io_redis.yaml

hack/generator/spec/crossplane.md
```diff
@@ -0,0 +1,71 @@
...
+  7. writeConnectionSecretToRef
+
+Since these are statically shaped they should be relatively easy to add with Crossplane specific generator pipeline stages.
+
+### Parent references in Crossplane are really a discriminated union of 3 options:
+1. XName (e.g. ResourceGroupName)
+2. XNameRef (e.g. ResourceGroupNameRef)
+3. XNameSelector (e.g. ResourceGroupNameSelector)
+
+We probably will eventually want to understand how `XName` differs from `XNameRef` (I think I understand the `XNameSelector`).
```
negz, line 23: We discussed this a little out of band, but the idea is:

* `X` - An actual field that exists in the Azure API. Can be set to any string you'd post to the Azure API.
* `XRef` - A reference to a named Kubernetes object of a preordained kind. Used to resolve a value for `X`.
* `XSelector` - A selector used to set `XRef`.

Note that these values cascade down from the selector, to the ref, to the underlying field. If the ref is already set the selector is ignored, and if the field is already set the ref is ignored.

In practice I think it's fair to call these parent references, but note that they don't _have_ to be. We call these "cross resource references" - the idea is that where they exist they can be used to derive a field value by inspecting another resource. Note also that the derived field doesn't necessarily have the suffix `Name` - that just happens to be the convention in the Azure API.

negz: Some breadcrumbs for the logic we use today; https://github.com/crossplane/provider-azure/blob/64dc01e/apis/cache/v1beta1/references.go#L29

hack/generator/spec/crossplane.md
```diff
@@ -0,0 +1,71 @@
...
+1. XName (e.g. ResourceGroupName)
+2. XNameRef (e.g. ResourceGroupNameRef)
+3. XNameSelector (e.g. ResourceGroupNameSelector)
+
+We probably will eventually want to understand how `XName` differs from `XNameRef` (I think I understand the `XNameSelector`). 
+
+Generating these might take a bit of work as we need to derive the name(s) for each of these for each of the parent resources. This should be doable in a pipeline stage by walking up the graph of owners that we already have and getting their names, then generating a set of these for each owner.
+
+### Status 
+Our status's include the "full" shape of the object, whereas Crossplane's seem to be limited (they don't duplicate the spec, or include other server-derived fields). We should ask if this is an intentional design decision or just something they did in the interest of time.
```
negz, line 28: Per out of band conversation, this decision was intentional though somewhat controversial. :)

There' some context on how we decide what goes into spec vs status in our [API design doc](https://github.com/crossplane/crossplane/blob/a3425d5a/design/one-pager-managed-resource-api-design.md#high-fidelity). In general if an API caller can ever set a field it goes into `spec`, and if they can never set a field it goes into `status`. Spec fields that a caller _could_ set but that are defaulted by Azure when left unset are [late initialized](https://github.com/kubernetes/community/blob/b43d5a2/contributors/devel/sig-architecture/api-conventions.md#late-initialization).

This does mean it's possible that our desired state could be different from actual state and we won't show it, per https://github.com/crossplane/crossplane/issues/925. Including the entire resource in status might help this, but it seems unconventional for Kubernetes resources to do so.

negz (commented):
Thanks for looking into this! We discussed a bunch of this on Slack, but I wanted to make sure the information is here too for future discoverability.

hack/generator/spec/crossplane.md
matthchr: > Note that these values cascade down from the selector, to the ref, to the underlying field. If the ref is already set the selector is ignored, and if the field is already set the ref is ignored.

@ngez just clarifying, this means that if only the `XSelector` is specified, the `XRef` and `X` fields will be visible by the customer when they examine the object? Or will they be empty?

hack/generator/spec/crossplane.md
matthchr: Thanks for this clarification. I've updated the document some with this information.

PR #255, opened 2020-09-11
Create Storage Types

theunrepentantgeek:
First stage of implementation of storage versioning, creating the objects themselves.

Includes fixup of the previous changes I made to introduce ArmType, improvements to test coverage, and some code gardening.

Conversion between types and tests of those conversions to follow in a separate PR (or two or three)


hack/generator/pkg/astmodel/resource.go
```diff
@@ -218,7 +222,6 @@ func (definition *ResourceType) RequiredImports() []PackageReference {
 
 	typeImports = append(typeImports, MetaV1PackageReference)
 	typeImports = append(typeImports, MakeGenRuntimePackageReference())
-	typeImports = append(typeImports, MakePackageReference("fmt"))
```
theunrepentantgeek, line 221: Removing this hard coded reference needed so the new storage types compile; the Arm functions generate the reference they need anyway.

hack/generator/pkg/codegen/pipeline_create_storage_types.go
```diff
@@ -0,0 +1,212 @@
...
+	property *astmodel.PropertyDefinition
+}
+
+func makeStorageTypesVisitorContext() StorageTypesVisitorContext {
+	return StorageTypesVisitorContext{}
+}
+
+func (context StorageTypesVisitorContext) forObject(object *astmodel.ObjectType) StorageTypesVisitorContext {
+	context.object = object
+	context.property = nil
```
matthchr, line 205: surprised this sets to `nil`?

Just wondering what context it wasn't `nil` and needs to be?

matthchr: Is it expected that this context only sets object + property once? If so, could panic if that were violated.

hack/generator/pkg/codegen/pipeline_create_storage_types.go
```diff
@@ -0,0 +1,212 @@
...
+			p = p.WithType(astmodel.StringType)
+		}
+	}
+
+	return p
+}
+
+// mapTypeName maps an existing type name into the right package for the matching storage type
+// Returns the original instance for any type that does not need to be mapped to storage
+func (factory *StorageTypeFactory) mapTypeName(name astmodel.TypeName) (*astmodel.TypeName, error) {
```
matthchr, line 156: minor: usage of `*astmodel.TypeName` rather than `astmodel.TypeName` seems unnessecary, as discussed previously `return astmodel.TypeName{}, err` (while icky) is a thing in Go and it seems like it was only promoted to a ptr here to avoid that?

This may be related to one of the comments I made above about why we tolerate `*astmodel.TypeName` when we hadn't previously.

matthchr, line 156: minor: Would `mapTypeNameToStoragePackage` or something be more clear about the intent of this function?

hack/generator/pkg/codegen/pipeline_create_storage_types.go
```diff
@@ -0,0 +1,212 @@
...
+
+type StorageTypeFactory struct {
+	types astmodel.Types
+}
+
+func (factory *StorageTypeFactory) visitTypeName(_ *astmodel.TypeVisitor, name astmodel.TypeName, ctx interface{}) (astmodel.Type, error) {
+	vc := ctx.(StorageTypesVisitorContext)
+
+	// Resolve the type name to the actual referenced type
+	actualDefinition, actualDefinitionFound := factory.types[name]
```
matthchr, line 88: Is `actualDefinitionFound` false actually a fatal error here?

Could this be:
```
 	// Check for property specific handling
	if vc.property != nil {
               actualDefinition, actualDefinitionFound := factory.types[name]
               if !actualDefinitionFound {
                    // return err or panic
                }
		if et, ok := actualDefinition.Type().(*astmodel.EnumType); ok {
			// Property type refers to an enum, so we use the base type instead
			return et.BaseType(), nil
		}
	}
```

hack/generator/pkg/codegen/pipeline_create_storage_types.go
```diff
@@ -0,0 +1,212 @@
...
+}
+
+func (factory *StorageTypeFactory) makeStorageProperty(prop *astmodel.PropertyDefinition, propertyType astmodel.Type) *astmodel.PropertyDefinition {
+	p := prop.WithType(propertyType).
+		MakeOptional().
+		WithoutValidation().
+		WithDescription("")
+
+	// If the property is *string we can simplify it to string
+	// This makes other generated code simpler
```
matthchr, line 144: Why is this safe? What about services where there is a difference between "" and `nil`? If we don't preserve that information in the storage type then we'll have a problem, yes?

hack/generator/pkg/codegen/pipeline_create_storage_types.go
```diff
@@ -0,0 +1,212 @@
...
+	storagePackage, err := astmodel.CreateStoragePackageReference(name.PackageReference)
+	if err != nil {
+		return nil, err
+	}
+
+	newName := astmodel.MakeTypeName(storagePackage, name.Name())
+	return &newName, nil
+}
+
+func (factory *StorageTypeFactory) visitArmType(
```
matthchr, line 171: Does this function ever get called?

You already skipped arm types in the loop, so how would a visit function run into an ARM type? I suppose it would only be possible if a non-ARM type had a property that was an ARM type (AFAIK that doesn't happen?)

is that actually an error condition?

hack/generator/pkg/codegen/pipeline_create_storage_types.go
matthchr: Same with `forProperty`

hack/generator/pkg/astmodel/type_definition_test.go
```diff
@@ -78,3 +79,139 @@ func createStringProperty(name string, description string) *PropertyDefinition {
 func createIntProperty(name string, description string) *PropertyDefinition {
 	return NewPropertyDefinition(PropertyName(name), name, IntType).WithDescription(description)
 }
+
+/*
+ * ApplyObjectTransformation() tests
+ */
+
+func TestApplyObjectTransformation_GivenObjectAndTransformation_AppliesTransformation(t *testing.T) {
```
matthchr, line 87: Nice

hack/generator/pkg/astmodel/arm_type.go
```diff
@@ -19,34 +19,34 @@ type ArmType struct {
 var _ Type = &ArmType{}
 
 // MakeArmType wraps an object type to indicate it's an ARM targeted variation
-func MakeArmType(object ObjectType) ArmType {
-	return ArmType{
+func MakeArmType(object ObjectType) *ArmType {
```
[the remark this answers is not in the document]
theunrepentantgeek: I encountered problems caused by implementing `Type` with `ArmType`, so I switched to implementing it with `*ArmType`.

hack/generator/pkg/codegen/code_generator.go
```diff
@@ -70,13 +71,14 @@ func (generator *CodeGenerator) Generate(ctx context.Context) error {
 	defs := make(astmodel.Types)
 	for i, stage := range generator.pipeline {
 		klog.V(0).Infof("Pipeline stage %d/%d: %s", i+1, len(generator.pipeline), stage.description)
-		updatedDefs, err := stage.Action(ctx, defs)
+		// Defensive copy so that we can compare types in vs out
+		defsOut, err := stage.Action(ctx, defs.Copy())
```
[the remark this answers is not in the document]
theunrepentantgeek: Yeah, it's a defensive copy because some of our stages modify the incoming set of values. Rather than allowing that to become a bug, I figured it was better to ensure it was never a problem.

hack/generator/pkg/codegen/pipeline_create_storage_types.go
```diff
@@ -0,0 +1,212 @@
...
+
+				def, err := visitor.VisitDefinition(d, vc)
+				if err != nil {
+					errs = append(errs, err)
+					continue
+				}
+
+				finalDefn := def.WithDescription(descriptionForStorageVariant(d))
+				storageTypes[finalDefn.Name()] = finalDefn
+
```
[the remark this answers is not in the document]
theunrepentantgeek: Removed extra blank line.

hack/generator/pkg/codegen/pipeline_create_storage_types.go
```diff
@@ -0,0 +1,212 @@
...
+func makeStorageTypesVisitor(types astmodel.Types) astmodel.TypeVisitor {
+	factory := &StorageTypeFactory{
+		types: types,
+	}
+
+	result := astmodel.MakeTypeVisitor()
+	result.VisitTypeName = factory.visitTypeName
+	result.VisitObjectType = factory.visitObjectType
+	result.VisitArmType = factory.visitArmType
+	//result.VisitResourceType = factory.visitResourceType
```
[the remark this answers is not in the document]
theunrepentantgeek: Don't need separate handling for Resource types; have removed the commented code.

hack/generator/pkg/codegen/pipeline_create_storage_types.go
theunrepentantgeek: It's not a fatal error because it might be referring to an imported type (not that I'm whether that's happening yet).

hack/generator/pkg/codegen/pipeline_create_storage_types.go
```diff
@@ -0,0 +1,212 @@
...
+		return nil, errors.Wrapf(err, "failed to map name into storage namespace")
+	}
+
+	return visitedName, nil
+}
+
+func (factory *StorageTypeFactory) visitObjectType(
+	visitor *astmodel.TypeVisitor,
+	object *astmodel.ObjectType,
+	ctx interface{},
```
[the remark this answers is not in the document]
theunrepentantgeek: Reformatted to put the return type on the same line as the final parameter.

hack/generator/pkg/codegen/pipeline_create_storage_types.go
theunrepentantgeek: It's safe because we're not storing patches, but the whole thing. With `omitempty` on the types, there's no difference between `""` and **nil** anyway.

hack/generator/pkg/codegen/pipeline_create_storage_types.go
theunrepentantgeek: Changed return type to `TypeName` instead of `*Typename`

hack/generator/pkg/codegen/pipeline_create_storage_types.go
theunrepentantgeek: It happens if an ArmType gets wrapped by something else. Won't happen via a property as we do different things there - but mostly this is defensive.

hack/generator/pkg/codegen/pipeline_create_storage_types.go
theunrepentantgeek: It creates a clone (note the receiver doesn't use a pointer) and then modifies it; if used for a different object, any property value would be stale - so I defensively set it to ensure nothing goes wrong.

theunrepentantgeek: In _current use_ it's likely already nil, but it doesn't cost to set it and it prevents bugs if our usage pattern changes.

theunrepentantgeek (commented):
Updating code in response to PR feedback

hack/generator/pkg/astmodel/armconversion/shared.go
```diff
@@ -62,17 +62,20 @@ func GetAzureNameProperty(idFactory astmodel.IdentifierFactory) *astmodel.Proper
...
-	receiverType, err := codeGenerationContext.GetImportedDefinition(receiver)
+	rt, err := codeGenerationContext.GetImportedDefinition(receiver)
 	if err != nil {
 		panic(err)
 	}
 
-	kubeType, ok := receiverType.Type().(*astmodel.ObjectType)
-	if !ok {
-		panic(fmt.Sprintf("receiver for ArmConversionFunction is not of type ObjectType. TypeName: %v, Type %T", receiver, receiverType.Type()))
+	switch receiverType := rt.Type().(type) {
```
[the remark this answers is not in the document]
matthchr: I think the entire purpose of the wrapper types as we have them (ArmType, StorageType, One/etc) is to give us a way to identify types while we're performing type translations. Information like that certainly isn't ever going to be required at code generation time (because its purpose is for the pipeline phases)?

I could see the argument that there may be other types that wrap struct but can't be translated back to struct, but those types are then by definition not wrapper types, I think?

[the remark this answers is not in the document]
matthchr: Put another way:

It doesn't seem worth it to me to pay the price (coupling in the code generation code) now, when we don't have any examples where we actually need that functionality. We could go with something that removed that coupling (pipeline stage projecting all the wrapper types away) and always revisit it later if a type came up that didn't work for.

hack/generator/pkg/astmodel/storage_type.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+// AsDeclarations renders as a Go abstract syntax tree for a declaration
+func (st *StorageType) AsDeclarations(codeGenerationContext *CodeGenerationContext, name TypeName, description []string) []ast.Decl {
+	return st.objectType.AsDeclarations(codeGenerationContext, name, description)
+}
+
+// Equals decides if the types are the same
+func (st *StorageType) Equals(t Type) bool {
+	if ost, ok := t.(*StorageType); ok {
```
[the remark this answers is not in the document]
matthchr: I struggle with it too. `ost` as a receiver for example wouldn't have tripped my naming watcher, because I'm beginning to accept them as weird abbreviations...

hack/generator/pkg/astmodel/type_visitor.go
```diff
@@ -67,7 +69,11 @@ func (tv *TypeVisitor) VisitDefinition(td TypeDefinition, ctx interface{}) (*Typ
 
 	name, ok := visitedName.(TypeName)
 	if !ok {
-		return nil, errors.Errorf("expected visit of %q to return TypeName, not %T", td.Name(), visitedName)
+		if t, ok := visitedName.(*TypeName); ok {
+			name = *t
```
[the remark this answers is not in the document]
matthchr: I think it was only coming as `*TypeName` because of the place where you returned a `*TypeName` (Which I believe I highlighted in a different comment?)

I think if you change that you can drop this too.

hack/generator/pkg/codegen/code_generator.go
```diff
@@ -60,6 +60,7 @@ func corePipelineStages(idFactory astmodel.IdentifierFactory, configuration *con
 		applyKubernetesResourceInterface(idFactory),
 		checkForAnyType(configuration.AnyTypePackages),
 		checkForMissingStatusInformation(),
+		createStorageTypes(),
```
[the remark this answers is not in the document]
matthchr: Just before the `checkForX` methods, as those are supposed to validate the shape of objects just before being written. Running something (other than code generation) after them bypasses their checks.

hack/generator/pkg/codegen/code_generator.go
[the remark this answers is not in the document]
matthchr: Yeah, agree - resolving this.

hack/generator/pkg/astmodel/storage_type.go
```diff
@@ -0,0 +1,89 @@
...
+package astmodel
+
+import (
+	"fmt"
+	"go/ast"
+	"strings"
+)
+
+// StorageType wraps an existing type to indicate that it is a storage focussed variation
+type StorageType struct {
```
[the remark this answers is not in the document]
matthchr: Fair enough

hack/generator/pkg/codegen/pipeline_create_storage_types.go
matthchr: I think since you've changed this, you can do away with the bits talked about [here](https://github.com/Azure/k8s-infra/pull/255#discussion_r489813686) as well

hack/generator/pkg/codegen/pipeline_create_storage_types.go
matthchr: > With omitempty on the types, there's no difference between "" and nil anyway.

That is not true, [there is a difference](https://play.golang.org/p/qP6nUKTvdXO). I believe we need to preserve this difference as it can have semantic meaning to APIs.

hack/generator/pkg/codegen/pipeline_create_storage_types.go
matthchr: Rather than be defensive here wouldn't it just be better to error?

It seems to me that "storage types shall not have/consume ARM types" is a fundamental truth, yes?

hack/generator/pkg/astmodel/arm_type_test.go
```diff
@@ -4,75 +4,3 @@
...
-	. "github.com/onsi/gomega"
-
-	"testing"
-)
-
-/*
- * IsArmType() tests
- */
-
-func Test_IsArmType_GivenType_ReturnsExpectedResult(t *testing.T) {
```
matthchr, line 18: Was deleting this code intended?

hack/generator/pkg/astmodel/reference_graph.go
```diff
@@ -28,7 +28,7 @@ func CollectResourceDefinitions(definitions Types) TypeNameSet {
 }
 
 // CollectArmSpecDefinitions returns a TypeNameSet of all of the
-// arm spec definitions passed in.
+// ARM spec definitions passed in.
 func CollectArmSpecAndStatusDefinitions(definitions Types) TypeNameSet {
 	findType := func(t Type) (TypeName, error) {
```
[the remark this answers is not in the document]
matthchr: Yeah the downside of the wrapper is just the boilerplate - wondering if in simple cases we can embed or something and thus avoid most of it.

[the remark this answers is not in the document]
matthchr: Anyway not something you need to fix in this PR so resolving this.

hack/generator/pkg/astmodel/reference_graph.go
```diff
@@ -39,17 +39,21 @@ func CollectArmSpecAndStatusDefinitions(definitions Types) TypeNameSet {
...
 		return armName, nil
 	}
 
-	// TODO: We should be using a better way to identify ARM types. I believe
-	// TODO Bevan is working on it.
 	armSpecAndStatus := make(TypeNameSet)
 	for _, def := range definitions {
-		if resourceType, ok := def.Type().(*ResourceType); ok {
+
+		if IsStoragePackageReference(def.Name().PackageReference) {
```
matthchr, line 51: As mentioned in a different comment I think you could technically remove this and things would still work (because `ResolveResourceType` won't resolve `StorageTypes`?

If you did do that you could also delete this method.

I'm fine with it this way too though if you prefer so feel free to just resolve this if you wanna leave it as is

hack/generator/pkg/codegen/pipeline_create_storage_types.go
matthchr: It feels to me like it not being found deserves either a warning log or an error (I lean towards error, especially if you're not sure if it's happening at all).

If you make this return error and everything still runs, awesome. If not you can investigate the cause and if it's expected then maybe this is a warning or something.

hack/generator/pkg/astmodel/armconversion/shared.go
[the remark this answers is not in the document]
theunrepentantgeek: I'm torn because it seems as though the actual code generation step is already doing some heavy lifting, but I agree that I shouldn't be making things worse. 

I've created PR #261 to simplify types near the end of the pipeline; once that's merged I'll rebase this branch and simplify the method above.

hack/generator/pkg/astmodel/reference_graph.go
theunrepentantgeek: Comments added - let me know whether they're sufficiently clear

hack/generator/pkg/astmodel/type_visitor.go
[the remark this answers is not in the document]
theunrepentantgeek: Yes, it can be removed - I'll do so.

hack/generator/pkg/codegen/code_generator.go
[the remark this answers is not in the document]
theunrepentantgeek: Relocated the creation of storage types before the check stages.

hack/generator/pkg/codegen/pipeline_create_storage_types.go
theunrepentantgeek: We definitely have imported types (aka a `TypeName` containing an `ExternalPackageReference` instead of a `LocalPackageReference`) that are legit, so emitting a warning would just be noise. (They're types imported from the `genruntime` package.)

hack/generator/pkg/astmodel/armconversion/shared.go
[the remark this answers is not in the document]
theunrepentantgeek: Refactored.

hack/generator/pkg/astmodel/storage_type.go
```diff
@@ -0,0 +1,89 @@
...
+	return MakeLocalPackageReference(group, pkg+"s"), nil
+}
+
+func IsStoragePackageReference(ref PackageReference) bool {
+	_, pkg, err := ref.GroupAndPackage()
+	if err != nil {
+		return false
+	}
+
+	return strings.HasSuffix(pkg, "s")
```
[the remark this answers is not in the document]
theunrepentantgeek: No longer detecting storage packages by name.

hack/generator/pkg/codegen/pipeline_create_storage_types.go
theunrepentantgeek: Done.

hack/generator/pkg/codegen/pipeline_create_storage_types.go
theunrepentantgeek: I've removed the special casing for strings.

hack/generator/pkg/codegen/pipeline_create_storage_types.go
theunrepentantgeek: The visitor walks the entire tree; if we don't truncate the walk here, we'd end up creating a storage type based on the object type wrapped by the ArmType - which we don't want/need.

hack/generator/pkg/codegen/pipeline_create_storage_types.go
```diff
@@ -0,0 +1,234 @@
...
+type StorageTypeFactory struct {
+	types               astmodel.Types
+	propertyConversions []propertyConversion
+	visitor             astmodel.TypeVisitor
+}
+
+// A property conversion accepts a property definition and optionally applies a conversion to make
+// the property suitable for use on a storage type. Conversions return nil if they decline to
+// convert, deferring the conversion to another.
+type propertyConversion = func(property *astmodel.PropertyDefinition, cxt StorageTypesVisitorContext) (*astmodel.PropertyDefinition, error)
```
babbageclunk, line 92:
```suggestion
type propertyConversion = func(property *astmodel.PropertyDefinition, ctx StorageTypesVisitorContext) (*astmodel.PropertyDefinition, error)
```

hack/generator/pkg/codegen/pipeline_create_storage_types.go
```diff
@@ -0,0 +1,234 @@
...
+
+	result := astmodel.MakeTypeVisitor()
+	result.VisitTypeName = factory.visitTypeName
+	result.VisitObjectType = factory.visitObjectType
+	result.VisitArmType = factory.visitArmType
+
+	factory.visitor = result
+	factory.propertyConversions = []propertyConversion{
+		factory.kubernetesResourceStoragePropertyConversion,
+		factory.defaultStoragePropertyConversion,
```
babbageclunk, line 77: These would read better as verb phrases:
```suggestion
		factory.convertKubernetesResourceStorageProperty,
		factory.convertDefaultStorageProperty,
```

hack/generator/pkg/codegen/pipeline_create_storage_types.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+
+	// No opinion, defer to another conversion
+	return nil, nil
+}
+
+func (factory *StorageTypeFactory) defaultStoragePropertyConversion(
+	prop *astmodel.PropertyDefinition,
+	ctx StorageTypesVisitorContext) (*astmodel.PropertyDefinition, error) {
+	propertyType, err := factory.visitor.Visit(prop.PropertyType(), ctx.forProperty(prop))
```
babbageclunk, line 184: Haven't you already done the `forProperty` in `makeStorageProperty`? I guess it doesn't matter functionally though.

hack/generator/pkg/codegen/pipeline_create_storage_types.go
```diff
@@ -0,0 +1,234 @@
...
+		}
+	}
+
+	if len(errs) > 0 {
+		err := kerrors.NewAggregate(errs)
+		return nil, err
+	}
+
+	ot := astmodel.NewObjectType().WithProperties(properties...)
+	return astmodel.NewStorageType(*ot), nil
```
babbageclunk, line 144: Is this why you need the default `VisitStorageType` to accept a returned `StorageType`? I don't think you should though, because you're not going to run this visitor on a storage type. Is that right?

hack/generator/pkg/codegen/pipeline_create_storage_types.go
babbageclunk: I like this.

babbageclunk (approved):
This looks really good - a few minor suggestions. I'm not sure why you need to be permissive in the default visit Arm/Storage though.

hack/generator/pkg/astmodel/reference_graph.go
theunrepentantgeek: Assuming the comments are sufficiently clear.

hack/generator/pkg/astmodel/storage_type.go
```diff
@@ -0,0 +1,70 @@
...
+package astmodel
+
+import (
+	"fmt"
+	"go/ast"
+)
+
+// StorageType wraps an existing type to indicate that it is a storage focussed variation
+type StorageType struct {
+	objectType ObjectType
```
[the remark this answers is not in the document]
theunrepentantgeek: Wouldn't embedding `ObjectType` within `StorageType` mean that a cast to `ObjectType` would work? I explicitly don't want it to. I don't want `StorageType` to be convertable to an `ObjectType` except by someone deliberately doing so.

hack/generator/pkg/astmodel/type_visitor.go
```diff
@@ -188,17 +211,21 @@ func IdentityVisitOfResourceType(this *TypeVisitor, it *ResourceType, ctx interf
...
 
-	ot, ok := newType.(*ObjectType)
-	if !ok {
-		return nil, errors.Errorf("expected transformation of ARM underlying type %v to return ObjectType, not %v", at.objectType, ot)
-	}
+	switch newType := nt.(type) {
+	case *ObjectType:
+		return MakeArmType(*newType), nil
+
+	case *ArmType:
```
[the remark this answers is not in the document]
theunrepentantgeek: I added this in because it was needed - because I had the default case (below) return an error. 

More generally, the signature of the methods for `TypeVisitor` allow any transformation, so it's perfectly valid for an `ObjectType` to go in, and an `ArmType` to come out (and in fact we do this in a few different places).

hack/generator/pkg/astmodel/type_visitor.go
```diff
@@ -188,17 +211,21 @@ func IdentityVisitOfResourceType(this *TypeVisitor, it *ResourceType, ctx interf
...
+		return newType, nil
 
-	return MakeArmType(*ot), nil
+	default:
+		return nil, errors.Errorf("expected transformation of ARM underlying type %v to return ObjectType or ArmType, not %v", at.objectType, nt)
```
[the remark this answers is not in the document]
theunrepentantgeek: Agreed - will update the format string to use `%T`

hack/generator/pkg/astmodel/type_visitor.go
```diff
@@ -238,3 +265,21 @@ func IdentityVisitOfAllOfType(this *TypeVisitor, it AllOfType, ctx interface{})
...
+	nt, err := this.Visit(&st.objectType, ctx)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to visit storage type %v", st.objectType)
+	}
+
+	switch newType := nt.(type) {
+	case *ObjectType:
+		return NewStorageType(*newType), nil
+
+	case *StorageType:
```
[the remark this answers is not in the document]
theunrepentantgeek: And pretty much the same answer - I added this type check in because it was needed.

hack/generator/pkg/astmodel/types.go
```diff
@@ -93,3 +93,73 @@ func (types Types) Copy() Types {
...
+			return types.ResolveResourceType(def.theType)
+		}
+		return nil, false
+
+	default:
+		return nil, false
+	}
+}
+
+// IsArmType returns true if the passed type is a Arm type or names an Arm type; false otherwise.
```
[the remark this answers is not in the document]
theunrepentantgeek: Thanks for catching the typo.

hack/generator/pkg/astmodel/types.go
```diff
@@ -93,3 +93,73 @@ func (types Types) Copy() Types {
...
+func (types Types) IsArmDefinition(definition *TypeDefinition) bool {
+	return types.IsArmType(definition.Type())
+}
+
+// IsArmDefinition returns true if the passed resource contains Arm Types or names Arm types; false otherwise.
+func (types Types) IsArmResource(resource *ResourceType) bool {
+	return types.IsArmType(resource.SpecType()) || types.IsArmType(resource.StatusType())
+}
+
+// ResolveEnumType returns true if the passed type is a Arm type or names an Arm type; false otherwise.
```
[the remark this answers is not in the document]
theunrepentantgeek: And again, thanks for catching the typo.

hack/generator/pkg/codegen/pipeline_create_arm_types.go
```diff
@@ -290,15 +260,23 @@ func createArmTypeDefinition(definitions astmodel.Types, def astmodel.TypeDefini
...
-		// This type is the ARM type so give it the ARM name
-		def.WithName(astmodel.CreateArmTypeName(def.Name())),
-		[]conversionHandler{removeValidations, convertPropertiesToArmTypesWrapper})
+	armName := astmodel.CreateArmTypeName(def.Name())
+	armDef, err := def.WithName(armName).
+		ApplyObjectTransformations(removeValidations, convertPropertiesToArmTypesWrapper)
 	if err != nil {
-		return astmodel.TypeDefinition{}, err
+		return astmodel.TypeDefinition{},
+			errors.Wrapf(err, "Failed to create ARM prototype %v from Kubernetes definition %v", armName, def.Name())
```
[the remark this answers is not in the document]
theunrepentantgeek: Good call to improve the error message.

hack/generator/pkg/codegen/pipeline_create_storage_types.go
theunrepentantgeek: Yes, I have already converted it using `forProperty`; this one is redundant, removing.

hack/generator/pkg/codegen/pipeline_create_storage_types.go
```diff
@@ -0,0 +1,234 @@
...
+	storageRef := astmodel.MakeStoragePackageReference(localRef)
+	visitedName := astmodel.MakeTypeName(storageRef, name.Name())
+	return visitedName, nil
+}
+
+func (factory *StorageTypeFactory) visitObjectType(
+	_ *astmodel.TypeVisitor,
+	object *astmodel.ObjectType,
+	ctx interface{}) (astmodel.Type, error) {
+	vc := ctx.(StorageTypesVisitorContext)
```
matthchr, line 124: minor: I know we like abbreviations in go, but I think `visitorCtx` and `objectCtx` (rather than vc/oc) might help with readability.

hack/generator/pkg/astmodel/storage_type.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+// IsStorageDefinition returns true if the passed definition is for a storage type; false otherwise.
+func IsStorageDefinition(definition TypeDefinition) bool {
+	return IsStorageType(definition.Type())
+}
+
+// RequiredImports returns a list of packages required by this type
+func (st *StorageType) RequiredImports() []PackageReference {
+	return st.objectType.RequiredImports()
```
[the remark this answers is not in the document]
matthchr: Ah it looks like they aren't because you're using a revised `ObjectType` as the inner for the `StorageType`.

hack/generator/pkg/codegen/pipeline_create_storage_types.go
matthchr: (Applies across all of the visit methods in this file)

hack/generator/pkg/codegen/pipeline_create_storage_types.go
```diff
@@ -0,0 +1,234 @@
...
+		if p != nil {
+			// We have the conversion we need, return it promptly
+			return p, nil
+		}
+	}
+
+	return nil, fmt.Errorf("failed to find a conversion for property %v", prop.PropertyName())
+}
+
+// Preserve properties required by the KubernetesResource interface unchanged as they're always required
```
matthchr, line 167:
```suggestion
// preserveKubernetesResourceStorageProperties preserves properties required by the KubernetesResource interface unchanged as they're always required
```

hack/generator/pkg/codegen/pipeline_create_storage_types.go
```diff
@@ -0,0 +1,234 @@
...
+func (factory *StorageTypeFactory) preserveKubernetesResourceStorageProperties(
```
matthchr, line 168: One minor note for later (I've noted it in my notes), it's possible we can make it so that _only_ the storage types implement the `KubernetesResource` interface, since I _think_ that those are the types that the controller is going to operate on?

matthchr: Hmm, but it will have to convert to the actual APIVersion type in question before serializing to/from Azure... I'll think about this some more.

hack/generator/pkg/codegen/pipeline_create_storage_types.go
```diff
@@ -0,0 +1,234 @@
...
+	prop *astmodel.PropertyDefinition,
+	_ StorageTypesVisitorContext) (*astmodel.PropertyDefinition, error) {
+	if prop.HasName(astmodel.AzureNameProperty) ||
```
matthchr, line 171: minor: Wondering if we should make this an `IsKubernetesResourceProperty` in the `kubernetes_resource_interface.go` file just in case we add more properties to that interface then they won't be missed?

hack/generator/pkg/codegen/pipeline_create_storage_types.go
```diff
@@ -0,0 +1,234 @@
...
+		return prop, nil
+	}
+
+	// No opinion, defer to another conversion
+	return nil, nil
+}
+
+func (factory *StorageTypeFactory) convertPropertiesForStorage(
+	prop *astmodel.PropertyDefinition,
+	ctx StorageTypesVisitorContext) (*astmodel.PropertyDefinition, error) {
```
matthchr, line 183: This seems to not be used at all? Can it just be `_` then, or possibly removed entirely from the `propertyConversion` function signature since neither definition use it? (not sure removing it from func signature makes sense as maybe others might need it?)

matthchr: Oh, or is the parameter `prop` to the `Visit` call below supposed to be `ctx`?

matthchr: But if its supposed to be `ctx` and works as `prop` that seems to suggest it's not actually needed? I'm not actually sure how it works with prop as there are a bunch of casts in the visit handlers and I'd expect one of them to panic...

hack/generator/pkg/codegen/pipeline_create_storage_types.go
```diff
@@ -0,0 +1,234 @@
...
+
+	result := []string{
+		fmt.Sprintf("Storage version of %v.%v", pkg, definition.Name().Name()),
+	}
+	result = append(result, definition.Description()...)
+
+	return result
+}
+
+type StorageTypesVisitorContext struct {
```
matthchr, line 216: Might be worth leaving a comment on this context type? It's not clear to me what exactly it's actually for. I tried to follow its usage to figure out what it is doing but as far as I can tell:
1. There is a single call to `forProperty`, which doesn't actually do anything because neither of the conversion methods actually use the provided context.
2. There is a single call to `forObject` which is passed to `makeStorageProperty` which then has `forProp` called on it and is passed to the conversion method, but see item 1 above as that value isn't used, which suggests the result of `forObject` _also_ actually isn't used?
3. There's a use of `vc.property != nil` but based on 1 and 2 above I can't see how that's getting hit ever.

I must be missing something but not sure what, and definitely not sure I follow exactly what this is tracking and what its purpose is.

matthchr (approved):
Looks good to me except for my confusion around `StorageTypesVisitorContext` -- signing off but I think you should look at those comments before merge as either I'm blind (quite possible) or something is off.

hack/generator/pkg/astmodel/storage_type.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+// References returns the names of all types that this type references
+func (st *StorageType) References() TypeNameSet {
+	return st.objectType.References()
+}
+
+// AsType renders as a Go abstract syntax tree for a type
+func (st *StorageType) AsType(codeGenerationContext *CodeGenerationContext) ast.Expr {
+	return st.objectType.AsType(codeGenerationContext)
```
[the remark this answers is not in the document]
matthchr: Ok I think I'm crazy because at least this type and possibly ArmType are making it all the way to the end... so you should probably disregard this.

hack/generator/pkg/astmodel/arm_type.go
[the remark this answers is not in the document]
theunrepentantgeek: Renamed to `NewArmType()`.

hack/generator/pkg/codegen/pipeline_create_storage_types.go
theunrepentantgeek: I've renamed this to `visitTypeName()`.

hack/generator/pkg/astmodel/type_definition.go
```diff
@@ -100,3 +101,52 @@ func (def TypeDefinition) RequiredImports() []PackageReference {
...
+// definition, returning a new definition.
+// If the definition does not contain an object, an error will be returned
+// The transformations are constrained to return ObjectType results to allow them to be chained together.
+func (def TypeDefinition) ApplyObjectTransformations(transforms ...func(*ObjectType) (*ObjectType, error)) (*TypeDefinition, error) {
+	return def.ApplyObjectTransformation(func(objectType *ObjectType) (Type, error) {
+		result := objectType
+		for _, transform := range transforms {
+			rt, err := transform(result)
+			if err != nil {
+				return nil, err
```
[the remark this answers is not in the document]
theunrepentantgeek: Good idea; have wrapped the error as suggested.

hack/generator/pkg/astmodel/type_visitor.go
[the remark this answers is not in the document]
theunrepentantgeek: Note that we have transformations that take in an ObjectType and return an ArmType (part of the pipeline to create the ArmTypes); the same is happening with StorageTypes as well.

hack/generator/pkg/codegen/pipeline_create_storage_types.go
theunrepentantgeek: I used different names, but did make them into verb phrases.

hack/generator/pkg/astmodel/package_import.go
```diff
@@ -53,6 +53,11 @@ func (pi PackageImport) PackageName() string {
 	return pi.PackageReference.PackageName()
 }
 
+// PackagePath is the full path of the package reference
+func (pi PackageImport) PackagePath() string {
+	return pi.PackageReference.PackagePath()
```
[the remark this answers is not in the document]
theunrepentantgeek: I think this makes sense, but will do it separately. #291

hack/generator/pkg/astmodel/property_definition_test.go
```diff
@@ -226,6 +226,54 @@ func Test_PropertyDefinitionWithType_GivenSameType_ReturnsExistingReference(t *t
 	g.Expect(updated).To(BeIdenticalTo(original))
 }
 
+/*
+ * WithValidation() Tests
+ */
+
+func Test_PropertyDefinitionWithValidation_GivenValidation_AddsToValidation(t *testing.T) {
```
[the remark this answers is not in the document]
theunrepentantgeek: Yeah, they're similar but testing different entry points. I think they both have value.

hack/generator/pkg/codegen/pipeline_create_storage_types.go
theunrepentantgeek: I'll go through and give them better names.

hack/generator/pkg/codegen/pipeline_create_storage_types.go
theunrepentantgeek: Reworded comment.

hack/generator/pkg/codegen/pipeline_create_storage_types.go
theunrepentantgeek: Agreed, this is a good idea. Will do that separately as it's not directly related to this PR. See #293

hack/generator/pkg/codegen/pipeline_create_storage_types.go
theunrepentantgeek: Huh. This is one of those "I wonder how this ever worked" moments. In retesting, I find it does panic - should be passing `ctx` as the value to visit, not `prop`.

hack/generator/pkg/codegen/pipeline_create_storage_types.go
theunrepentantgeek: When visiting a `TypeName` different actions need to be taken, depending on what is being named. 

Currently, we only have one rule: An `EnumType` needs to be converted to the underlying base type, but only if the original `TypeName` is part of a property definition.

I'm expecting additional rules will be added as I work through milestones for storage types.

PR #258, opened 2020-09-15
Rename comment helper methods for clarity

theunrepentantgeek:
Changes as follows:

* Rename `addDocComments()` to `addWrappedComments()` (plural form)
* Rename `addDocComment()` to `addWrappedComment()` (singular form)
* Rename `formatDocComment()` to `formatComment()` 
* Introduces `addComment` (no wordwrap)

### DIFF
--- a/hack/generator/pkg/astmodel/comments.go
+++ b/hack/generator/pkg/astmodel/comments.go
@@ -26,19 +26,7 @@ func addWrappedComments(commentList *[]*ast.Comment, comments []string, width in
 
 func addWrappedComment(commentList *[]*ast.Comment, comment string, width int) {
 	for _, c := range formatComment(comment, width) {
-		line := strings.TrimSpace(c)
-
-		if !strings.HasPrefix(line, "//") {
-			line = "//" + line
-		}
-
-		if *commentList == nil {
-			line = "\n" + line
-		}
-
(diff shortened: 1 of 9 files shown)


matthchr (approved):
LGTM

hack/generator/pkg/astmodel/comments.go
```diff
@@ -42,8 +42,24 @@ func addDocComment(commentList *[]*ast.Comment, comment string, width int) {
 	}
 }
 
-// formatDocComment splits the supplied comment string up ready for use as a documentation comment
-func formatDocComment(comment string, width int) []string {
+func addComment(commentList *[]*ast.Comment, comment string) {
+	line := strings.TrimSpace(comment)
```
babbageclunk, line 46: Would it make sense for addWrappedComment to call addComment, rather than duplicating the body of it in the loop?

babbageclunk (approved):
Nice - I had a coworker who liked to make the point that functions and types were often initially misnamed as what they were used for rather than what they actually did.

hack/generator/pkg/astmodel/comments.go
theunrepentantgeek: Good idea; I'll reduce duplication by adding reuse.

PR #260, opened 2020-09-21
Convert PackageReference to an interface with multiple implementations

theunrepentantgeek:
Splits the existing `PackageReference` struct into two distinct implementations `LibraryPackageReference` and `LocalPackageReference`, simplifying the implementations and laying round work for future changes.

Prerequisite for #255


babbageclunk (approved):
Nice!

hack/generator/pkg/astmodel/package_reference.go
```diff
@@ -5,88 +5,30 @@
...
-	return PackageReference{packagePath: url}
-}
+var MetaV1PackageReference = MakeLibraryPackageReference("k8s.io/apimachinery/pkg/apis/meta/v1")
 
-// MakePackageReference creates a new package reference from a path
-func MakePackageReference(packagePath string) PackageReference {
-	return PackageReference{packagePath: packagePath}
+type PackageReference interface {
+	// IsLocalPackage returns a valud indicating whether this is a local package
+	IsLocalPackage() bool
```
matthchr, line 18: This doesn't feel quite right -- the interface here has knowledge that a specific implementer is going to be a `LocalPackage`, and that `Group` and `Version` are a thing?

Are there really two interfaces here?
`Package` and `LocalPackage` (or `GroupVersionPackage` or something) where `LocalPackage`/`GroupVersionPackage` has the `Group` and `Version` methods, and also embeds `Package` (which is the basic one)?

Then `IsLocalPackage` becomes `AsLocalPackage` that does a check + cast and returns a `LocalPackage`, where you can call the methods without fear of error (because you know you have the right type!)

matthchr: If you add `StoragePackage` that's fine it fits well into that model too as it'll just impl `LocalPackage` rather than `Package`

hack/generator/pkg/config/type_transformer.go
```diff
@@ -16,7 +16,9 @@ import (
 
 // A transformation target
 type TransformTarget struct {
-	PackagePath string `yaml:",omitempty"`
+	LibraryPath string `yaml:",omitempty"`
+	Group       string `yaml:",omitempty"`
+	PackageName string `yaml:"package,omitempty"`
```
matthchr, line 21: Same comment about `Version` vs `Package` here

hack/generator/pkg/config/type_transformer.go
```diff
@@ -16,7 +16,9 @@ import (
 
 // A transformation target
 type TransformTarget struct {
-	PackagePath string `yaml:",omitempty"`
+	LibraryPath string `yaml:",omitempty"`
```
matthchr, line 19: I don't think having this here makes sense because LibraryPath (as per your language above) was for dependencies that aren't GVK's, and the whole point of `TransformTarget` is that it's transforming types we're generating, which by definition have a GVK (and thus aren't a member of a "library")?

I think you just want Group+Version here?

hack/generator/pkg/astmodel/library_package_reference.go
```diff
@@ -0,0 +1,60 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ */
+
+package astmodel
+
+import (
```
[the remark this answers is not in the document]
theunrepentantgeek: Doesn't seem large enough (to me!) to justify having a new package just for these.

hack/generator/pkg/astmodel/library_package_reference.go
```diff
@@ -0,0 +1,60 @@
...
+	"errors"
+	"fmt"
+	"strings"
+)
+
+// LibraryPackageReference indicates a library package that needs to be imported
```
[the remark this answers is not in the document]
theunrepentantgeek: I like `ExternalPackageReference`, will rename it to that.

hack/generator/pkg/astmodel/local_package_reference.go
```diff
@@ -0,0 +1,62 @@
...
+import "fmt"
+
+const (
+	LocalPathPrefix = "github.com/Azure/k8s-infra/hack/generated/apis/" // TODO: From config?
+)
+
+// LocalPackageReference specifies a local package name or reference
+type LocalPackageReference struct {
+	groupName   string
+	packageName string
```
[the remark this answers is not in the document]
theunrepentantgeek: This naming probably dates back to when our AstModel was very firmly Go focused - renaming this to `version` makes sense to me, given our context.

hack/generator/pkg/astmodel/library_package_reference.go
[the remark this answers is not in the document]
matthchr: Yeah I wasn't sure.

hack/generator/pkg/astmodel/package_reference.go
theunrepentantgeek: I don't think there are different interfaces as both implementations are a `PackageReference` - just to packages that are found in different places.

Changing to `AsLocalPackage()` - or having consumers do the cast themselves - locks in the hierarchy and makes it difficult for other kinds of package reference to be added. 

Essentially, consumers shouldn't need to know about the various implementations; they should interact with solely the interface.

hack/generator/pkg/config/type_transformer.go
theunrepentantgeek: Agreed.

hack/generator/pkg/config/type_transformer.go
theunrepentantgeek: After checking with @babbageclunk, agreed.

hack/generator/pkg/astmodel/package_reference.go
matthchr: But if you had written the interface without knowing what the implementers were going to be, you'd never have included a `Group` and a `Package` function (as that doesn't make sense in the context of a `Reference`). That suggests to me that we've leaked implementation details into our interface.

I'm suggesting a hierarchy where Package and LocalPackage are interfaces. LocalPackage is a Package (with some extra methods), so they would still both be `PackageReferences` and you could use anything which impls `LocalPackage` as a `Package` (since `LocalPackage` is a superset).

When you need those extra methods, yes you have to cast if you've only got `LocalPackage` - but that's what the current impls of `Group` and `Package` functions are doing anyway effectively (return an error for some impls, not for others).

matthchr: For example if I wanted to add a `GithubPackage` which was a package specifically from GitHub, and it had its own `GithubStarCount()` function, we wouldn't put that function on the interface and have all the other impls go implement that method and return an error from it would we?

PR #261, opened 2020-09-22
Add pipeline stage to simplify types

theunrepentantgeek:
We're introducing a number of wrapper types to distinguish different types as we process them through the pipeline; this stage flattens all of those types just prior to generating the actual code.

Prerequisite for #255

### DIFF
--- a/hack/generator/pkg/codegen/code_generator.go
+++ b/hack/generator/pkg/codegen/code_generator.go
@@ -60,6 +60,7 @@ func corePipelineStages(idFactory astmodel.IdentifierFactory, configuration *con
 		applyKubernetesResourceInterface(idFactory),
 		checkForAnyType(configuration.AnyTypePackages),
 		checkForMissingStatusInformation(),
+		simplifyDefinitions(),
 	}
 }
 
--- a/hack/generator/pkg/codegen/code_generator.go
+++ b/hack/generator/pkg/codegen/code_generator.go
@@ -58,9 +58,9 @@ func corePipelineStages(idFactory astmodel.IdentifierFactory, configuration *con
 		stripUnreferencedTypeDefinitions(),
 		createArmTypesAndCleanKubernetesTypes(idFactory),
 		applyKubernetesResourceInterface(idFactory),
+		simplifyDefinitions(),
 		checkForAnyType(configuration.AnyTypePackages),
 		checkForMissingStatusInformation(),
-		simplifyDefinitions(),
 	}
 }
 
--- a/hack/generator/pkg/codegen/pipeline_simplify_definitions.go
+++ b/hack/generator/pkg/codegen/pipeline_simplify_definitions.go
@@ -16,7 +16,7 @@ import (
 func simplifyDefinitions() PipelineStage {
 	return MakePipelineStage(
 		"simplifyDefinitions",
-		"Simplify definitions",
+		"Flatten definitions by removing wrapper types",
 		func(ctx context.Context, defs astmodel.Types) (astmodel.Types, error) {
 			visitor := createSimplifyingVisitor()
 			var errs []error
@@ -26,9 +26,9 @@ func simplifyDefinitions() PipelineStage {
 				if err != nil {
 					errs = append(errs, err)
 				} else {
-					result[d.Name()] = *d
+					result.Add(*d)
 					if !def.Type().Equals(d.Type()) {
-						klog.V(3).Infof("Simplified %v", def.Name())
+						klog.V(3).Infof("Simplified %v from %v to %v", def.Name(), def.Type(), d.Type())
 					}
 				}
 			}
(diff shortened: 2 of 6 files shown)


hack/generator/pkg/codegen/code_generator.go
```diff
@@ -60,6 +60,7 @@ func corePipelineStages(idFactory astmodel.IdentifierFactory, configuration *con
 		applyKubernetesResourceInterface(idFactory),
 		checkForAnyType(configuration.AnyTypePackages),
 		checkForMissingStatusInformation(),
+		simplifyDefinitions(),
```
matthchr, line 63: Should also be before the `checkForX` pipeline stages I think

hack/generator/pkg/codegen/pipeline_simplify_definitions.go
```diff
@@ -0,0 +1,53 @@
...
+	"github.com/Azure/k8s-infra/hack/generator/pkg/astmodel"
+	kerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/klog"
+)
+
+// simplifyDefinitions creates a pipeline stage that removes any wrapper types prior to actual code generation
+func simplifyDefinitions() PipelineStage {
+	return MakePipelineStage(
+		"simplifyDefinitions",
+		"Simplify definitions",
```
matthchr, line 19: Maybe make this textual description a bit closer to the comment above (about removing wrapper types and whatnot)?

hack/generator/pkg/codegen/pipeline_simplify_definitions.go
```diff
@@ -0,0 +1,53 @@
...
+		func(ctx context.Context, defs astmodel.Types) (astmodel.Types, error) {
+			visitor := createSimplifyingVisitor()
+			var errs []error
+			result := make(astmodel.Types)
+			for _, def := range defs {
+				d, err := visitor.VisitDefinition(def, nil)
+				if err != nil {
+					errs = append(errs, err)
+				} else {
+					result[d.Name()] = *d
```
matthchr, line 29: prefer `result.Add`, as that has nice overwrite checks?

hack/generator/pkg/codegen/pipeline_simplify_definitions.go
```diff
@@ -0,0 +1,53 @@
...
+					if !def.Type().Equals(d.Type()) {
+						klog.V(3).Infof("Simplified %v", def.Name())
```
matthchr, line 31: Worth logging from what to what?

Might be worth checking with @porges as I think he may be doing something similar to this in his OneOf/AllOf PR too, since I think those are also projected away before codegen?

matthchr (approved):
LGTM: May be worth syncing with @porges for if you think this should be combined with the pipeline stage he is adding in his PR that does a similar thing for oneOf/allOf
https://github.com/Azure/k8s-infra/pull/237/files#diff-7282d783a1354669d07cf64357f3646e

hack/generator/pkg/codegen/pipeline_simplify_definitions.go
```diff
@@ -0,0 +1,59 @@
...
+	result := astmodel.MakeTypeVisitor()
+
+	// Unwrap ArmTypes, promoting the object within
+	result.VisitArmType = func(tv *astmodel.TypeVisitor, at *astmodel.ArmType, ctx interface{}) (astmodel.Type, error) {
+		ot := at.ObjectType()
+		return tv.Visit(&ot, ctx)
+	}
+
+	// Don't need to waste type iterating within complex objects
+	result.VisitObjectType = func(_ *astmodel.TypeVisitor, ot *astmodel.ObjectType, _ interface{}) (astmodel.Type, error) {
```
babbageclunk, line 54: Is this because we only wrap ArmTypes around top-level types?

hack/generator/pkg/codegen/pipeline_simplify_definitions.go
```diff
@@ -0,0 +1,59 @@
...
+func createSimplifyingVisitor() astmodel.TypeVisitor {
+	result := astmodel.MakeTypeVisitor()
+
+	// Unwrap ArmTypes, promoting the object within
+	result.VisitArmType = func(tv *astmodel.TypeVisitor, at *astmodel.ArmType, ctx interface{}) (astmodel.Type, error) {
+		ot := at.ObjectType()
+		return tv.Visit(&ot, ctx)
+	}
+
+	// Don't need to waste type iterating within complex objects
```
babbageclunk, line 53:
```suggestion
	// Don't need to waste time iterating within complex objects
```

babbageclunk (approved):
LGTM too.

hack/generator/pkg/codegen/pipeline_simplify_definitions.go
theunrepentantgeek: An ObjectType` doesn't contain any other types, but it does contain a complex structure that we don't need to waste time on.

hack/generator/pkg/codegen/pipeline_simplify_definitions.go
theunrepentantgeek: Changed to use `result.Add()`.

PR #262, opened 2020-09-24
Split AnyType check into two pipeline stages

theunrepentantgeek:
The existing pipeline stage that catches use of `AnyType` is split into two.

The stage `filterOutDefinitionsUsingAnyType` is used earlier in the pipeline to do appropriate filtering, reducing the amount of work done by later stages.

The stage `ensureDefinitionsDoNotUseAnyTypes` is used later in the pipeline to catch any issues introduced by stages than ran after the filter stage.

Prerequisite of PR #255


hack/generator/pkg/codegen/pipeline_check_for_anytype.go
```diff
@@ -16,22 +16,33 @@ import (
...
+}
+
+// ensureDefinitionsDoNotUseAnyTypes returns a stage that will check for any
+// definitions containing AnyTypes. The stage will return errors for each type
+// found that uses an AnyType.
+func ensureDefinitionsDoNotUseAnyTypes() PipelineStage {
+	return checkForAnyTypeCore("Catch rogue definitions using AnyTypes", []string{})
+}
+
+func checkForAnyTypeCore(description string, packages []string) PipelineStage {
```
babbageclunk, line 37: Maybe just `checkForAnyType`? Then it's still a verb phrase.

babbageclunk (approved):
Nice - glad it dropped out simply!

hack/generator/pkg/codegen/pipeline_check_for_anytype.go
theunrepentantgeek: I renamed it early on when intending to retain `checkForAnyType()` as the name of the method that created the stage; now that I'm using `filter`/`ensure` for those methods, changing this back makes sense.

PR #270, opened 2020-10-01
Functions have names

theunrepentantgeek:
Tidy up our object model by ensuring that Functions know their own names, consistent with the way properties and interface implementations also know their own names.

Interface `Function` now includes `Name()`, allowing simplification of calls to `ObjectType.WithFunction()`.

Also simplified `NewInterfaceImplementation()` to accept a variadic slice of functions instead of a map.

We now directly sort functions where needed, instead of sorting their names and then looking them up from a map later on.

Added sorting to `ObjectType.generateMethodDecls()` to ensure output order is consistent; I suspect we were only getting away with this because most methods were added via interface implementations.

### DIFF
--- a/hack/generator/pkg/astmodel/function.go
+++ b/hack/generator/pkg/astmodel/function.go
@@ -11,6 +11,9 @@ import (
 
 // Function represents something that is an (unnamed) Go function
 type Function interface {
+	// The unique name of this interface
+	Name() string
+
 	RequiredImports() []PackageReference
 
 	// References returns the set of types to which this function refers.
@@ -18,7 +21,7 @@ type Function interface {
 	References() TypeNameSet
 
 	// AsFunc renders the current instance as a Go abstract syntax tree
-	AsFunc(codeGenerationContext *CodeGenerationContext, receiver TypeName, methodName string) *ast.FuncDecl
+	AsFunc(codeGenerationContext *CodeGenerationContext, receiver TypeName) *ast.FuncDecl
 
 	// Equals determines if this Function is equal to another one
 	Equals(f Function) bool
--- a/hack/generator/pkg/astmodel/function.go
+++ b/hack/generator/pkg/astmodel/function.go
@@ -11,7 +11,8 @@ import (
 
 // Function represents something that is an (unnamed) Go function
 type Function interface {
-	// The unique name of this interface
+	// The unique name of this function
+	// (You can't have two functions with the same name on the same object or resource)
 	Name() string
 
 	RequiredImports() []PackageReference
--- a/hack/generator/pkg/astmodel/interface_implementation.go
+++ b/hack/generator/pkg/astmodel/interface_implementation.go
@@ -12,8 +12,13 @@ type InterfaceImplementation struct {
 }
 
 // NewInterfaceImplementation creates a new interface implementation with the given name and set of functions
-func NewInterfaceImplementation(name TypeName, functions map[string]Function) *InterfaceImplementation {
-	return &InterfaceImplementation{name: name, functions: functions}
+func NewInterfaceImplementation(name TypeName, functions ...Function) *InterfaceImplementation {
+	result := &InterfaceImplementation{name: name, functions: make(map[string]Function)}
+	for _, f := range functions {
+		result.functions[f.Name()] = f
+	}
+
+	return result
 }
 
 // Name returns the name of the interface
--- a/hack/generator/pkg/astmodel/interface_implementer.go
+++ b/hack/generator/pkg/astmodel/interface_implementer.go
@@ -49,19 +49,19 @@ func (i InterfaceImplementer) AsDeclarations(
 	// interfaces must be ordered by name for deterministic output
 	// (We sort them directly to skip future lookups)
 	var interfaces []*InterfaceImplementation
-	for _, intf := range i.interfaces {
-		interfaces = append(interfaces, intf)
+	for _, iface := range i.interfaces {
+		interfaces = append(interfaces, iface)
 	}
 
 	sort.Slice(interfaces, func(i int, j int) bool {
-		return interfaces[i].Name().name < interfaces[j].Name().Name()
+		return interfaces[i].Name().name < interfaces[j].Name().name
 	})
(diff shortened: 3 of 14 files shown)


hack/generator/pkg/astmodel/interface_implementer.go
```diff
@@ -46,31 +46,31 @@ func (i InterfaceImplementer) AsDeclarations(
...
+	// (We sort them directly to skip future lookups)
+	var interfaces []*InterfaceImplementation
+	for _, intf := range i.interfaces {
+		interfaces = append(interfaces, intf)
 	}
 
-	sort.Slice(ifaceNames, func(i int, j int) bool {
-		return ifaceNames[i].name < ifaceNames[j].name
+	sort.Slice(interfaces, func(i int, j int) bool {
+		return interfaces[i].Name().name < interfaces[j].Name().Name()
```
babbageclunk, line 57: Probably best to be consistent on either side of the `<`.

hack/generator/pkg/astmodel/kubernetes_resource_interface.go
```diff
@@ -66,6 +70,11 @@ type objectFunction struct {
 
 var _ Function = &objectFunction{}
 
+// Name returns the unique name of this function
```
babbageclunk, line 73: Is it unique?

babbageclunk (approved):
Looks great!

hack/generator/pkg/astmodel/function.go
```diff
@@ -11,14 +11,17 @@ import (
 
 // Function represents something that is an (unnamed) Go function
 type Function interface {
+	// The unique name of this interface
```
matthchr, line 14:
```suggestion
	// The name of this function
```

matthchr: Could keep "unique" too although it's sorta wonky because the next question is "unique in what context" and that depends on where the function is used.

hack/generator/pkg/astmodel/interface_implementer.go
```diff
@@ -46,31 +46,31 @@ func (i InterfaceImplementer) AsDeclarations(
...
 	})
 
-	for _, ifaceName := range ifaceNames {
-		iface := i.interfaces[ifaceName]
+	for _, intf := range interfaces {
```
matthchr, line 60: minor: Is `intf` a more clear abbreviation for interface than `iface`? I did write it the other way so am clearly biased but I think `iface` is better =)

hack/generator/pkg/astmodel/kubernetes_resource_interface.go
matthchr: Agree with @babbageclunk (sorta same comment I made above)

hack/generator/pkg/astmodel/object_type.go
```diff
@@ -56,8 +56,19 @@ func (objectType *ObjectType) AsDeclarations(codeGenerationContext *CodeGenerati
...
+	var functions []Function
+	for _, f := range objectType.functions {
+		functions = append(functions, f)
+	}
+
+	sort.Slice(functions, func(i int, j int) bool {
+		return functions[i].Name() < functions[j].Name()
+	})
+
+	for _, f := range objectType.functions {
```
matthchr, line 70: Shouldn't you be iterating over `functions` here, since that's the collection which was sorted?

matthchr: Although interestingly we weren't sorting these before and it was ok? They must have been being sorted elsewhere?

matthchr (approved):

hack/generator/pkg/astmodel/interface_implementer.go
theunrepentantgeek: Fixed to directly access the member `name` on both sides of the ordering comparison.

hack/generator/pkg/astmodel/kubernetes_resource_interface.go
theunrepentantgeek: I believe comments should add information that's not discernable by reading the code, so I've extended the comment to explain how it is unique.

hack/generator/pkg/astmodel/function.go
theunrepentantgeek: Fixed to refer to `function` instead of `interface`. Thanks for catching the typo.
Also extended the comment to explain how the name needs to be unique.

hack/generator/pkg/astmodel/interface_implementer.go
theunrepentantgeek: 😁 My understanding is abbreviated terms where later letters are omitted are easier for readers to decipher, which is why I prefer `intf` over `iface` (don't recall ever seeing `iface` before). 

But, no biggie, I'll change it back.

hack/generator/pkg/astmodel/object_type.go
theunrepentantgeek: Good catch, fixed to use the local `functions`.

The `AsFunc()` call creates an AST fragment, so there's no opportunity for sorting elsewhere.

This list contains directly implemented functions (any functions included to implement an interface are elsewhere), so I think we got lucky that there were only ever zero or one present.

hack/generator/pkg/astmodel/interface_implementation.go
```diff
@@ -12,8 +12,13 @@ type InterfaceImplementation struct {
 }
 
 // NewInterfaceImplementation creates a new interface implementation with the given name and set of functions
-func NewInterfaceImplementation(name TypeName, functions map[string]Function) *InterfaceImplementation {
-	return &InterfaceImplementation{name: name, functions: functions}
+func NewInterfaceImplementation(name TypeName, functions ...Function) *InterfaceImplementation {
+	result := &InterfaceImplementation{name: name, functions: make(map[string]Function)}
+	for _, f := range functions {
+		result.functions[f.Name()] = f
```
Porges, line 18: check for duplicate names here?

Porges (approved):

PR #271, opened 2020-10-05
Introduce PackageImportSet type 

theunrepentantgeek:
I was writing some code for handling test case generation and it felt like someone else must have already written the same code. I was right - multiple times. Fortunately, it all looked consistent

This PR introduces a proper `PackageImportSet` abstraction (with tests), making a bunch of code more declarative and easy to read.

### DIFF
--- a/hack/generator/pkg/astmodel/allof_type.go
+++ b/hack/generator/pkg/astmodel/allof_type.go
@@ -116,8 +116,9 @@ func (allOf AllOfType) AsDeclarations(codeGenerationContext *CodeGenerationConte
 	panic("should have been replaced by generation time by 'convertAllOfAndOneOf' phase")
 }
 
-// RequiredImports returns the union of the required imports of all the AllOf types
-func (allOf AllOfType) RequiredImports() *PackageImportSet {
+// RequiredPackageReferences always panics; AllOf cannot be represented by the Go AST and must be
+// lowered to an object type
+func (allOf AllOfType) RequiredPackageReferences() []PackageReference {
 	panic("should have been replaced by generation time by 'convertAllOfAndOneOf' phase")
 }
 
--- a/hack/generator/pkg/astmodel/allof_type.go
+++ b/hack/generator/pkg/astmodel/allof_type.go
@@ -106,13 +106,13 @@ func (allOf AllOfType) References() TypeNameSet {
 
 // AsType always panics; AllOf cannot be represented by the Go AST and must be
 // lowered to an object type
-func (allOf AllOfType) AsType(codeGenerationContext *CodeGenerationContext) ast.Expr {
+func (allOf AllOfType) AsType(_ *CodeGenerationContext) ast.Expr {
 	panic("should have been replaced by generation time by 'convertAllOfAndOneOf' phase")
 }
 
 // AsDeclarations always panics; AllOf cannot be represented by the Go AST and must be
 // lowered to an object type
-func (allOf AllOfType) AsDeclarations(codeGenerationContext *CodeGenerationContext, name TypeName, description []string) []ast.Decl {
+func (allOf AllOfType) AsDeclarations(_ *CodeGenerationContext, _ TypeName, _ []string) []ast.Decl {
 	panic("should have been replaced by generation time by 'convertAllOfAndOneOf' phase")
 }
 
--- a/hack/generator/pkg/astmodel/allof_type.go
+++ b/hack/generator/pkg/astmodel/allof_type.go
@@ -117,7 +117,7 @@ func (allOf AllOfType) AsDeclarations(codeGenerationContext *CodeGenerationConte
 }
 
 // RequiredImports returns the union of the required imports of all the AllOf types
-func (allOf AllOfType) RequiredImports() []PackageReference {
+func (allOf AllOfType) RequiredImports() *PackageImportSet {
 	panic("should have been replaced by generation time by 'convertAllOfAndOneOf' phase")
 }
 
--- a/hack/generator/pkg/astmodel/allof_type_test.go
+++ b/hack/generator/pkg/astmodel/allof_type_test.go
@@ -106,6 +106,6 @@ func TestAllOfRequiredImportsPanics(t *testing.T) {
 
 	x := AllOfType{}
 	g.Expect(func() {
-		x.RequiredImports()
+		x.RequiredPackageReferences()
 	}).To(PanicWith("should have been replaced by generation time by 'convertAllOfAndOneOf' phase"))
 }
--- a/hack/generator/pkg/astmodel/arm_type.go
+++ b/hack/generator/pkg/astmodel/arm_type.go
@@ -25,9 +25,9 @@ func MakeArmType(object ObjectType) ArmType {
 	}
 }
 
-// RequiredImports returns a list of packages required by this type
-func (at ArmType) RequiredImports() *PackageImportSet {
-	return at.objectType.RequiredImports()
+// RequiredPackageReferences returns a list of packages required by this type
+func (at ArmType) RequiredPackageReferences() []PackageReference {
+	return at.objectType.RequiredPackageReferences()
 }
 
 // References returns the names of all types that this type references
--- a/hack/generator/pkg/astmodel/arm_type.go
+++ b/hack/generator/pkg/astmodel/arm_type.go
@@ -26,7 +26,7 @@ func MakeArmType(object ObjectType) ArmType {
 }
 
 // RequiredImports returns a list of packages required by this type
-func (at ArmType) RequiredImports() []PackageReference {
+func (at ArmType) RequiredImports() *PackageImportSet {
 	return at.objectType.RequiredImports()
 }
 
--- a/hack/generator/pkg/astmodel/code_generation_context.go
+++ b/hack/generator/pkg/astmodel/code_generation_context.go
@@ -14,7 +14,7 @@ import (
 // in which the field declaration occurs - for example in a file with two conflicting package references
 // a disambiguation must occur and field types must ensure they correctly refer to the disambiguated types
 type CodeGenerationContext struct {
-	packageImports map[PackageReference]PackageImport
+	packageImports *PackageImportSet
 	currentPackage PackageReference
 
 	generatedPackages map[PackageReference]*PackageDefinition
@@ -23,17 +23,15 @@ type CodeGenerationContext struct {
 // New CodeGenerationContext creates a new immutable code generation context
 func NewCodeGenerationContext(
 	currentPackage PackageReference,
-	packageImports map[PackageImport]struct{},
+	packageImports *PackageImportSet,
 	generatedPackages map[PackageReference]*PackageDefinition) *CodeGenerationContext {
 
-	packageImportsMap := make(map[PackageReference]PackageImport)
-	for imp := range packageImports {
-		packageImportsMap[imp.PackageReference] = imp
-	}
+	imports := EmptyPackageImportSet()
+	imports.Merge(packageImports)
 
 	return &CodeGenerationContext{
 		currentPackage:    currentPackage,
-		packageImports:    packageImportsMap,
+		packageImports:    imports,
 		generatedPackages: generatedPackages}
 }
 
@@ -43,23 +41,19 @@ func (codeGenContext *CodeGenerationContext) CurrentPackage() PackageReference {
 }
 
 // PackageImports returns the set of package references in the current context
-func (codeGenContext *CodeGenerationContext) PackageImports() map[PackageReference]PackageImport {
+func (codeGenContext *CodeGenerationContext) PackageImports() *PackageImportSet {
 	// return a copy of the map to ensure immutability
-	result := make(map[PackageReference]PackageImport)
-
-	for key, value := range codeGenContext.packageImports {
-		result[key] = value
-	}
+	result := EmptyPackageImportSet()
+	result.Merge(codeGenContext.packageImports)
 	return result
 }
 
 // GetImportedPackageName gets the imported packages name or an error if the package was not imported
 func (codeGenContext *CodeGenerationContext) GetImportedPackageName(reference PackageReference) (string, error) {
-	packageImport, ok := codeGenContext.packageImports[reference]
+	packageImport, ok := codeGenContext.packageImports.ImportFor(reference)
 	if !ok {
 		return "", errors.Errorf("package %s not imported", reference)
 	}
-
 	return packageImport.PackageName(), nil
 }
 
--- a/hack/generator/pkg/astmodel/code_generation_context.go
+++ b/hack/generator/pkg/astmodel/code_generation_context.go
@@ -26,7 +26,7 @@ func NewCodeGenerationContext(
 	packageImports *PackageImportSet,
 	generatedPackages map[PackageReference]*PackageDefinition) *CodeGenerationContext {
 
-	imports := EmptyPackageImportSet()
+	imports := NewPackageImportSet()
 	imports.Merge(packageImports)
 
 	return &CodeGenerationContext{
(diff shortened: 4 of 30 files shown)


hack/generator/pkg/astmodel/package_import_set.go
```diff
@@ -0,0 +1,81 @@
...
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ */
+
+package astmodel
+
+import "sort"
+
+// PackageImportSet represents a set of distinct PackageImport references
+type PackageImportSet struct {
```
matthchr, line 11: Would it be better to just make this:
`type PackageImportSet map[PackageImport]struct{}`? We do this for other set types (for example `Types`, `TypeNameSet`)

hack/generator/pkg/astmodel/package_import_set.go
```diff
@@ -0,0 +1,81 @@
...
+	imports map[PackageImport]bool
```
matthchr, line 12: I think the standard we've used for set is `map[T]struct{}`, see `TypeNameSet`

hack/generator/pkg/astmodel/allof_type.go
```diff
@@ -117,7 +117,7 @@ func (allOf AllOfType) AsDeclarations(codeGenerationContext *CodeGenerationConte
 }
 
 // RequiredImports returns the union of the required imports of all the AllOf types
-func (allOf AllOfType) RequiredImports() []PackageReference {
+func (allOf AllOfType) RequiredImports() *PackageImportSet {
```
matthchr, line 120: I think you're conflating `PackageReference` and `PackageImport` here (and in other places below where this same substitution was performed).

A `PackageReference` is a link in a dependency graph (Type `A` uses types from package `P`).
A `PackageImport` is a specific, realized import that satisfies a `PackageReference` requirement.

The key difference being that a `PackageImport` can have a name (`myfancypackage` rather than just a reference `github.com/foo/bar/v1`). `PackageImport` was created to only be used at code generation time (You can see they are first constructed in `generateImports` of `file_definition.go`).

These two types could theoretically be merged into one type, but I personally find it clearer as two - one which is about the graph relationship (and thus cannot be named), and one which is about code (and thus can be named).

I think this means either:
1. We should have a `PackageReferenceSet` in addition to a `PackageImportSet`.
2. We should leave `[]PackageReference`'s as is, and only change places where we were already using `map[PackageImport]` to `PackageImportSet`.

hack/generator/pkg/astmodel/arm_type.go
```diff
@@ -26,7 +26,7 @@ func MakeArmType(object ObjectType) ArmType {
 }
 
 // RequiredImports returns a list of packages required by this type
-func (at ArmType) RequiredImports() []PackageReference {
+func (at ArmType) RequiredImports() *PackageImportSet {
```
matthchr, line 29: As mentioned above in long comment, I think this is conflating two things (Import vs Reference)

Won't repeat comment on each location though as there are a number of occurrences.

hack/generator/pkg/astmodel/package_import_set.go
matthchr: If you did this you could also drop the `*` everywhere on this type since `map`'s are always pointers.

hack/generator/pkg/astmodel/package_import_set.go
```diff
@@ -0,0 +1,81 @@
...
+		if imp.PackageReference.Equals(ref) {
+			return imp, true
+		}
+	}
+
+	return PackageImport{}, false
+}
+
+// AsSlice() returns a sorted slice containing all the imports
+func (set *PackageImportSet) AsSlice() []PackageImport {
```
matthchr, line 65: There seem to be at least two legitimate orderings to items from this set:
1. Ordered alphabetically including name (if applicable) - which is what this does.
2. Ordered alphabetically ignoring name (i.e. just ordered alphabetically by the `PackageImport`).

I could envision cases where we want all the package references with a name first and all without afterwards too maybe?

I guess I'm saying: Why is this particular ordering blessed as _the_ ordering for getting a slice? What if I wanted a slice in a different order? Is this ordering really intrinsic to the `PackageImportSet`-ness of things or would it make more sense to have this logic defined as a helper in the context of whatever code is trying to render these things into lines in a file?

matthchr (changes requested):

hack/generator/pkg/astmodel/allof_type.go
matthchr: Or, if we really want to remove the distinction between the two at the type level (and just say that when a package import doesn't have a name then it can also be used in the "graph" context), then we ought to be deleting/combining `PackageImport` and `PackageReference` - but as I mentioned above I'm not a huge fan of that because I think they really are two different things.

hack/generator/pkg/astmodel/package_import_set.go
theunrepentantgeek: There was a reason why that didn't work - but I don't recall it. I'll spike the change and will use it if it works.

hack/generator/pkg/astmodel/allof_type.go
theunrepentantgeek: Not conflating the two, but fixing a problem. The method is used to answer the question _"What imports does this Type need?"_ but the return type wasn't a list of imports, but a list of references. Some types need to have imports with a particular name, and couldn't specify it - so there was special casing for `metav1` elsewhere to compensate. This method isn't used for the reference graph.

matthchr: It's not possible for the type to answer the question: `what imports does this type need` though, because names need to be derived after conflict resolution which needs the picture of all imports in a file (not just a single type).

The way it was supposed to work here was the type specifies a package reference, and then at code-gen time it's named appropriately. We can have a chat about this if you want over teams?

hack/generator/pkg/astmodel/package_import_set.go
theunrepentantgeek: Changed the `AsSlice()` method to return in an undefined order and introduced `AsSortedSlice()` for when the order matters.

hack/generator/pkg/astmodel/package_import_set.go
theunrepentantgeek: Changed to use `map[PackageImport]struct{}`.

hack/generator/pkg/astmodel/allof_type_test.go
```diff
@@ -106,6 +106,6 @@ func TestAllOfRequiredImportsPanics(t *testing.T) {
 
 	x := AllOfType{}
 	g.Expect(func() {
-		x.RequiredImports()
+		x.RequiredPackageReferences()
```
matthchr, line 109: Nice

hack/generator/pkg/astmodel/code_generation_context.go
```diff
@@ -23,17 +23,15 @@ type CodeGenerationContext struct {
...
 	currentPackage PackageReference,
-	packageImports map[PackageImport]struct{},
+	packageImports *PackageImportSet,
 	generatedPackages map[PackageReference]*PackageDefinition) *CodeGenerationContext {
 
-	packageImportsMap := make(map[PackageReference]PackageImport)
-	for imp := range packageImports {
-		packageImportsMap[imp.PackageReference] = imp
-	}
+	imports := EmptyPackageImportSet()
```
matthchr, line 29: Wondering if this being `NewPackageImportSet` would:
1. Help discoverability through autocompletion (since `New`/`Make` are usually prefixes for these sorts of things?)
2. More closely match other "constructor" funcs.

hack/generator/pkg/astmodel/package_import_set.go
```diff
@@ -0,0 +1,138 @@
...
+
+// AddImport ensures the set includes the specified import
+// Adding an import already present in the set is fine.
+func (set *PackageImportSet) AddImport(packageImport PackageImport) {
+	set.imports[packageImport] = struct{}{}
+}
+
+// AddReference ensures this set includes an import of the specified reference
+// Adding a reference already in the set is fine.
+func (set *PackageImportSet) AddReference(ref PackageReference) {
```
matthchr, line 30: minor: I feel like given the ease of creating an import from a `PackageReference`, and the subtle distinction between the two of them, it would be better to not have this method on this type. It's only used in 1 place and I think it would be clearer if that usage were:
`set.AddImport(NewPackageImport(ref))` rather than `set.AddReference(ref)`.

The second flavoring suggests some "reference-aware storage" in the `set` which isn't the case.

hack/generator/pkg/astmodel/package_import_set.go
```diff
@@ -0,0 +1,138 @@
...
+
+// Contains allows checking to see if an import is included
+func (set *PackageImportSet) ContainsImport(packageImport PackageImport) bool {
+	_, found := set.imports[packageImport]
+	return found
+}
+
+// ImportFor looks up a package reference and returns its import, if any
+func (set *PackageImportSet) ImportFor(ref PackageReference) (PackageImport, bool) {
+	for imp := range set.imports {
```
matthchr, line 55: minor: The previous incarnations of code that used this did a lookup in a `map[PackageReference]PackageImport`.

I think this is fine (overall numbers are small), but we are losing a bit of efficiency here and I figured I would call it out in case others objected.

hack/generator/pkg/astmodel/package_import_set.go
matthchr: Nice.

hack/generator/pkg/astmodel/package_import_set.go
```diff
@@ -0,0 +1,138 @@
...
+		return result[i].String() < result[j].String()
+	})
+
+	return result
+}
+
+// AsSortedSlice() return a sorted slice containing all the imports
+// less specifies how to order the imports
+func (set *PackageImportSet) AsSortedSlice(less func(i PackageImport, j PackageImport) bool) []PackageImport {
+	result := set.AsSlice()
```
matthchr, line 81: Sort happens 2x in this codepath - might be worth a private function for initial sliceification (non-deterministic order) which `AsSlice` and `AsSortedSlice` could call? That would seem to give a bit more code reuse and also avoid the double-sort.

hack/generator/pkg/astmodel/package_import_set.go
matthchr: Actually, it's not an undefined order right now - is that intentional (so that it's at least repeatable), or an oversight? See my other comment below about double-sorting as well.

hack/generator/pkg/astmodel/package_import_set.go
```diff
@@ -0,0 +1,138 @@
...
+			found = true
+		}
+	}
+
+	if found {
+		set.AddImport(NewPackageImport(ref).WithName(name))
+	}
+}
+
+// ByPackageImportName() orders PackageImport instances by name,
```
matthchr, line 112: minor: Comment doesn't match function name

hack/generator/pkg/astmodel/package_import_set.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+// ApplyName replaces any existing PackageImport for the specified reference with one using the
+// specified name
+func (set *PackageImportSet) ApplyName(ref PackageReference, name string) {
+	found := false
+	// Iterate over a slice as that freezes the list
+	for _, imp := range set.AsSlice() {
+		if imp.PackageReference.Equals(ref) {
+			set.Remove(imp)
```
matthchr, line 102: There's a possible problem here (I'm not 100% sure it's a problem but it might be):
This set allows representation of multiple imports with the same package but different names. That's really an error (we have a TODO I believe to check for it and resolve it), but the way this function is written it will remove all N occurrences of an import with matching `PackageReference`s and replace them with a single one (the one so named).

On the one hand... this function almost could be used to solve the TODO - on the other hand it's not clear that is the intent of this function and it feels like that might be an unintentional side-effect rather than the real goal.

I wonder if it would be better to make the caller supply the specific `PackageImport` they want to modify - that way you can do a direct lookup and it also sidesteps the duplicates problem. I think doing this at the only current callsite would be easy.
Implementation could also then be something like the below (which seems a bit cleaner to me):
```
current, ok := set.imports[packageImport]
if !ok {
    return
}
set.Remove(imp)
set.AddImport(packageImport.WithName(name))
```

hack/generator/pkg/astmodel/package_import_set_test.go
```diff
@@ -0,0 +1,196 @@
...
+	set.AddImport(simpleTestImport)
+	set.Remove(pathTestImport)
+	g.Expect(set.ContainsImport(pathTestImport)).To(BeFalse())
+}
+
+/*
+ * ByNameInGroups() tests
+ */
+
+func TestByNameInGroups_AppliesExpectedOrdering(t *testing.T) {
```
matthchr, line 147: Nice -- I saw the impl of the `ByNameInGroups` function and thought: "Hey that's actually kinda complicated, we should have a test".

You already wrote it though, sweet!

hack/generator/pkg/astmodel/resource.go
```diff
@@ -226,22 +226,22 @@ func (definition *ResourceType) WithOwner(owner *TypeName) *ResourceType {
...
 
 	if definition.status != nil {
-		typeImports = append(typeImports, definition.status.RequiredImports()...)
+		references = append(references, definition.status.RequiredPackageReferences()...)
 	}
 
-	typeImports = append(typeImports, MetaV1PackageReference)
-	typeImports = append(typeImports, MakeGenRuntimePackageReference())
-	typeImports = append(typeImports, MakeExternalPackageReference("fmt"))
+	references = append(references, MetaV1PackageReference)
```
matthchr, line 237: random observation (don't take this as something you should fix...): It's sorta wonky we have a hardcoded reference for `metav1PackageReference`, and we could for `GenRuntime` too but we don't. Just once of those really small inconsistencies.

matthchr (approved):
Few comments, none are really blockers though

hack/generator/pkg/astmodel/package_import_set_test.go
theunrepentantgeek: Writing the test was the only way to make sure it worked right; the comparison is complicated enough that I tied myself in knots overthinking it. 😁

hack/generator/pkg/astmodel/package_import_set.go
theunrepentantgeek: I've changed the internal representation of `PackageImportSet` to `map[PackageReference]PackageImport` to allow for this quick lookup.

hack/generator/pkg/astmodel/package_import_set.go
theunrepentantgeek: An oversight on my part - should have removed the sorting from `AsSlice()` but I got sidetracked doing the testing on `AsSortedSlice()`.

hack/generator/pkg/astmodel/package_import_set.go
theunrepentantgeek: See my other comment - `AsSlice()` shouldn't have been sorting.

hack/generator/pkg/astmodel/package_import_set.go
theunrepentantgeek: The implementation was deliberate - and would only have impacted things if there were multiple imports for the same reference. I've changed the type definition to use `map[PackageReference]PackageImport` which makes the implementation here trivial and prevents multiple incompatible imports of the same reference, solving the TODO as well.

hack/generator/pkg/astmodel/package_import_set.go
theunrepentantgeek: Fixed comment to match function name.

hack/generator/pkg/astmodel/resource.go
theunrepentantgeek: Something to fix later, perhaps.

hack/generator/pkg/astmodel/package_import_set.go
matthchr: Yeah that's also fine

PR #275, opened 2020-10-09
Add support for resource ownership in generic controller

matthchr:


hack/generated/controllers/generic_controller.go
```diff
@@ -411,6 +439,45 @@ func (gr *GenericReconciler) MonitorDeployment(ctx context.Context, action Recon
...
+		}, nil
+	}
+
+
+	err = gr.applyOwnership(ctx, data)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	// TODO: Fast requeue as we're moving to the next stage... Do we prefer this or doing it "all at once"?
```
matthchr, line 475: We should answer this question (do we like this as a separate "state" in the state machine-thingy we have going on) or would we rather just always do this at the beginning of the `BeginDeployment` phase?

hack/generated/controllers/generic_controller.go
```diff
@@ -535,3 +601,69 @@ func (gr *GenericReconciler) Patch(
...
+}
+
+func (gr *GenericReconciler) deleteResourceSucceeded(ctx context.Context, data *ReconcileMetadata) error {
+	err := gr.Patch(ctx, data, func(ctx context.Context, mutData *ReconcileMetadata) error {
+		controllerutil.RemoveFinalizer(data.metaObj, GenericControllerFinalizer)
+		return nil
+	})
+
+	// patcher will try to fetch the object after patching, so ignore not found errors
+	if apierrors.IsNotFound(err) {
```
Porges, line 662: is this bit of code what `client.IgnoreNotFound` does?

matthchr: [yes](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/client/interfaces.go#L135)

matthchr: Will update to use that

hack/generated/controllers/generic_controller.go
```diff
@@ -535,3 +601,69 @@ func (gr *GenericReconciler) Patch(
 		Name:      data.metaObj.GetName(),
 	}, data.metaObj)
 }
+
+// isOwnerReady returns true if the owner is ready or if there is no owner required
+func (gr *GenericReconciler) isOwnerReady(ctx context.Context, data *ReconcileMetadata) (bool, error) {
+	_, err := gr.ResourceResolver.GetOwner(ctx, data.metaObj)
+	if err != nil {
+		switch causeErr := errors.Unwrap(err).(type) {
```
babbageclunk, line 609: I think this should be using either `errors.Cause` or `.As`. `errors.Unwrap` only peels off one layer, so you're assuming too much about the call stack of `GetOwner` - if it gets changed to add another layer of context at some point this will stop working.

hack/generated/controllers/generic_controller.go
babbageclunk: I kind of like having a separate state for this - is the state stored on the resource? Oh, I guess it's logged by the recorder. This way if there was some bug that was preventing the owner from being ready it would be clearer why the object was stuck, rather than just being in BeginDeployment - that would make it look like maybe arm was the holdup.

babbageclunk (approved):
Looks nice - I think the separate state is better for debuggability. A question about the owner not found check.

hack/generated/controllers/generic_controller.go
matthchr: I used `Cause` originally but the problem is that it unwraps to the first error that doesn't implement the `Cause` interface - but since I made my error implement that interface it bypassed that.

`As` sounds right though I think.

hack/generated/controllers/generic_controller.go
matthchr: Right now there isn't a real state - it's examining a combination of properties on the resource to determine what state it's in and then acting based on that. The advantage being that we can't have drift between reported state and the properties that the state is about.

Specifically in the case of ownership, if there's no Kubernetes owner assigned to the resource and there's supposed to be (can be determined by checking `Owner()`) then we do the `ManageOwnership` step, whose output is a resource with the Kubernetes `OwnerReference` set. If you then went and manually cleared `OwnerReference` (but changed nothing else), we'd do that step again, whereas if we had a tracked state we wouldn't (because the tracked state wasn't updated back to `NeedsOwnership`) if that makes sense?

I think the notion of states is useful for us internally but may not make sense to expose to the customer (maybe events would be ok?). They're a cognitive tool to make it easier to make reconciliation eventually consistent.

hack/generated/controllers/generic_controller.go
```diff
@@ -535,3 +601,69 @@ func (gr *GenericReconciler) Patch(
...
+		controllerutil.RemoveFinalizer(data.metaObj, GenericControllerFinalizer)
+		return nil
+	})
+
+	// patcher will try to fetch the object after patching, so ignore not found errors
+	if apierrors.IsNotFound(err) {
+		err = nil
+	}
+
+	data.log.V(0).Info("Deleted resource")
```
theunrepentantgeek, line 666: Can/should this log message identify the resource just deleted?

theunrepentantgeek (approved):
One minor comment, otherwise looks good.

hack/generated/controllers/generic_controller.go
```diff
@@ -216,26 +234,39 @@ func (gr *GenericReconciler) StartDeleteOfResource(
...
+	}
+
+	// TODO: Drop this entirely in favor if calling the genruntime.MetaObject interface methods that
+	// TODO: return the data we need.
+	// TODO(matthchr): For now just emulate this with reflection
 	resource, err := gr.constructArmResource(ctx, data)
+
 	if err != nil {
-		return ctrl.Result{}, errors.Wrapf(err, "converting %q to armResourceSpec", data.metaObj.GetName())
+		// If the error is that the owner isn't found, that probably means that the resource was deleted
```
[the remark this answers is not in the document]
matthchr: I've clarified this comment significantly.

hack/generated/controllers/generic_controller.go
matthchr: All of the `data.log` methods already have attached as key-value pairs the name and namespace of the resource in question, so the logging only needs to append whatever other information is relevant. We probably are "over-specific" in that some log messages are re-including info that's already in the log context.

PR #276, opened 2020-10-13
Improve OneOf property naming

Porges:
When producing an object for a `OneOf` type, naming information is lost when collapsing types like `AllOf` into objects. To fix this we will produce names before collapsing types in order to preserve the original names.

If there are multiple names produced for an `AllOf` we prefer the one that is derived from a `TypeName` or is at a lower “inheritance” depth.


hack/generator/pkg/codegen/pipeline_convert_allof_and_oneof_to_objects.go
```diff
@@ -118,63 +122,70 @@ type synthesizer struct {
...
-		return nil
-	})
+func (ns names) betterThan(other names) bool {
+	if ns.goodName && !other.goodName {
+		return true
+	}
 
-	if err != nil {
-		return nil, err
+	if other.goodName && !ns.goodName {
```
theunrepentantgeek, line 139: I think this would be clearer (easier to compare both conditions) if the parameters were in the same order as above:
```suggestion
	if !ns.goodName && other.goodName {
```

hack/generator/pkg/codegen/pipeline_convert_allof_and_oneof_to_objects.go
```diff
@@ -118,63 +122,70 @@ type synthesizer struct {
...
-	// OpenAPI does, see https://github.com/Azure/k8s-infra/issues/71
-	var properties []*astmodel.PropertyDefinition
-
-	propertyDescription := "Mutually exclusive with all other properties"
-	err := oneOf.Types().ForEachError(func(t astmodel.Type, i int) error {
-		prop, err := s.convertToOneOfProperty(i, t)
-		if err != nil {
-			return err
-		}
+type names struct {
```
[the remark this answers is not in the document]
theunrepentantgeek: Following the precept that short class names represent overarching concepts (and short variable names represent short lived things), perhaps this needs a longer name?

hack/generator/pkg/codegen/pipeline_convert_allof_and_oneof_to_objects.go
```diff
@@ -118,63 +122,72 @@ type synthesizer struct {
...
-	// OpenAPI does, see https://github.com/Azure/k8s-infra/issues/71
-	var properties []*astmodel.PropertyDefinition
-
-	propertyDescription := "Mutually exclusive with all other properties"
-	err := oneOf.Types().ForEachError(func(t astmodel.Type, i int) error {
-		prop, err := s.convertToOneOfProperty(i, t)
-		if err != nil {
-			return err
-		}
+type propertyNames struct {
```
matthchr, line 125: Is this a `propertyNames` or a `propertyName`?

hack/generator/pkg/codegen/pipeline_convert_allof_and_oneof_to_objects.go
```diff
@@ -118,63 +122,72 @@ type synthesizer struct {
...
-		return nil, err
+	if !ns.isGoodName && other.isGoodName {
+		return false
 	}
 
-	objectType := astmodel.NewObjectType().WithProperties(properties...)
-	objectType = objectType.WithFunction(astmodel.NewOneOfJSONMarshalFunction(objectType, s.idFactory))
+	// both are good or !good
+	// return the name closer to the top
+	// (i.e. “lower” in inheritance hierarchy)
```
matthchr, line 145: minor: Add a comment explaining what this does?

I _think_ it's basically saying that the property closest to the "leaf" of the oneof/allof graph is the one whose name wins? Why is that a desirable characteristic?

matthchr (approved):

hack/generator/pkg/codegen/pipeline_convert_allof_and_oneof_to_objects.go
Porges: It's the other way around, “higher” in the inheritance hierarchy = deeper in the graph. I’ll expand this a bit more

hack/generator/pkg/codegen/pipeline_convert_allof_and_oneof_to_objects.go
Porges: There’s two names for the property in here…

matthchr: I guess I was looking at it like there's one name, and some metadata (the `json` bit).
It's not a big deal though.

PR #277, opened 2020-10-14
Workaround for child resources not referenced in main spec

matthchr:


theunrepentantgeek (approved):
Good enough workaround for now.

PR #278, opened 2020-10-15
Add support for ResourceGroup

matthchr:
- Includes support for "subscription" deployments in addition to "resourceGroup" deployments.

### DIFF
--- a/hack/generated/controllers/generic_controller.go
+++ b/hack/generated/controllers/generic_controller.go
@@ -206,7 +206,7 @@ func (gr *GenericReconciler) DetermineReconcileAction(data *ReconcileMetadata) (
 	// TODO: See: https://github.com/Azure/k8s-infra/issues/274
 	// Determine if we need to update ownership first
 	owner := data.metaObj.Owner()
-	if owner != nil && owner.Kind != "ResourceGroup" && len(data.metaObj.GetOwnerReferences()) == 0 { // TODO: Remove RG hack
+	if owner != nil && len(data.metaObj.GetOwnerReferences()) == 0 {
 		// TODO: This could all be rolled into CreateDeployment if we wanted
 		return ReconcileActionManageOwnership, gr.ManageOwnership, nil
 	}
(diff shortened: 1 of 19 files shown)


hack/generated/controllers/generic_controller.go
```diff
@@ -559,26 +562,46 @@ func (gr *GenericReconciler) resourceSpecToDeployment(ctx context.Context, data
...
+			res.ResourceGroup(),
+			deploymentName,
+			res.Spec())
+	case *genruntime.SubscriptionResource:
+		deployment = gr.ARMClient.NewSubscriptionDeployment(
+			res.Location(),
+			deploymentName,
+			res.Spec())
+	default:
+		// TODO: This is a code bug -- do we want to panic or error here
```
Porges, line 594: panic!

hack/generated/pkg/genruntime/base_types.go
```diff
@@ -89,24 +89,52 @@ func (resource *armResourceImpl) GetId() string {
...
 
-func (r *DeployableResource) ResourceGroup() string {
+func NewDeployableSubscriptionResource(location string, spec ArmResourceSpec) *SubscriptionResource {
+	return &SubscriptionResource{
+		location: location,
+		spec:     spec,
+	}
+}
+
+// DeployableResource represents a resource which can be deployed to Azure. In addition
```
Porges, line 110: wrong type name

hack/generator/azure-arm.yaml
```diff
@@ -203,6 +203,12 @@ typeTransformers:
   # only readonly properties that have all been removed in the swagger
   # -> deployment template conversion.
 
+  - group: deploymenttemplate
+    name: ResourceLocations
+    target:
+      name: string
+    because: Modeling this as an enum doesn't work well in the context of CRDs because new regions are regularly added
```
Porges, line 210: can we rewrite this to add (instead of replace) `string`? it's nice to have the current list for autocomplete

probably we need to do some work to enable this

Porges (approved):

hack/generated/controllers/generic_controller.go
matthchr: removed TODO

hack/generator/azure-arm.yaml
matthchr: Not sure I follow what you're saying here? You mean to add an element to `oneOf` so it's always `oneOf` the `enum` and `string`?

That's actually what it is on ResourceBase I think - the problem is the resulting type is pretty awkward to work with.

PR #279, opened 2020-10-15
Improve filtering to allow semicolon separated wildcards

theunrepentantgeek:


Porges:
We can maintain more of the original code if we turn `;` into `|` (and parenthesize each alternative), which would let us keep one regex (as opposed to an array) everywhere.

theunrepentantgeek:
I've made the suggested change @Porges - it's simpler now. 😁

matthchr (approved):

PR #281, opened 2020-10-16
Add struct/interface embedding support

matthchr:


hack/generator/pkg/astmodel/object_type.go
```diff
@@ -105,11 +109,28 @@ func (objectType *ObjectType) Properties() []*PropertyDefinition {
...
 
+// EmbeddedProperties returns all the embedded properties
+// A sorted slice is returned to preserve immutability and provide determinism
+func (objectType *ObjectType) EmbeddedProperties() []*PropertyDefinition {
+	var result []*PropertyDefinition
+	for _, embedded := range objectType.embedded {
+		result = append(result, embedded)
+	}
+
+	// TODO: This doesn't work
```
matthchr, line 126: Oops, need to fix this (and add a test for the Optional case)

matthchr:
`controller-gen` doesn't support embedded type?

matthchr:
Confirmed that this works fine as far as `controller-gen` is concerned.

Porges (approved):

PR #283, opened 2020-10-19
Remove unnecessary references in resource.go

matthchr:
- genruntime can be used by interfaces, which will be automatically
    included via InterfaceImplementer. It can also be included in
    specs for properties such as Owner, but in that case it will
    already be included on the Spec as a requirement (via the property
    itself).
  - fmt can be used in functions, but those will report their need for
    it directly.


babbageclunk (approved):
Nice.

PR #285, opened 2020-10-20
Add simple controller tests

matthchr:
- Add simple test infrastructure and two end to end tests.
- Build Docker image and deploy controller    
  - Add Dockerfile for generic controller.
  - Add RBAC support for generic controller manager.
  - Update "make deploy" target to deploy controller.
  - Update "make kind-create" target to allow importing
    private docker images.
  - Add new "make test-int" target for running integration tests.
- Install generated CRDs in Kind to ensure validity.


hack/generated/Makefile
```diff
@@ -2,9 +2,10 @@ SHELL             = /bin/bash
 PACKAGE           = github.com/Azure/k8s-infra/hack/generated
 APP               = k8sinfra-controller
 
-timestamp         := $(shell /bin/date "+%Y%m%d-%H%M%S")
-# CONFIG_REGISTRY = kind-registry:5000/fake/k8s-infra-controller:latest
-IMG               ?= k8s-infra-generated-contoller:$(timestamp)
+# timestamp         := $(shell /bin/date "+%Y%m%d-%H%M%S")
```
matthchr, line 5: clean up commented out stuff

hack/generated/config/manager/manager_image_patch.yaml
```diff
@@ -0,0 +1,33 @@
...
+            - name: AZURE_TENANT_ID
+              valueFrom:
+                secretKeyRef:
+                  name: k8sinfra-controller-settings
+                  key: AZURE_TENANT_ID
+            - name: AZURE_SUBSCRIPTION_ID
+              valueFrom:
+                secretKeyRef:
+                  name: k8sinfra-controller-settings
+                  key: AZURE_SUBSCRIPTION_ID
```
matthchr, line 33: Missing newline at EOF

hack/generated/controllers/controller_test.go
```diff
@@ -0,0 +1,113 @@
...
+	g.Expect(rg.Status.ID).ToNot(BeNil())
+	armId := rg.Status.ID
+
+	// Delete the resource group
+	err = testContext.KubeClient.Delete(ctx, rg)
+	g.Expect(err).ToNot(HaveOccurred())
+	g.Eventually(testContext.Ensure.Deleted(ctx, rg)).Should(BeTrue())
+
+	// Ensure that the resource group was really deleted in Azure
+	// TODO: Do we want to just use an SDK here? This process is quite icky as is...
```
matthchr, line 48: This is really icky -- we need a better way to do it; unclear to me if that's fixing our internal interface to make this cleaner or just using an SDK.

If using an SDK we'll end up with a ton of SDKs which is sorta a downside.

hack/generated/controllers/suite_test.go
```diff
@@ -0,0 +1,152 @@
...
+	err = testContext.KubeClient.Create(ctx, sharedResourceGroup)
+	if err != nil {
+		return errors.Wrapf(err, "creating shared resource group")
+	}
+
+	// TODO: Should use AzureName here once it's always set
+	log.Printf("Created shared resource group %s\n", sharedResourceGroup.Name)
+
+	// It should be created in Kubernetes
+	// TODO: Generic polling helper needed here
```
matthchr, line 67: Probably need to do this

hack/generated/pkg/testcommon/ensure.go
```diff
@@ -0,0 +1,133 @@
...
+func NewEnsure(c client.Client, stateAnnotation string, errorAnnotation string) *Ensure {
+	return &Ensure{
+		kubeClient:      c,
+		stateAnnotation: stateAnnotation,
+		errorAnnotation: errorAnnotation,
+	}
+}
+
+// ProvisioningComplete ensures that the object specified has reached a terminal provisioning state.
+// THIS STATE CAN BE FAILED. Use the ProvisioningStateSuccessful function to ensure it was successful.
```
matthchr, line 35: Should be "Use the ProvisioningSuccessful function"

hack/generated/pkg/testcommon/ensure.go
```diff
@@ -0,0 +1,133 @@
...
+		}
+
+		// Ensure that we have the right annotation
+		state := metaObj.GetAnnotations()[e.stateAnnotation]
+
+		return state == string(armclient.SucceededProvisioningState) || state == string(armclient.FailedProvisioningState), nil
+	}
+}
+
+// ProvisioningStateSuccessful checks to ensure the provisioning state of the resource is successful.
```
matthchr, line 62: minor: Wrong name

scripts/deploy_testing_secret.sh
```diff
@@ -0,0 +1,26 @@
...
+kind: Secret
+metadata:
+  name: k8sinfra-controller-settings
+  namespace: k8s-infra-system
+stringData:
+  AZURE_SUBSCRIPTION_ID: "$AZURE_SUBSCRIPTION_ID"
+  AZURE_TENANT_ID: "$AZURE_TENANT_ID"
+  AZURE_CLIENT_ID: "$AZURE_CLIENT_ID"
+  AZURE_CLIENT_SECRET: "$AZURE_CLIENT_SECRET"
+EOF
```
matthchr, line 26: Newline

hack/generated/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+COPY pkg/ pkg/
+
+# Build
+# TODO: Use Makefile here -- right now it's awkward to do so because:
+#   1. tools.mk is required for the makefile from the above directory, but Dockerfile can only look in its directory and below.
+#   2. Having Dockerfile here but building it from above could work except that there's another Dockerfile and a .dockerignore
+#      up above that break things. For now we just build by hand
+# RUN make build
+
+# TODO: Do we want CGO_ENALBED=0 and the other options below in the makefile?
```
Porges, line 27: I would say so

hack/generated/Makefile
```diff
@@ -70,16 +65,52 @@ generate: $(CONTROLLER_GEN) $(CONVERSION_GEN) ## Generate code
...
+# Eventually when we start supporting features like MSI that only work in Azure,
+# it's likely we'll need to make the integration tests (or at least some of them)
+# run in an AKS cluster.
+.PHONY: test-int-no-cleanup
+test-int-no-cleanup: $(ROOT_DIR)/.env ; $(info $(M) running controller integration test…)
+	# MUST be executed as single command, or env vars will not propagate to test execution
+	$(Q) . $(ROOT_DIR)/.env && $(GO) test $(GO_DIRS_TO_TEST) # TODO: better way to just run integration tests?
+
+.PHONY: test-int
+test-int: kind-create deploy test-int-no-cleanup kind-delete cleanup-test-azure-resources
```
Porges, line 95: we really are not using Make as intended but as a series of commands. perhaps we should look into an alternate solution

hack/generated/controllers/controller_test.go
```diff
@@ -0,0 +1,113 @@
...
+			Properties: &storage.StorageAccountPropertiesCreateParameters{
+				AccessTier: &accessTier,
+			},
+		},
+	}
+	err := testContext.KubeClient.Create(ctx, acct)
+	g.Expect(err).ToNot(HaveOccurred())
+
+	// It should be created in Kubernetes
+	g.Eventually(testContext.Ensure.ProvisioningComplete(ctx, acct)).Should(BeTrue())
```
matthchr, line 91: Ensure returning a boolean means the error messages when this fails are not that useful -- maybe an `err` instead?

hack/generated/Dockerfile
matthchr: Added

hack/generated/Makefile
matthchr: Agreed -- file an item for it maybe?

hack/generated/controllers/controller_test.go
matthchr: Tried err and it was pretty ugly. Will attempt what @porges suggested of:
```
g.Eventually(acct).Should(BeProvisioned())
g.Expect(acct).Should(BeSuccessfullyProvisioned())
```

hack/generated/controllers/suite_test.go
```diff
@@ -0,0 +1,175 @@
...
+	ctx := context.Background()
+
+	// List all of the resource groups
+	rgList := &resources.ResourceGroupList{}
+	err := testContext.KubeClient.List(ctx, rgList)
+	if err != nil {
+		return errors.Wrap(err, "listing resource groups")
+	}
+
+	// Delete any leaked resource groups
```
theunrepentantgeek, line 87: How do we know that they leaked from us?

Would be really bad form to delete any resource groups set up that aren't to do with our operator.

hack/generated/controllers/suite_test.go
```diff
@@ -0,0 +1,175 @@
...
+	log.Println("Finished common controller test teardown")
+	return nil
+}
+
+type controllerTest struct {
+	name string
+	test func(ctx ControllerTestContext, t *testing.T)
+}
+
+func runTests(t *testing.T, tests []controllerTest) {
```
theunrepentantgeek, line 132: Seems like we should be able to make this easier to call than having to create the slice ahead of time.

theunrepentantgeek (commented):
Partial review.

hack/generated/controllers/controller_test.go
```diff
@@ -0,0 +1,103 @@
...
+	g := NewGomegaWithT(t)
+	ctx := context.Background()
+
+	// Create a resource group
+	rg := testContext.NewTestResourceGroup()
+	err := testContext.KubeClient.Create(ctx, rg)
+	g.Expect(err).ToNot(HaveOccurred())
+
+	// It should be created in Kubernetes
+	g.Eventually(testContext.Ensure.ProvisioningComplete(ctx, rg)).Should(BeTrue())
```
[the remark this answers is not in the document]
matthchr: They're already configured for this package in the `setup` method - I set them to 2m timeout and 5s polling interval, but we can fine-tune them as well.

hack/generated/controllers/controller_test.go
```diff
@@ -0,0 +1,103 @@
...
+				AccessTier: &accessTier,
+			},
+		},
+	}
+	err := testContext.KubeClient.Create(ctx, acct)
+	g.Expect(err).ToNot(HaveOccurred())
+
+	// It should be created in Kubernetes
+	g.Eventually(testContext.Ensure.ProvisioningComplete(ctx, acct)).Should(BeTrue())
+	g.Expect(testContext.Ensure.ProvisioningSuccessful(ctx, acct)).Should(BeTrue())
```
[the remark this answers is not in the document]
matthchr: > Note that Should and To are just syntactic sugar and are functionally identical. Same is the case for ToNot and NotTo.
There isn't one -- but I should change the second `Should` here to be a `To` because it reads better. I'll do that.

hack/generated/controllers/suite_test.go
matthchr: By definition the list of resources discovered in this cleanup method is from Kubernetes, so currently the assumption is that we own the cluster and it's being solely used for testing.

Since we have a specific test namespace I'll limit this to querying just that, so that it'll play nicer in a cluster being used for something else as well.

hack/generated/controllers/suite_test.go
matthchr: I think I'm going to change things back to using `TestMain` which will allow us to have test autodiscover (but has other issues...)

hack/generated/controllers/controller_test.go
matthchr: Take a look at the new way -- I think it's quite a bit cleaner

hack/generated/controllers/controller_test.go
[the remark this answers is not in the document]
matthchr: Actually deleted this second line as we don't need it anymore with the new custom gomega.matcher.

hack/generated/controllers/controller_test.go
```diff
@@ -0,0 +1,103 @@
...
+	err = testContext.KubeClient.Delete(ctx, acct)
+	g.Expect(err).ToNot(HaveOccurred())
+	g.Eventually(testContext.Ensure.Deleted(ctx, acct)).Should(BeTrue())
+
+	// Ensure that the resource group was really deleted in Azure
+	// TODO: Do we want to just use an SDK here? This process is quite icky as is...
+	exists, err := testContext.AzureClient.HeadResource(
+		ctx,
+		armId,
+		"2019-04-01")
```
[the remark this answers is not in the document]
matthchr: Yes, it does.

I'm going to split this bit off to be a different PR I think though, as I feel like the right way to do this is to make the client support retries automatically...

I spoke with @devigned and he suggested that in the actual reconciliation loop we probably don't need retries (as the outer loop will retry anyway), although there are throttling considerations we'll need to take into account there.

In any case for now I think the move is to give this call retries in the tests but not in the actual controller.

.github/workflows/test.yml
```diff
@@ -45,6 +50,15 @@ jobs:
       - name: Run generator CI (Linux)
         if: ${{ runner.os == 'Linux' }} 
         run: make -C ./hack/generator ci
+      - name: Run generated CI (Linux)
+        if: ${{ runner.os == 'Linux' }}
+        run: make -C ./hack/generated ci
```
Porges, line 55: duplicated entries here

hack/generated/controllers/controller_test.go
```diff
@@ -0,0 +1,101 @@
...
+	namer := testContext.Namer.WithSeparator("")
+
+	// Create a storage account
+	accessTier := storage.StorageAccountPropertiesCreateParametersAccessTierHot
+	acct := &storage.StorageAccount{
+		ObjectMeta: testContext.MakeObjectMetaWithName(namer.GenerateName("stor")),
+		Spec: storage.StorageAccounts_Spec{
+			Location:   testContext.AzureRegion,
+			ApiVersion: "2019-04-01", // TODO: This should be removed from the storage type eventually
+			Owner:      genruntime.KnownResourceReference{Name: testContext.SharedResourceGroup.Name},
```
Porges, line 65: perhaps we should have a helper on the context for this?

hack/generated/controllers/suite_test.go
```diff
@@ -0,0 +1,189 @@
...
+	*testcommon.TestContext
+	SharedResourceGroup *resources.ResourceGroup
+}
+
+func setup() (ControllerTestContext, error) {
+	ctx := context.Background()
+	log.Println("Running test setup")
+
+	SetDefaultEventuallyTimeout(DefaultResourceTimeout)
+	SetDefaultEventuallyPollingInterval(5 * time.Second)
```
Porges, line 45: can we qualify these with `gomega` in this file specifically? just for clarity

hack/generated/pkg/testcommon/be_provisioned_matcher.go
```diff
@@ -0,0 +1,82 @@
...
+	if actual == nil {
+		return false
+	}
+
+	obj, err := actualAsObj(actual)
+	if err != nil {
+		panic(err)
+	}
+	state := obj.GetAnnotations()[m.ensure.stateAnnotation]
+	return state != string(armclient.SucceededProvisioningState) && state != string(armclient.FailedProvisioningState)
```
Porges, line 81: function `IsInTerminalState(obj)` for clarity

.github/workflows/test.yml
matthchr: Oops

hack/generated/Makefile
```diff
@@ -70,15 +63,60 @@ generate: $(CONTROLLER_GEN) $(CONVERSION_GEN) ## Generate code
...
+
+.PHONY: test-int-cover
+test-int-cover: kind-create deploy test-int-no-cleanup-cover kind-delete cleanup-test-azure-resources
+
+# Cleanup resource groups created by tests -- this isn't strictly required as the tests
+# clean up after themselves, but doing it here anyway just to be doubly sure we don't leak
+# resources in cases where the test pass is terminated, panics, etc
+.PHONY: cleanup-test-azure-resources
```
matthchr, line 114: I just realized that this isn't going to work well for parallel CI runs - will need to revisit it

hack/generated/controllers/suite_test.go
matthchr: Good idea

hack/generated/controllers/controller_test.go
matthchr: Done

hack/generated/Makefile
matthchr: Improved this to do time-based deletion where we won't delete resource groups that were leaked if they're newer than a day (this caters for parallel runs and anyway the cleanup make target is just for extra safety as the tests cleanup after themselves anyway.

hack/generated/controllers/controller_test.go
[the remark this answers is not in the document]
matthchr: I have a follow up PR once this gets merged where I add support for retries on the client and make use of that support in the tests

Porges (approved):

PR #287, opened 2020-10-21
Deepcopy maps, arrays and optional fields when converting to/from ARM types

### LINKED ISSUE #286: ARM conversion functions should deepcopy maps, arrays and optional fields
At the moment any fields that have the same type on the kubernetes and ARM versions of a struct are assigned directly. This means that any reference type values will be shared between the k8s and ARM struct. This is different from any fields that are generated types, since thos are necessarily deepcopied while converting them to the ARM equivalent.  That's bound to bite us if we don't change it.

babbageclunk:
Before this some of the fields in a converted object were deepcopied in the process of converting the generated types to their ARM equivalents but other reference-type fields (maps/arrays/optionals of enums or primitive types) would be shared between the ARM object and the kubernetes one. That seems a lot more error-prone - it would be easy to inadvertently update some shared data because we'd been lulled into a false sense of security by the other fields that weren't shared.

Fixes #286.

### DIFF
--- a/hack/generator/pkg/astmodel/armconversion/convert_to_arm_function_builder.go
+++ b/hack/generator/pkg/astmodel/armconversion/convert_to_arm_function_builder.go
@@ -184,6 +184,32 @@ func (builder *convertToArmBuilder) propertiesWithSameNameAndTypeHandler(
 	if !ok || !toProp.PropertyType().Equals(fromProp.PropertyType()) {
 		return nil
 	}
+
+	if typeRequiresCopying(fromProp.PropertyType()) {
+		// We can't get away with just assigning this field, since
+		// it's a reference type. Use the conversion code to copy the
+		// elements.
+		source := &ast.SelectorExpr{
+			X:   builder.receiverIdent,
+			Sel: ast.NewIdent(string(toProp.PropertyName())),
+		}
+		destination := &ast.SelectorExpr{
+			X:   builder.resultIdent,
+			Sel: ast.NewIdent(string(toProp.PropertyName())),
+		}
+		return builder.toArmComplexPropertyConversion(
+			complexPropertyConversionParameters{
+				source:            source,
+				destination:       destination,
+				destinationType:   toProp.PropertyType(),
+				nameHint:          string(toProp.PropertyName()),
+				conversionContext: nil,
+				assignmentHandler: assignmentHandlerAssign,
+				sameTypes:         true,
+			},
+		)
+	}
+
 	result := astbuilder.SimpleAssignment(
 		&ast.SelectorExpr{
 			X:   builder.resultIdent,
@@ -242,7 +268,15 @@ func (builder *convertToArmBuilder) toArmComplexPropertyConversion(
 	case *astmodel.MapType:
 		return builder.convertComplexMapProperty(params)
 	case astmodel.TypeName:
+		if params.sameTypes {
+			// The only type names we leave alone are enums, which
+			// don't need conversion.
+			return builder.assignPrimitiveType(params)
+		}
 		return builder.convertComplexTypeNameProperty(params)
+	case *astmodel.PrimitiveType:
+		// No conversion needed in this case.
+		return builder.assignPrimitiveType(params)
 	default:
 		panic(fmt.Sprintf("don't know how to perform toArm conversion for type: %T", params.destinationType))
 	}
@@ -254,6 +288,15 @@ func assignmentHandlerDefine(lhs ast.Expr, rhs ast.Expr) ast.Stmt {
 
 func assignmentHandlerAssign(lhs ast.Expr, rhs ast.Expr) ast.Stmt {
 	return astbuilder.SimpleAssignment(lhs, token.ASSIGN, rhs)
+}
+
+// assignPrimitiveType just assigns source to destination directly,
+// no conversion needed.
+func (builder *convertToArmBuilder) assignPrimitiveType(
+	params complexPropertyConversionParameters) []ast.Stmt {
+	return []ast.Stmt{
+		params.assignmentHandler(params.destination, params.source),
+	}
 }
 
 // convertComplexOptionalProperty handles conversion for optional properties with complex elements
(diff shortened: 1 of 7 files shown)


hack/generator/pkg/astmodel/armconversion/convert_to_arm_function_builder.go
```diff
@@ -498,3 +537,11 @@ func callToArmFunction(source ast.Expr, destination ast.Expr, methodName string)
 
 	return results
 }
+
+func typeRequiresCopying(theType astmodel.Type) bool {
+	switch theType.(type) {
+	case *astmodel.OptionalType, *astmodel.MapType, *astmodel.ArrayType:
```
theunrepentantgeek, line 543: Why does an `OptionalType` require copying? Doesn't that depend on what's wrapped?

babbageclunk: No, because it's implemented as a pointer. If the conversion function just has:
```go
k8sResource.OptionalString = armResource.OptionalString
```
... and then some code assigns through the pointer:
```go
*armResource.OptionalString = "oops"
```
... then `k8sResource` will see that update.

https://play.golang.org/p/RSR5io24eXB

babbageclunk: I mean, it does depend on what's wrapped as well. If it's a generated non-enum type then it won't come through this check because the source and destination types will be different - one will be an Arm type.

hack/generator/pkg/astmodel/armconversion/convert_to_arm_function_builder.go
```diff
@@ -270,11 +313,17 @@ func (builder *convertToArmBuilder) convertComplexOptionalProperty(
 	tempVarIdent := ast.NewIdent(builder.idFactory.CreateIdentifier(params.nameHint+"Typed", astmodel.NotExported))
 	tempVarType := destinationType.Element()
 
+	newSource := &ast.UnaryExpr{
+		X:  params.source,
+		Op: token.MUL,
```
Porges, line 318: 😢

hack/generator/pkg/astmodel/armconversion/convert_to_arm_function_builder.go
```diff
@@ -498,3 +537,11 @@ func callToArmFunction(source ast.Expr, destination ast.Expr, methodName string)
 
 	return results
 }
+
+func typeRequiresCopying(theType astmodel.Type) bool {
```
Porges, line 541: should we put this on `Type` instead?

Porges: it would be _so nice_ if you could add interfaces to external types…

babbageclunk: Yeah, I'll move this. One slightly weird thing is that in general it should return true for ObjectType, but that path will never be exercised by this code because it only checks when the source and destination types are the same, and we'll always generate a different dest type in that case.

babbageclunk: Oh duh, that's not right - this is really "does this type have reference behaviour?", so for ObjectType it should be false.

babbageclunk: Man, that's not right either! An ObjectType needs deep-copying if it has any reference types inside it. (Makes me feel like maybe I shouldn't promote this up to Type?)

hack/generator/pkg/astmodel/armconversion/shared.go
```diff
@@ -129,7 +129,12 @@ type complexPropertyConversionParameters struct {
 	destinationType   astmodel.Type
 	nameHint          string
 	conversionContext []astmodel.Type
-	assignmentHandler func(result ast.Expr, destination ast.Expr) ast.Stmt
+	assignmentHandler func(destination, source ast.Expr) ast.Stmt
+
+	// sameTypes indicates that the source and destination types are
```
matthchr, line 134: Rather than "base-level" say something about "type"? Basically this is saying: "If the types are the same we don't need to worry about renaming any of the `TypeName`'s"?

hack/generator/pkg/codegen/testdata/ArmResource/Arm_test_simple_resource_array_properties.golden
```diff
@@ -133,18 +133,19 @@ func (fakeResourceSpec *FakeResource_Spec) ConvertToArm(name string) (interface{
...
 		elemTyped := make(map[string]FooArm)
-		if item != nil {
-			for key, value := range item {
-				elem, err := value.ConvertToArm(name)
-				if err != nil {
-					return nil, err
-				}
-				elemTyped1 := elem.(FooArm)
-				elemTyped[key] = elemTyped1
+		for key, value := range item {
```
matthchr, line 142: Ah, range over nil just skips it doesn't it!

Nice catch.

matthchr (approved):
I like that you were able to make the `convertComplexOptionalProperty` in `convert_from_arm.go` look the same as the one in `convert_to_arm.go` -- the files were supposed to be mirrors of one another but that bit hadn't made it all the way.

hack/generator/pkg/astmodel/armconversion/convert_to_arm_function_builder.go
babbageclunk: I carried on with this for the various Type implementations because it seemed like it could be useful for other situations, but it's not definable for TypeName without having access to the full set of types. So I'm leaving this as a specific utility function for this code (where the philosophical problem happens to be sidestepped).

Porges: all good!

hack/generator/pkg/astmodel/armconversion/shared.go
babbageclunk: Good call, I've reworded it to say "so no conversion between Arm and non-Arm types is required (although structure copying is)".

hack/generator/pkg/astmodel/armconversion/convert_to_arm_function_builder.go
theunrepentantgeek: Fair enough.

PR #288, opened 2020-10-21
Remove use of global context

Porges:


hack/generated/controllers/suite_test.go
```diff
@@ -124,22 +124,45 @@ func teardown() error {
...
+}
+
+func Test_Controller_Integrations_Slow(t *testing.T) {
+	if testing.Short() {
+		t.Skip("Skipping slow tests in short mode")
 	}
 
-	os.Exit(code)
+	tests := []controllerTest{
+		{"ResourceGroup CRUD", Integration_ResourceGroup_CRUD},
```
matthchr, line 157: kinda annoying that we have to add stuff here when we add a test. I wonder if we could discover them via reflection?

hack/generated/controllers/suite_test.go
```diff
@@ -124,22 +124,45 @@ func teardown() error {
...
+type controllerTest struct {
+	name string
+	test func(ctx ControllerTestContext, t *testing.T)
+}
 
-	err := setup()
-	if err != nil {
-		panic(err)
+func runTests(t *testing.T, tests []controllerTest) {
+	ctx, setupErr := setup()
```
matthchr, line 133: Are we worried about `ctx context.Context` and `ctx ControllerTestContext` confusion here?

Should we call the `ControllerTestContext` something else?

matthchr (approved):

hack/generated/controllers/suite_test.go
Porges: no, there is no way to reflect over packages in Go

PR #292, opened 2020-10-28
Add support for running operator tests in envtest

Porges:
You can now run the controller tests with a flag `-envtest` to use the `envtest` infrastructure. This means you do not need to spin up a kubernetes cluster or run the operator on its own container.


hack/generated/controllers/suite_test.go
```diff
@@ -41,19 +55,101 @@ func (tc *ControllerTestContext) SharedResourceGroupOwner() genruntime.KnownReso
 	return genruntime.KnownResourceReference{Name: tc.SharedResourceGroup.Name}
 }
 
-func setup() error {
+func setupEnvTest() (*rest.Config, error) {
+	envtestContext = &EnvtestContext{
+		testenv: envtest.Environment{
+			CRDDirectoryPaths: []string{
+				"../config/crd/bases/valid",
```
Porges, line 62: TODO: I made this 'valid' directory to put only the needed CRDs in as not all of them are able to be loaded (due to the Any-type stuff that @babbageclunk is working on…)

hack/generated/pkg/testcommon/wait.go
```diff
@@ -18,49 +19,49 @@ func WaitFor(ctx context.Context, timeout time.Duration, check func(context.Cont
 	waitCtx, cancel := context.WithTimeout(ctx, timeout)
 	defer cancel()
 
-	done := false
-
-	for !done {
+	for {
```
Porges, line 22: This was reworked as it slept for an additional period in the success case.

hack/generated/controllers/suite_test.go
```diff
@@ -31,6 +38,13 @@ const (
 )
 
 var testContext ControllerTestContext
+var envtestContext *EnvtestContext
+
+type EnvtestContext struct {
```
matthchr, line 43: Is it `Envtest` or `EnvTest`? I guess `envtest` is the name of the module so it's all one word?

hack/generated/controllers/suite_test.go
```diff
@@ -41,19 +55,101 @@ func (tc *ControllerTestContext) SharedResourceGroupOwner() genruntime.KnownReso
...
+		return nil, errors.Wrapf(err, "creating ARM client")
+	}
+
+	log.Print("Registering custom controllers")
+	errs := controllers.RegisterAll(
+		envtestContext.manager,
+		armClient,
+		controllers.KnownTypes,
+		klogr.New(),
+		controller.Options{})
```
Porges, line 102: @matthchr should we be setting anything here?

hack/generated/pkg/testcommon/wait.go
```diff
@@ -18,49 +19,49 @@ func WaitFor(ctx context.Context, timeout time.Duration, check func(context.Cont
...
 
-// testMainWrapper is a wrapper that can be called by TestMain so that we can use defer
 func SetupTeardownTestMain(
 	m *testing.M,
-	skipShortTests bool,
+	skipSlowTests bool,
 	setup func() error,
 	teardown func() error) int {
 
-	if skipShortTests {
```
matthchr, line 48: Thanks -- old name was super wrong =)

matthchr (commented):
LGTM but since you have a TODO you're still working on will save signoff til after you're done with that.

hack/generated/controllers/suite_test.go
matthchr: Looks like standard usage is showing just `concurrency(1)`, so I'm thinking no?

matthchr (approved):

PR #294, opened 2020-10-29
Refactor arm client to use autorest stack

matthchr:
- Added support for retries as suggested by @theunrepentantgeek.
- Added some end to end tests for the `armclient` module.


matthchr:
This is a draft as it's not really done. It's also going to conflict quite a bit with #292 by @porges, my plan is to let that merge first and then deal with the merge - posting this a bit early though in case @Porges wants to look at it.

matthchr:
This is probably ready for review now

Porges (approved):

PR #295, opened 2020-10-29
Replace AnyTypes with apiextensions.JSON 

### LINKED ISSUE #133: Fix handling of "empty" types
There are a number of these that are due to conversion from Swagger where all of the fields of a type are read-only:

For example:
1. https://schema.management.azure.com/schemas/2019-03-01/Microsoft.Compute.Galleries.json 
    - GalleryProperties.Identifier
2. https://schema.management.azure.com/schemas/2019-07-01/Microsoft.Compute.json 
    - VirtualMachineIdentity.UserAssignedIdentities
Both of the above are types that are entirely readonly

This causes at least a few errors when controller-gen runs that come out looking like this:
```
/home/matthchr/work/github/k8s-infra/hack/generator/apis/microsoft.compute.galleries/v20180601/galleries_types.go:75:15: unsupported AST kind *ast.InterfaceType
```

babbageclunk:
Adds a pipeline stage to convert AnyTypes to `k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1.JSON`, which can store arbitrary JSON values in the API. This enables generating all of the packages that were previously being filtered out because they generated interface{} fields.

Removes the `filterOutDefinitionsUsingAnyType` stage, although leaves the `ensureDefinitionsDoNotUseAnyTypes` check at the end just in case later stages introduce AnyTypes again.

A lot of packages include fields that come out as `map[string]map[string]v1.JSON` (especially for UserAssignedIdentities fields), but `controller-gen` can't generate deep-copy code for these types. To handle these the stage collapses them down to `map[string]v1.JSON` as well. https://github.com/kubernetes-sigs/controller-tools/issues/287

Adds special-case handling in the ARM conversion generation to ensure that JSON fields are deep-copied when converting to/from ARM types.

There are still a couple of packages that need to be excluded because the ARM conversion code generated for them has an issue or hits `controller-gen` limitations.

Fixes #133.


hack/generator/azure-arm.yaml
```diff
@@ -70,96 +70,19 @@ exportFilters:
...
+  - action: exclude
+    group: microsoft.web
+    version: v20150801
+    because: ARM conversion is generating bad code for handling SitesSpecResourcesAppsettingsSpec.Name - should be casting to the right type
 
 # Exclusions for packages that currently produce types including AnyType.
 # TODO: get rid of these, either by chasing the teams generating
 # weird json or by handling them differently in the generator.
 anyTypePackages:
-  - microsoft.authorization/v20180301
```
Porges, line 78: wooh!

hack/generator/azure-arm.yaml
```diff
@@ -70,96 +70,19 @@ exportFilters:
     group: deploymenttemplate
     name: '*ResourceBase*'
     because: Not actually a resource, used for inheritance only.
+  - action: exclude
+    group: microsoft.machinelearning
+    version: v20170101
+    because: ExampleRequest.Inputs is a map[string][][]map[string]v1.JSON, which controller-gen doesn't know how to handle.
```
Porges, line 76: oh wow

Porges (approved):

Porges:
This came out nicely!

hack/generator/azure-arm.yaml
theunrepentantgeek: #mindboggling

hack/generator/pkg/astmodel/armconversion/shared.go
```diff
@@ -40,6 +41,30 @@ func (builder conversionBuilder) propertyConversionHandler(
...
+	}
+	assignmentHandler := params.assignmentHandler
+	if assignmentHandler == nil {
+		assignmentHandler = assignmentHandlerAssign
+	}
+	return []ast.Stmt{
+		assignmentHandler(params.destination, newSource),
+	}
+}
+
```
theunrepentantgeek, line 67: Once my Test Cases branch is merged, we should loop back here and use some of the helper methods I've written to simplify the code.

hack/generator/pkg/astmodel/json_type.go
```diff
@@ -0,0 +1,12 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ */
+
+package astmodel
+
+// APIExtensions is the package containing the type we use to
+// represent arbitrary JSON fields.
+const APIExtensions = "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
```
theunrepentantgeek, line 10: Consider changing this to a `PackageReference`
```suggestion
var APIExtensionsReference PackageReference = NewExternalPackageReference("k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1")
```

theunrepentantgeek (approved):

hack/generator/pkg/astmodel/json_type.go
babbageclunk: Yup, I've done this.

PR #299, opened 2020-10-29
Exclude the generated files from linting

matthchr:
- It takes a long time

### DIFF
--- a/hack/generated/Makefile
+++ b/hack/generated/Makefile
@@ -11,7 +11,8 @@ include ../../tools.mk
 CRD_OPTIONS     ?= "crd:crdVersions=v1,allowDangerousTypes=true"
 GO_DIRS         := $(shell $(GO) list -f '{{.Dir}}' ./...)
 # We exclude the apis folder because it's really large and test discovery takes a good amount of time (>10s)
-GO_DIRS_TO_TEST := $(shell $(GO) list -f '{{.Dir}}' ./... | grep -v /apis/)
+DIRS_TO_SKIP    := "/apis/(?!microsoft.resources)"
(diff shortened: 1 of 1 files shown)


hack/generated/Makefile
```diff
@@ -11,7 +11,11 @@ include ../../tools.mk
 CRD_OPTIONS     ?= "crd:crdVersions=v1,allowDangerousTypes=true"
 GO_DIRS         := $(shell $(GO) list -f '{{.Dir}}' ./...)
 # We exclude the apis folder because it's really large and test discovery takes a good amount of time (>10s)
-GO_DIRS_TO_TEST := $(shell $(GO) list -f '{{.Dir}}' ./... | grep -v /apis/)
+DIRS_TO_SKIP    := "/apis/(?!microsoft.resources)"
+# We ideally could just pass the DIRS_TO_SKIP to golangci-lint, but it uses Go's regex's which don't support
+# negative lookahead... so we have to do it ourselves with some horrible grep + sed magic
+GO_DIRS_TO_SKIP_LINT = $(shell $(GO) list -f '{{.Dir}}' ./... | grep -P $(DIRS_TO_SKIP) | tr '\n' '|' | sed 's/^/\x28/' | sed 's/.$$//' | sed 's/$$/\x29/')
```
babbageclunk, line 17: Maybe add a comment with what transformation this does? Then if someone comes up with a nicer way then they can be sure they're understanding the original right.

babbageclunk (approved):

hack/generated/Makefile
babbageclunk: Converts
```
dir1
dir2
dir3
```
... to the skip-dirs regex `(dir1|dir2|dir3)`

matthchr: Done

PR #300, opened 2020-10-30
Rework devcontainer and CI setup

Porges:
Alternate title: #300: “THIS IS DOCKER”.
THIS IS DOCKER

This change:
1. reworks the devcontainer setup for the purposes of the codegen work. We can put this into the README for newcomers to the project, which should help them get up and running quickly.
2. uses the same Dockerfile for local dev as for doing CI. This will help with replicability and dev sanity.
3. switches out Makefiles for Taskfiles (using [go-task](https://github.com/go-task/task)), as we are (ab)using `make` as a task runner.
4. as a side-effect we are now using the `envtest` integration test configuration which allows us to capture code coverage for these tests.


babbageclunk:
In my headcanon you created this just to grab the good PR number.

hack/generated/Taskfile.yml
```diff
@@ -0,0 +1,113 @@
...
+
+tasks:
+  # Note: test-int can be run manually but is not included here so as to avoid slow
+  # tests bogging down the dev cycle
+  default:
+    desc: Perform all local pre-commit tasks.
+    deps: [generate-crds, header-check, format-code, test]
+  
+  # There is a ci specific target because we want the CI pass to fail if
+  # the code has not been go fmt-ed, whereas locally we want "make all"
```
matthchr, line 30: not `make all` anymore?

hack/generated/Taskfile.yml
```diff
@@ -0,0 +1,113 @@
...
+  ci:
+    # no description as it is not for local use
+    # order must be explicit as CRDs must be generated before header check
+    cmds:
+    - task: generate-crds
+    - task: ci-rest
+
+  ci-rest:
+    # these can all be run in parallel
+    deps: [header-check, test-cover, test-integration-envtest]
```
matthchr, line 41: `test-integration-envtest-cover`?

hack/generated/Taskfile.yml
```diff
@@ -0,0 +1,113 @@
...
+
+  lint:
+    desc: Run lint checks.
+    cmds: [golangci-lint run -v]
```
matthchr, line 45: Accidentally dropped the improvements I made to exclude the apis folder, or was that on purpose?

hack/generated/Taskfile.yml
```diff
@@ -0,0 +1,113 @@
...
+    - cp config/crd/bases/microsoft.resources.infra.azure.com_resourcegroups.yaml config/crd/bases/valid
+    # -race fails at the moment in controller-runtime
+    - go test -covermode atomic -coverprofile=cover-integration-envtest.out -coverpkg="./..." ./controllers/... -envtest 
+    - gcov2lcov -infile cover-integration-envtest.out -outfile coverage-integration-envtest.lcov
+  
+  cleanup-azure-resources:
+    desc: Removes any old resources created by tests.
+    cmds:
+      - |
+        rgs=`az group list --query '[*].{Name: name, CreatedAt: tags.CreatedAt}' \
```
matthchr, line 103: minor: Include comment as well?

> This finds all resource groups which match the specified pattern (k8sinfratest) and are older than a day
(86400 seconds). This is a bit horrible but it works...

hack/generated/Taskfile.yml
matthchr: oh I see, this target just always does cover

.devcontainer/Dockerfile
```diff
@@ -1,75 +1,57 @@
-#-------------------------------------------------------------------------------------------------------------
-# Copyright (c) Microsoft Corporation. All rights reserved.
-# Licensed under the MIT License. See https://go.microsoft.com/fwlink/?linkid=2090316 for license information.
-#-------------------------------------------------------------------------------------------------------------
+# See here for image contents: https://github.com/microsoft/vscode-dev-containers/tree/v0.137.0/containers/go/.devcontainer/base.Dockerfile
+ARG VARIANT="1.15"
```
matthchr, line 2: Wondering if building this is going to slow things down significantly? Do you know how long it takes to build?

hack/generated/Taskfile.yml
Porges: do you think there's benefit to having a non-cover version?

.devcontainer/Dockerfile
Porges: it takes about 3 minutes (can see in the Checks below); the new validation takes 12 minutes end-to-end compared to 17 minutes for the current version

hack/generated/Taskfile.yml
matthchr: Is there a benefit to having a non-cover version for unit tests? I just was asking because we have cover vs not for the other test target and was wondering if there was a reason to avoid it for this target.

.devcontainer/Dockerfile
matthchr: Nice, if it's faster that's good. Might get even faster if you bypass the linting on the apis directory again (I think you dropped that before as I mentioned above)

hack/generated/Taskfile.yml
Porges: I think the perf impact is higher for unit tests; with integration tests we are doing more waiting.

hack/generated/Taskfile.yml
Porges: restored

.devcontainer/Dockerfile
Porges: fixed that, wasn't actually invoking lint on CI at all :embarrassing_face:

hack/generated/Taskfile.yml
Porges: Decided I will add a non-coverage version in a later change as `gcov2lcov` is quite slow.

matthchr (approved):

PR #306, opened 2020-10-30
Allow DefineFunc to define funcs without receiver

matthchr:


babbageclunk (approved):

theunrepentantgeek:
I've got a nicer version of this in my Test Cases branch, should get the PR out this week.

theunrepentantgeek (changes requested):
Probably best to compare contrast with my branch before we merge one or the other.

matthchr:
@theunrepentantgeek - Take a look at https://github.com/Azure/k8s-infra/pull/312/files#r516265457 which I left on your PR. I think we can probably fold this PR into yours, main question I have is around merge timelines and if it'd be better to split some of the `astbuilder` stuff out of your PR into its own PR.

theunrepentantgeek:
I've incorporated the extra check into #312

matthchr:
Abandoning this in favor of #313

PR #307, opened 2020-10-30
Mark storage version as part of pipeline

matthchr:
- This makes the golden tests more accurate as we can shim output without also losing storage versions.

### DIFF
--- a/hack/generator/pkg/codegen/code_generator.go
+++ b/hack/generator/pkg/codegen/code_generator.go
@@ -86,6 +86,8 @@ func corePipelineStages(idFactory astmodel.IdentifierFactory, configuration *con
 		createStorageTypes(),
 		simplifyDefinitions(),
 
+		markStorageVersion(),
+
 		// Safety checks at the end:
 		ensureDefinitionsDoNotUseAnyTypes(),
 		checkForMissingStatusInformation(),
--- a/hack/generator/pkg/codegen/pipeline_mark_storage_version.go
+++ b/hack/generator/pkg/codegen/pipeline_mark_storage_version.go
@@ -1,0 +1,109 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ */
+
+package codegen
+
+import (
+	"context"
+	"sort"
+
+	"github.com/pkg/errors"
+
+	"github.com/Azure/k8s-infra/hack/generator/pkg/astmodel"
+)
+
+// markStorageVersion creates a PipelineStage to mark a particular version as a storage version
+func markStorageVersion() PipelineStage {
+	return MakePipelineStage(
+		"markStorageVersion",
+		"Marking the latest version of each resource as the storage version",
+		func(ctx context.Context, types astmodel.Types) (astmodel.Types, error) {
+			updatedDefs, err := MarkLatestResourceVersionsForStorage(types)
+			if err != nil {
+				return nil, errors.Wrapf(err, "unable to mark latest resource version as storage version")
+			}
+
+			return updatedDefs, nil
+		})
+}
+
+// MarkLatestResourceVersionsForStorage marks the latest version of each resource as the storage version
+func MarkLatestResourceVersionsForStorage(types astmodel.Types) (astmodel.Types, error) {
+
+	result := make(astmodel.Types)
+	resourceLookup, err := groupResourcesByVersion(types)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, def := range types {
+		// see if it is a resource
+		if resourceType, ok := def.Type().(*astmodel.ResourceType); ok {
+
+			unversionedName, err := getUnversionedName(def.Name())
+			if err != nil {
+				// should never happen as all resources have versioned names
+				return nil, err
+			}
+
+			allVersionsOfResource := resourceLookup[unversionedName]
+			latestVersionOfResource := allVersionsOfResource[len(allVersionsOfResource)-1]
+
+			thisPackagePath := def.Name().PackageReference.PackagePath()
+			latestPackagePath := latestVersionOfResource.Name().PackageReference.PackagePath()
+
+			// mark as storage version if it's the latest version
+			isLatestVersion := thisPackagePath == latestPackagePath
+			if isLatestVersion {
+				def = astmodel.MakeTypeDefinition(def.Name(), resourceType.MarkAsStorageVersion()).
+					WithDescription(def.Description())
+			}
+		}
+		result.Add(def)
+	}
+
+	return result, nil
+}
+
+func groupResourcesByVersion(types astmodel.Types) (map[unversionedName][]astmodel.TypeDefinition, error) {
+
+	result := make(map[unversionedName][]astmodel.TypeDefinition)
+
+	for _, def := range types {
+		if astmodel.IsResourceDefinition(def) {
+			name, err := getUnversionedName(def.Name())
+			if err != nil {
+				// this should never happen as resources will all have versioned names
+				return nil, errors.Wrapf(err, "Unable to extract unversioned name in groupResources")
+			}
+
+			result[name] = append(result[name], def)
+		}
+	}
+
+	// order each set of resources by package name (== by version as these are sortable dates)
+	for _, slice := range result {
+		sort.Slice(slice, func(i, j int) bool {
+			return slice[i].Name().PackageReference.PackageName() < slice[j].Name().PackageReference.PackageName()
+		})
+	}
+
+	return result, nil
+}
+
+func getUnversionedName(name astmodel.TypeName) (unversionedName, error) {
+
+	if localRef, ok := name.PackageReference.AsLocalPackage(); ok {
+		group := localRef.Group()
+		return unversionedName{group, name.Name()}, nil
+	}
+
+	return unversionedName{}, errors.New("cannot get unversioned name from non-local package")
+}
+
+type unversionedName struct {
+	group string
+	name  string
+}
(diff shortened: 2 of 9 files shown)


matthchr:
@theunrepentantgeek - pinging you specifically on this as it's related to an area you're working on. This happening before we write the files out (as a separate step) is important so that we actually end up with types that have the right labels in memory. I need that in order to autogenerate some of the registration stuff for the controller.

It's also just cleaner too though I think. Let me know what you think.

hack/generator/pkg/codegen/pipeline_mark_storage_version.go
```diff
@@ -0,0 +1,112 @@
...
+
+	result := make(astmodel.Types)
+	resourceLookup, err := groupResourcesByVersion(types)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, def := range types {
+		// see if it is a resource
+		if resourceType, ok := def.Type().(*astmodel.ResourceType); ok {
```
theunrepentantgeek, line 42: Need to restrict the loop to looking in Storage packages - use `IsStoragePackageReference`

hack/generator/pkg/codegen/pipeline_mark_storage_version.go
```diff
@@ -0,0 +1,112 @@
...
+
+			unversionedName, err := getUnversionedName(def.Name())
+			if err != nil {
+				// should never happen as all resources have versioned names
+				return nil, err
+			}
+
+			allVersionsOfResource := resourceLookup[unversionedName]
+			latestVersionOfResource := allVersionsOfResource[len(allVersionsOfResource)-1]
```
theunrepentantgeek, line 51: Can't just take the latest version, we want to avoid picking preview releases for storage as they're too volatile.

See https://github.com/Azure/k8s-infra/blob/master/hack/generator/spec/case-study-rolling-storage-versions.md#version-2014-04-04-preview---schema-change

hack/generator/pkg/codegen/pipeline_mark_storage_version.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	for _, def := range types {
+		// see if it is a resource
+		if resourceType, ok := def.Type().(*astmodel.ResourceType); ok {
+
+			unversionedName, err := getUnversionedName(def.Name())
+			if err != nil {
+				// should never happen as all resources have versioned names
+				return nil, err
```
theunrepentantgeek, line 47: Use `errors.Wrap()` to give more information about the error

hack/generator/pkg/codegen/pipeline_mark_storage_version.go
```diff
@@ -0,0 +1,112 @@
...
+		})
+}
+
+// MarkLatestResourceVersionsForStorage marks the latest version of each resource as the storage version
+func MarkLatestResourceVersionsForStorage(types astmodel.Types) (astmodel.Types, error) {
+
+	result := make(astmodel.Types)
+	resourceLookup, err := groupResourcesByVersion(types)
+	if err != nil {
+		return nil, err
```
theunrepentantgeek, line 37: Use `errors.Warp()` to give more information about the error.

hack/generator/pkg/codegen/pipeline_mark_storage_version.go
```diff
@@ -0,0 +1,112 @@
...
+
+	return result, nil
+}
+
+func groupResourcesByVersion(types astmodel.Types) (map[unversionedName][]astmodel.TypeDefinition, error) {
+
+	result := make(map[unversionedName][]astmodel.TypeDefinition)
+
+	for _, def := range types {
+		if _, ok := def.Type().(*astmodel.ResourceType); ok {
```
theunrepentantgeek, line 78: Use `IsResourceDefinition()`

hack/generator/pkg/codegen/pipeline_mark_storage_version.go
```diff
@@ -0,0 +1,112 @@
...
+			name, err := getUnversionedName(def.Name())
+			if err != nil {
+				// this should never happen as resources will all have versioned names
+				return nil, errors.Wrapf(err, "Unable to extract unversioned name in groupResources")
```
theunrepentantgeek, line 82: Should we accumulate all the errors to return, or is this rare enough that returning the first instance will be fine?

hack/generator/pkg/codegen/code_generator.go
```diff
@@ -86,6 +86,8 @@ func corePipelineStages(idFactory astmodel.IdentifierFactory, configuration *con
 		createStorageTypes(),
 		simplifyDefinitions(),
 
+		markStorageVersion(),
```
theunrepentantgeek, line 89: Should this perhaps be plural? `markStorageVersions()` #conflicted

hack/generator/pkg/codegen/pipeline_mark_storage_version.go
```diff
@@ -0,0 +1,112 @@
...
+			if isLatestVersion {
+				def = astmodel.MakeTypeDefinition(def.Name(), resourceType.MarkAsStorageVersion()).
+					WithDescription(def.Description())
+			}
+
+			result.Add(def)
+		} else {
+			// otherwise simply add it
+			result.Add(def)
+		}
```
theunrepentantgeek, line 67: Since the **true** branch is overwriting `def`, and `result.Add(def)` is the last statement in both branches, you could move it out of both branches and get rid of the **false** branch completely.

theunrepentantgeek (changes requested):

hack/generator/pkg/codegen/pipeline_mark_storage_version.go
matthchr: My intent here was just to move the existing logic that was in `pipeline_export_generated_code` to its own stage, not necessarily to change the logic of it.

Did you want me to change the logic too?

hack/generator/pkg/codegen/pipeline_mark_storage_version.go
matthchr: See comment about about not changing the logic we had, just refactoring -- same questions apply here.

hack/generator/pkg/codegen/pipeline_mark_storage_version.go
matthchr: The error provided by the `getUnversionedName` function already contains all of the information we really would need here, there's not anything to add as far as I can tell.

hack/generator/pkg/codegen/pipeline_mark_storage_version.go
matthchr: There's not really much to add here other than "calling function X failed" - and the function errors can/should already be saying that so I don't think there's a need to wrap here?

hack/generator/pkg/codegen/pipeline_mark_storage_version.go
matthchr: I think rare enough that failing on the first one is fine

hack/generator/pkg/codegen/code_generator.go
matthchr: I think singular because at any given time there can only be a single storage version as far as each CRD is concerned?

hack/generator/pkg/codegen/pipeline_mark_storage_version.go
matthchr: Good eye -- done

hack/generator/pkg/codegen/pipeline_mark_storage_version.go
theunrepentantgeek: Happy for this to stay as it is - I can tweak as a part of upcoming PRs.

hack/generator/pkg/codegen/code_generator.go
theunrepentantgeek: Hmm, I guess it's a matter of perspective. Is cardinality governed by the version or the types?

There's only one version of `microsoft.batch` that is our canonical storage version. 
But there are very many types in that one version.

😁 

Happy for this to stay as it is.

hack/generator/pkg/codegen/pipeline_mark_storage_version.go
```diff
@@ -0,0 +1,109 @@
...
+
+// markStorageVersion creates a PipelineStage to mark a particular version as a storage version
+func markStorageVersion() PipelineStage {
+	return MakePipelineStage(
+		"markStorageVersion",
+		"Marking the latest version of each resource as the storage version",
+		func(ctx context.Context, types astmodel.Types) (astmodel.Types, error) {
+			updatedDefs, err := MarkLatestResourceVersionsForStorage(types)
+			if err != nil {
+				return nil, errors.Wrapf(err, "unable to mark latest resource versions for as storage versions")
```
Porges, line 25: the error message here doesn't make much sense

Porges (approved):

hack/generator/pkg/codegen/pipeline_mark_storage_version.go
matthchr: lol no it does not

hack/generator/pkg/codegen/pipeline_mark_storage_version.go
theunrepentantgeek: Fair enough.

hack/generator/pkg/codegen/pipeline_mark_storage_version.go
theunrepentantgeek: Fair enough.

theunrepentantgeek (approved):
Looks like a good start.

PR #309, opened 2020-10-30
OneOf MarshalJson function should only apply to Arm

matthchr:
- Allowing it to apply to Kubernetes types was never the intent. The idea was that customers
  would send a type to Kuberentes that looked "normal" as far as Kubernetes API guidelines
  were concerned - this means OneOf modelled as mutually exclusive properties. The Azure types
  must then have that shape transformed into the actual shape Azure expects.


babbageclunk (changes requested):
Making sure I understand the sequence of events here:

1. The OneOf/AllOf pipeline stage collapses OneOfs to explicit discriminated union types - eg, in the golden test Properties with .Foo, .Bar and .Baz (mutually exclusive)
2. It also adds a MarshalJSON to Properties so that it serialises directly as whichever property is is set (no wrapping object)
3. When creating the ARM type we also copy the MarshalJSON across (in transformTypeDefinition).
4. Then we remove MarshalJSON from the k8s type, since we want it to be represented explicitly in the serialisation in the k8s API.

Is that right? It feels like more implicit coupling between those stages than I'd like. The ARM stage assumes that the MarshalJSON function on the type is really for the ARM version of the type - we'd get bitten if we wanted to customise the serialisation of a type in any other way. 

Maybe it would be better to make that explicit? Add a place to store ARM-specific functions on ObjectType which the OneOf stage would use. Those functions wouldn't generate any code normally, but the ARM stage would know to promote them to actual methods when making the ARM version of a type. Still kind of messy (one stage stashing information for another stage), but hopefully less error prone?

theunrepentantgeek:
> Maybe it would be better to make that explicit? Add a place to store ARM-specific functions on ObjectType which the OneOf stage would use. Those functions wouldn't generate any code normally, but the ARM stage would know to promote them to actual methods when making the ARM version of a type. Still kind of messy (one stage stashing information for another stage), but hopefully less error prone?

I think this would be extremely messy - and what would we then do to handle the requirements for storage types? Add yet another specialist storage place on ObjectType for those? I don't think the approach scales.

That said, I agree that adding MarshalJSON to the k8s type only so it can be relocated to the ARM type is messy; can we defer the creation of that method until we have the right place for it to live?

hack/generator/pkg/astmodel/object_type.go
```diff
@@ -378,6 +378,15 @@ func (objectType *ObjectType) WithFunction(function Function) *ObjectType {
 	return result
 }
 
+// WithoutFunction creates a new ObjectType without the specified function
+func (objectType *ObjectType) WithoutFunction(name string) *ObjectType {
```
theunrepentantgeek, line 382: I saw this earlier and wasn't sure why we need it - sorta feels as though this is a way to paper over a deeper issue.

matthchr:
From @babbageclunk 
> Is that right? It feels like more implicit coupling between those stages than I'd like. The ARM stage assumes that the MarshalJSON function on the type is really for the ARM version of the type - we'd get bitten if we wanted to customise the serialisation of a type in any other way.

Yes, your breakdown of the sequence of events is correct, and I agree that it is way more coupling between stages than we should really have.

From @theunrepentantgeek 
> That said, I agree that adding MarshalJSON to the k8s type only so it can be relocated to the ARM type is messy; can we defer the creation of that method until we have the right place for it to live?

Not without some way of identifying an arbitrary `ObjectType` as really a `OneOf`. Right now the method on the `ObjectType` is serving as a sort of hacky label. Without it we need a different label.

In my opinion the right solution here is actually something like the labels solution which @theunrepentantgeek had proposed a while back. If we had a `[]string` labels (or maybe more well-typed than that), we could label this type as `oneof` and let any subsequent stage optionally read that label. It's technically true we would do this by defining another type as well (`OneOfKubernetesWrapper` or something), but it seems really heavy weight to do that to me as that would require updates to the visitor and a whole new type - of course this approach doesn't scale to multiple dimensions well either because if you want to also slice things by "Kuberentes types" vs "ARM types" for example then you need `OneOfKubernetes` and `OneOfArm`

hack/generator/pkg/astmodel/object_type.go
matthchr: Removed this in favor of a better way

dev.sh
```diff
@@ -11,7 +11,7 @@ if [ ! -f "$TOOL_DEST/task" ]; then # check for local installation
     fi
 fi
 
-export PATH="$PATH:$TOOL_DEST:$TOOL_DEST/kubebuilder/bin"
+export PATH="$TOOL_DEST:$TOOL_DEST/kubebuilder/bin:$PATH"
```
matthchr, line 14: We prepend here because we want to override whatever tools they have in their environment path already to ensure the right ones are used.

matthchr:
I've switched this to use labels which makes it significantly better.

Since I was using `FlaggedType` I also rolled the fixes to ensure that `FlaggedType` types make it all the way through the pipeline, so in the future if we want to add flags for things like `Spec` or `Status` we can and it'll make it to the pipeline stages where they are needed.

hack/generator/pkg/codegen/pipeline_name_types_for_crd.go
```diff
@@ -90,6 +90,30 @@ func nameInnerTypes(
...
+	visitor.VisitFlaggedType = func(this *astmodel.TypeVisitor, it *astmodel.FlaggedType, ctx interface{}) (astmodel.Type, error) {
+		// Because we're returning type names here, we need to look up the name returned by visit and wrap that with the correct flags
+		nameHint := ctx.(string)
+
+		name, err := this.Visit(it.Element(), nameHint)
+		if err != nil {
+			return nil, err
+		}
+
+		// There are no words for how much I want LINQ right here
```
Porges, line 102: ❤️

hack/generator/pkg/codegen/pipeline_remove_type_aliases.go
```diff
@@ -63,6 +63,8 @@ func resolveTypeName(visitor *astmodel.TypeVisitor, name astmodel.TypeName, type
 		return def.Name(), nil // must remain named for controller-gen
 	case astmodel.ValidatedType:
 		return def.Name(), nil // must remain named so there is somewhere to put validations
+	case *astmodel.FlaggedType:
+		return def.Name(), nil // must remain named so there is somewhere to put flags
```
Porges, line 67: uh, is this correct?

Porges (approved):

hack/generator/pkg/codegen/pipeline_remove_type_aliases.go
matthchr: They need to make it all the way through the pipeline though, so that each stage has an opportunity to consume/remove/alter them as appropriate.

Maybe I should add a check right before the output step that they're all gone? Or we could leave it as is where if a `FlaggedType` makes it through it doesn't really matter as it just delegates all of its behavior to its element.

matthchr: Maybe what's throwing you is the comment though... which yeah doesn't really make much sense now that I re-read it.

Porges: inlining the type here shouldn't remove the fact that it is a flagged type though, it just won't be named? or does it need to be named because the pipeline stage that expects to process it doesn't recurse through other types looking for flag?

matthchr: Actually I'm not sure -- I think that this needs to return the name here because that's what we're doing for `ObjectType` too. Basically this function is saying: "When I resolve a `TypeName` and find a `FlaggedType`, then it's not a type alias so we can keep the name"

The only way something is a type alias is if when you resolve `TypeName` you find another `TypeName` (which is why that's the main case that has special logic in the switch).

Porges: at least we should fix the comment :)

matthchr: Yeah, will fix the comment for sure - are you ok with the rest of it though (i.e. the fact these types progress through the pipeline and are treated like `ObjectType` here?)

babbageclunk (approved):
Yay, much nicer!

hack/generator/pkg/codegen/pipeline_convert_allof_and_oneof_to_objects_test.go
```diff
@@ -308,6 +308,8 @@ func TestOneOfResourceSpec(t *testing.T) {
 	g.Expect(err).To(BeNil())
 
 	result := synth.oneOfObject(oneOf, names)
+	result, err = astmodel.TypeAsObjectType(result)
```
theunrepentantgeek, line 311: I thought we were getting rid of th helper method `TypeAsObjectType()` as overpowered and likely to cause subtle bugs.

hack/generator/pkg/codegen/pipeline_name_types_for_crd.go
```diff
@@ -90,6 +90,30 @@ func nameInnerTypes(
...
+		var found astmodel.TypeDefinition
+		for i, item := range resultTypes {
+			if item.Name().Equals(name) {
+				found = item
+				resultTypes[i] = resultTypes[len(resultTypes)-1]
+				resultTypes = resultTypes[:len(resultTypes)-1]
+				break
```
theunrepentantgeek, line 109: What does this do? #opaque

hack/generator/pkg/codegen/pipeline_remove_type_aliases.go
theunrepentantgeek: I don't know about @Porges, but it make sense to me! 😁

matthchr: Improved the comment, left the rest as is

hack/generator/pkg/codegen/pipeline_name_types_for_crd.go
matthchr: Added this comment:
```
// The above visit of ObjectType will have mutated resultTypes to include a mapping of the type name
// to the object type. Because we need to preserve flag types, we must find the ObjectType and re-wrap
// it in the flags it had before. Note that we cannot just bypass the ObjectType visit as it may make mutations
// to the Object (to name the types of its properties) which we also need to preserve.
```

hack/generator/pkg/codegen/pipeline_convert_allof_and_oneof_to_objects_test.go
matthchr: Now that you mention it I do vaguely recall having a discussion about that. There do seem to be some risks, but I like the usage.

Maybe I should instead make a specific `ObjectTypeOrUnwrappedFlaggedType` (name TBD) that has similar usage for my use cases but only works on a singly-nested flagged type (to avoid the "accidentally unwrapped 90 things") problem.

PR #310, opened 2020-11-01
Derive the JSON omitempty annotation directly from the property type

theunrepentantgeek:
At the moment we've tightly coupled use of the `omitempty` JSON tag with configuration of whether the property is optional or mandatory in the JSON schema. This is causing issues where sometimes we don't round-trip via JSON with sufficient fidelity.

For example, an uninitialized slice differs from an empty slice, with different serializations:
```
// Uninitialized slice
{"Items":null}

// Initialized with zero items
{"Items":[]}
```

Unfortunately, if `omitempty` is specified, these both serialize by omitting the tag completely:
```
// Uninitialized slice
{}

// Initialized with zero items
{}
```

We need to keep the separate representations so that we can submit the right things to ARM. This also has the benefit of making our round-trip tests work without needing any kludges.

### DIFF
--- a/hack/generator/pkg/codegen/testdata/EmbeddedTypes/Embedded_type_simple_resource.golden
+++ b/hack/generator/pkg/codegen/testdata/EmbeddedTypes/Embedded_type_simple_resource.golden
@@ -43,7 +43,7 @@ type FakeResource_SpecArm struct {
 	ApiVersion       FakeResourceSpecApiVersion `json:"apiVersion"`
 	Color            *FakeResourceSpecColor     `json:"color,omitempty"`
 	Foo              FooArm                     `json:"foo"`
-	Name             string                     `json:"name"`
+	Name             string                     `json:"name,omitempty"`
 	OptionalFoo      *FooArm                    `json:"optionalFoo,omitempty"`
 	Type             FakeResourceSpecType       `json:"type"`
 }
@@ -92,7 +92,7 @@ type FakeResource_Spec struct {
 
 	//AzureName: The name of the resource in Azure. This is often the same as the name
 	//of the resource in Kubernetes but it doesn't have to be.
-	AzureName string                 `json:"azureName"`
+	AzureName string                 `json:"azureName,omitempty"`
 	Color     *FakeResourceSpecColor `json:"color,omitempty"`
 
 	// +kubebuilder:validation:Required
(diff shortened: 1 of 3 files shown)


babbageclunk:
It'd be good to have the rationale on the PR so we can see it easily later.

hack/generator/pkg/codegen/testdata/EmbeddedTypes/Embedded_type_simple_resource.golden
```diff
@@ -43,7 +43,7 @@ type FakeResource_SpecArm struct {
 	ApiVersion       FakeResourceSpecApiVersion `json:"apiVersion"`
 	Color            *FakeResourceSpecColor     `json:"color,omitempty"`
 	Foo              FooArm                     `json:"foo"`
-	Name             string                     `json:"name"`
+	Name             string                     `json:"name,omitempty"`
```
babbageclunk, line 46: This seems weird and makes me think that the right value for `jsonOmitIfEmpty(PrimitiveType)` is false. Why shouldn't it be?

hack/generator/pkg/codegen/testdata/EmbeddedTypes/Embedded_type_simple_resource.golden
```diff
@@ -92,7 +92,7 @@ type FakeResource_Spec struct {
 
 	//AzureName: The name of the resource in Azure. This is often the same as the name
 	//of the resource in Kubernetes but it doesn't have to be.
-	AzureName string                 `json:"azureName"`
+	AzureName string                 `json:"azureName,omitempty"`
```
babbageclunk, line 95: Same here.

babbageclunk (commented):
This looks mostly good, but I feel like having `omitempty` on mandatory fields like `Name` or `AzureName` gives the wrong idea.

Also, will there need to be another change in the arm code to ensure that map/array fields are initialised to empty values? Otherwise they'll be sent as nulls.

hack/generator/pkg/codegen/testdata/EmbeddedTypes/Embedded_type_simple_resource.golden
theunrepentantgeek: Partly this is because _whether-we-serialize_ is independent from _this-property-is-mandatory_.

Also, while the primary consumer of these files is another process, they are visible to end users; leaving out properties that have been left blank reduces the amount of dross that end users need to wade through. This is particularly relevant when we're reporting on errors, as many of those have extra fields that don't provide any value when blank.

theunrepentantgeek:
Due to ARM requirements, we actually need the `omitempty` annotation to be derived from the optional/mandatory nature of the property, it can't be decoupled. Abandoning this PR.

PR #311, opened 2020-11-02
Replay tests

Porges:
Use `go-vcr` to record interactions with ARM and play them back. This solves a few things:

1. Eventually we expect the test suite to be large and it will take a long time to run the entire thing, making it not feasible to run as part of PR validation.
2. On GitHub, PRs from forks cannot use secrets in any way so external contributors will not be able to have full validation run on their PRs. With record-replay we can validate against canned responses and external contributors can also update the cassette files locally using their own Azure subscription.

`envtest` is used to run the operator in-proc so that we can change the ARM client it is using. A separate `envtest` instance is used for each test in order to isolate them from one-another.

All name generation for resources and deployments has been changed to be deterministic, based upon the test (and resource) names. This is needed to successfully recreate identical request/responses. Subscription IDs are all replaced with the null UUID, so that if an external contributor updates the tests it (a) doesn’t leak their Subscription ID and (b) doesn’t result in a big churn across all cassette files. Authorization headers are always dropped from cassette files.


.github/workflows/pr-validation.yml
```diff
@@ -19,10 +19,7 @@ jobs:
       - name: Run CI tasks
         run: docker run -w /workspace -v $GITHUB_WORKSPACE:/workspace -e AZURE_TENANT_ID -e AZURE_CLIENT_ID -e AZURE_CLIENT_SECRET -e AZURE_SUBSCRIPTION_ID devcontainer:1.0 task ci
         env:
-          AZURE_TENANT_ID: ${{ secrets.AZURE_TENANT_ID }}
-          AZURE_CLIENT_ID: ${{ secrets.AZURE_CLIENT_ID }}
-          AZURE_CLIENT_SECRET: ${{ secrets.AZURE_CLIENT_SECRET }}
-          AZURE_SUBSCRIPTION_ID: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
+          AZURE_SUBSCRIPTION_ID: '82acd5bb-4206-47d4-9c12-a65db028483d'
```
Porges, line 22: we should post-process the cassette file to remove references to any specific subscription, so that anyone can update it

hack/generated/controllers/suite_test.go
```diff
@@ -98,13 +184,22 @@ func setupEnvTest() (*rest.Config, error) {
 		armClient,
 		controllers.KnownTypes,
 		klogr.New(),
-		controller.Options{})
+		controllers.Options{
+			DeploymentNameGenerator: func() (string, error) {
```
Porges, line 188: figure out a way to do this without order mattering (based on test name?)

hack/generated/pkg/testcommon/resource_namer.go
```diff
@@ -19,18 +18,19 @@ type ResourceNamer struct {
 	separator   string
 }
 
-func NewResourceNamer(prefix string, separator string, randomChars int) *ResourceNamer {
+func NewResourceNamer(prefix string, separator string, randomChars int, seed int64) *ResourceNamer {
```
Porges, line 21: need to make this based on test name rather than order

hack/generated/pkg/testcommon/kube_test_context.go
```diff
@@ -7,11 +7,14 @@ package testcommon
...
 import (
 	"context"
+	"log"
+	"testing"
 	"time"
 
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
matthchr, line 17: This is already included above? use `metav1` rather than `v1`?

hack/generated/pkg/testcommon/kube_test_context.go
```diff
@@ -22,57 +25,108 @@ import (
...
+type KubePerTestContext struct {
+	*KubeGlobalContext
+	KubeBaseTestContext
+
+	KubeClient client.Client
+	Ensure     *Ensure
+	Match      *KubeMatcher
+}
+
+func (tc *KubePerTestContext) createTestNamespace() error {
```
matthchr, line 124: inconsistent receivers -- ptr or not?

hack/generated/pkg/testcommon/kube_test_context_envtest.go
```diff
@@ -0,0 +1,131 @@
...
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/envtest"
+)
+
+func createEnvtestContext(perTestContext PerTestContext) (*KubeBaseTestContext, error) {
+	log.Printf("Creating envtest for test %s", perTestContext.TestName)
+
+	environment := envtest.Environment{
+		CRDDirectoryPaths: []string{
+			"../config/crd/bases/valid",
```
matthchr, line 29: Should this have a TODO or something that it should be changed... or go away?

hack/generated/pkg/testcommon/test_context.go
```diff
@@ -6,47 +6,160 @@ Licensed under the MIT license.
...
+		return PerTestContext{}, errors.Wrapf(err, "creating ARM client")
 	}
 
-	return &TestContext{
-		AzureClient:       armClient,
-		AzureRegion:       region,
-		AzureSubscription: subscription,
-		Namer:             NewResourceNamer(ResourcePrefix, "-", 6),
+	// replace the ARM client transport (a bit hacky)
+	httpClient := armClient.RawClient.Sender.(*http.Client)
```
matthchr, line 63: Can you not just do what you need to do with a `SendDecorator`? See its usage for retries in `raw_client`.

hack/generated/pkg/testcommon/test_context.go
```diff
@@ -6,47 +6,160 @@ Licensed under the MIT license.
...
+		}
+	} else {
+		// if we are replaying, we won't need auth
+		// and we use a dummy subscription ID
+		subscriptionID = uuid.Nil.String()
+		authorizer = nil
+	}
+
+	r.AddSaveFilter(func(i *cassette.Interaction) error {
+		// rewrite all request/response fields to hide the real subscription ID
```
matthchr, line 117: Nice

hack/generated/pkg/testcommon/test_context.go
```diff
@@ -6,47 +6,160 @@ Licensed under the MIT license.
...
 	}, nil
 }
 
-func (tc *TestContext) NewTestResourceGroup() *resources.ResourceGroup {
+func createRecorder(testName string, recordReplay bool) (autorest.Authorizer, string, *recorder.Recorder, error) {
+	cassetteName := "fixtures/" + testName
+
+	var err error
+	var r *recorder.Recorder
+	if recordReplay {
```
matthchr, line 90: minor: slightly confusing variable name...
are we recording or are we replaying? Seems it would be cleaner as `replay` or `record`

hack/generated/pkg/testcommon/test_context.go
matthchr: Ah, the issue is that the recorder is a `Transport`...

matthchr (changes requested):

hack/generated/pkg/testcommon/kube_test_context.go
Porges: Bah, gopls!

hack/generated/pkg/testcommon/kube_test_context.go
Porges: originally this did mutation. it does not now. changed to non-ptr

hack/generated/pkg/testcommon/kube_test_context_envtest.go
Porges: yes, it should

hack/generated/pkg/testcommon/test_context.go
Porges: @matthchr yes, for some reason ARM adds its own `Sender` interface even though `http.Roundtripper` with the exact same signature exists ¯\_(ツ)_/¯

hack/generated/controllers/fixtures/Test_ResourceGroup_CRUD.yaml
```diff
@@ -0,0 +1,601 @@
+---
+version: 1
+interactions:
```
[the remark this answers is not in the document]
Porges: I was going by the sample docs for `go-vcr` but I agree

hack/generated/pkg/testcommon/test_context.go
Porges: (once the `recorder` is created, it _does_ report which mode it is in (record/replay) and we use that below, but you can't start it in one of these modes)

matthchr (approved):

Taskfile.yml
```diff
@@ -1,16 +1,223 @@
...
+  # for go-fmt the root dir must be excluded (it recurses)
+  GENERATED_DIRS_TO_FMT_CMD: go list -f '{{"{{.Dir}}"}}' ./...  | grep -vE '/apis(/|$)|/generated$' | tr '\n' ' '
+  # for lint/test the root dir must be included
+  GENERATED_DIRS_TO_LINT_CMD: go list -f '{{"{{.Dir}}"}}' ./...  | grep -vE '/apis(/|$)' | tr '\n' ' '
 
 tasks:
+  default:
+    desc: Perform all fast local pre-commit tasks.
+    cmds: 
+    - task: generator:quick-checks
```
matthchr, line 20: Shouldn't the default target also do a build?

Porges: It runs tests which should check all the code?

matthchr: Ahh, didn't realize the test target automatically did a build - I guess it must though? (go test contains go build?).

Sounds good.

Porges: This now does a build anyway because the `generator:default/:ci` has to do a `generate-types` anyway (to prevent two concurrent `generate-types` happening in the `controller:default/:ci`… `go-task` should be fixing this soon I think)

PR #314, opened 2020-11-04
Exclude certain ResourceBase properties

matthchr:
- There are some properties that don't make sense in the context
    of CRDs.
  - Fix a bug with property transformers that caused them to only
    apply the first matching transformer.

### DIFF
--- a/hack/generator/azure-arm.yaml
+++ b/hack/generator/azure-arm.yaml
@@ -141,6 +141,41 @@ typeTransformers:
     remove: true
     because: it has no writable properties in swagger
 
+  - name: "*"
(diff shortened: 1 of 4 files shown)


hack/generator/azure-arm.yaml
```diff
@@ -141,6 +141,41 @@ typeTransformers:
     remove: true
     because: it has no writable properties in swagger
 
+  - name: "*"
+    ifType:
```
Porges, line 145: I'd prefer `ifType` to go after `property` here as that is the way it is applied

Porges (approved):

Issue #319, opened 2020-11-10
Bad pluralizations


Porges:
For example, “services” comes out as “servex”: https://github.com/gobuffalo/flect/issues/49

PR #320, opened 2020-11-10
Add tests for two more Storage resource types, and fixes to support

Porges:
* Add create/delete tests for the `StorageAccountsBlobService` and `StorageAccountsBlobServicesContainer` resource types. 
* Attempt to make deployment more robust, if we deploy a second time and the deployment already exists. This could happen due to faults or spurious failures in patching resources.
  * TODO: We need to check the existing deployment against spec and delete/recreate it if it is different?
*  Handle authentication configuration errors as early as possible; the `autorest` library does not check this and will give back an authorizer that does not work.
* Added temporary fix for #319.
* Fix bad log message in one of the matchers.
* Only save recording for a test when it is successful. This makes it easier to develop when trying to fix a test.

## Outstanding items

We need to think about how to handle resources like `StorageAccountsBlobService` better. It _must_ be named `"default"` to work but this is not enforced at the moment.

One idea is that we could merge this into the `StorageAccount` resource type so that it can be configured there, and the controller takes care of creating it separately.

Another issue is that the name is not going to be unique in Kubernetes. For resources with an owner we might need to scope the name under the owner name, but I’m not sure that this is possible; the object is created in Kubernetes before we can touch the name. The user can set AzureName to "default" and Name to something else but this is not obvious.

I opened an issue for this here: https://github.com/Azure/k8s-infra/issues/324

### DIFF
--- a/hack/generated/controllers/controller_test.go
+++ b/hack/generated/controllers/controller_test.go
@@ -143,6 +143,8 @@ func StorageAccount_BlobServices_CRUD(t *testing.T, testContext testcommon.KubeP
 	})
 
 	// TODO: Delete doesn't seem to work?
+	// — is this because it is not a real resource but properties on the storage account?
+	// We probably need to ask the Storage team.
 	/*
 		err = testContext.KubeClient.Delete(ctx, blobService)
 		g.Expect(err).ToNot(HaveOccurred())
--- a/hack/generated/controllers/controller_test.go
+++ b/hack/generated/controllers/controller_test.go
@@ -10,10 +10,12 @@ import (
 	"testing"
 
 	. "github.com/onsi/gomega"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	storage "github.com/Azure/k8s-infra/hack/generated/apis/microsoft.storage/v20190401"
 	"github.com/Azure/k8s-infra/hack/generated/pkg/armclient"
 	"github.com/Azure/k8s-infra/hack/generated/pkg/testcommon"
+	ctrl "sigs.k8s.io/controller-runtime"
 )
 
 func Test_ResourceGroup_CRUD(t *testing.T) {
@@ -53,7 +55,6 @@ func Test_ResourceGroup_CRUD(t *testing.T) {
 }
 
 func Test_StorageAccount_CRUD(t *testing.T) {
-	t.Parallel()
 
 	g := NewGomegaWithT(t)
 	ctx := context.Background()
@@ -72,7 +73,7 @@ func Test_StorageAccount_CRUD(t *testing.T) {
 		ObjectMeta: testContext.MakeObjectMetaWithName(namer.GenerateName("stor")),
 		Spec: storage.StorageAccounts_Spec{
 			Location:   testContext.AzureRegion,
-			ApiVersion: "2019-04-01", // TODO: This should be removed from the storage type eventually
+			ApiVersion: "2019-04-01", // TODO [apiversion]: This should be removed from the storage type eventually
 			Owner:      testcommon.AsOwner(rg.ObjectMeta),
 			Kind:       storage.StorageAccountsSpecKindBlobStorage,
 			Sku: storage.Sku{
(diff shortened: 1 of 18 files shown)


hack/generated/controllers/generic_controller.go
```diff
@@ -681,6 +689,8 @@ func (gr *GenericReconciler) Patch(
 func (gr *GenericReconciler) isOwnerReady(ctx context.Context, data *ReconcileMetadata) (bool, error) {
 	_, err := gr.ResourceResolver.GetOwner(ctx, data.metaObj)
 	if err != nil {
+		data.log.Info("error", "Err", err)
```
Porges, line 692: Remove before merge

hack/generated/pkg/armclient/template_client.go
```diff
@@ -142,6 +143,20 @@ func AuthorizerFromEnvironment() (autorest.Authorizer, error) {
 		return nil, err
 	}
 
+	// the previous never returns an error, so we must do
+	// the checks ourselves…
+	var errs []error
+	requiredEnvVars := []string{auth.SubscriptionID, auth.ClientSecret, auth.ClientID, auth.TenantID}
```
Porges, line 149: do we need to support anything other than client-secret authentication?

hack/generated/controllers/controller_test.go
```diff
@@ -53,7 +55,6 @@ func Test_ResourceGroup_CRUD(t *testing.T) {
 }
 
 func Test_StorageAccount_CRUD(t *testing.T) {
-	t.Parallel()
```
matthchr, line 56: Removing this intentional?

matthchr: Seems strange the above test has it but this doesn't

hack/generated/controllers/controller_test.go
```diff
@@ -96,17 +97,77 @@ func Test_StorageAccount_CRUD(t *testing.T) {
...
 }
+
+func StorageAccount_BlobServices_CRUD(t *testing.T, testContext testcommon.KubePerTestContext, storageAccount metav1.ObjectMeta) {
+	ctx := context.Background()
+
+	g := NewGomegaWithT(t)
+
+	blobService := &storage.StorageAccountsBlobService{
+		ObjectMeta: ctrl.ObjectMeta{
+			Name:      "default", // MUST be 'default'
```
matthchr, line 126: how awkward

hack/generated/controllers/controller_test.go
```diff
@@ -96,17 +97,77 @@ func Test_StorageAccount_CRUD(t *testing.T) {
...
+	// Create
+	err := testContext.KubeClient.Create(ctx, blobService)
+	g.Expect(err).ToNot(HaveOccurred())
+	g.Eventually(blobService).Should(testContext.Match.BeProvisioned(ctx))
+
+	t.Run("Container CRUD", func(t *testing.T) {
+		StorageAccount_BlobServices_Container_CRUD(t, testContext, blobService.ObjectMeta)
+	})
+
+	// TODO: Delete doesn't seem to work?
```
matthchr, line 144: What's the alternative? Does deletion at the ResourceGroup level work? If it does might be worth saying that since we're not going to leak any resources at least...

hack/generated/controllers/generic_controller.go
```diff
@@ -364,27 +369,29 @@ func (gr *GenericReconciler) MonitorDelete(
...
+	// Try to create deployment:
+	data.log.Info("Starting new deployment to Azure", "action", string(action))
+	deployment, err := gr.ARMClient.CreateDeployment(ctx, toDeploy)
 
-		err = mutData.Update(deployment, nil) // Status is always nil here
-		if err != nil {
-			return err
+	if err != nil {
+		var reqErr *autorestAzure.RequestError
+		if errors.As(err, &reqErr) && reqErr.StatusCode == 409 { /* == Conflict */
```
matthchr, line 383: Could use this: https://golang.org/src/net/http/status.go

That way it reads better.

matthchr: Also is there a TODO here where we need to diff the deployment we want to create and the existing deployment and do... something?

As a matter of fact it might be worth leaving a bit of an explanation here as to how a collision can even happen and what our "desired" behavior in that case is.

hack/generated/pkg/armclient/raw_client.go
```diff
@@ -102,6 +102,9 @@ func (c *Client) PutDeployment(ctx context.Context, deployment *Deployment) (*De
 		autorest.ByUnmarshallingJSON(deployment),
 		autorest.ByClosing())
 	if err != nil {
+		// TODO: rethink how to do this
```
matthchr, line 105: super minor: Comment would be clearer if instead of "this" you said specifically what had to be re-thought. I only say this because I used to do this to myself all the time and then when I came back I'm always like... I don't see what the problem is, what did I want to fix?

matthchr: Presumably I assume the issue here is that the ID is set even if we've failed, which seems weird?

hack/generated/pkg/armclient/template_client.go
matthchr: Eventually we will need to support MSI too.

hack/generated/pkg/armclient/template_client.go
```diff
@@ -142,6 +143,20 @@ func AuthorizerFromEnvironment() (autorest.Authorizer, error) {
 		return nil, err
 	}
 
+	// the previous never returns an error, so we must do
```
matthchr, line 146: File an issue for them, or find an existing issue and link against it possibly?

matthchr (approved):

matthchr:
Consider adding samples for the two new resource types in addition to tests?

hack/generated/controllers/controller_test.go
Porges: @matthchr This is owned by a StorageAccount resource, so the lifetime would be tied to that.

hack/generated/controllers/generic_controller.go
Porges: Yes, I noted that in the PR description, but I'll make it explicit in the code.

hack/generated/controllers/controller_test.go
Porges: Was removed while testing. Restored now.

hack/generated/controllers/controller_test.go
Porges: yes!

hack/generated/pkg/armclient/raw_client.go
Porges: A few things: We need the ID in the caller, which is why I added this, but also the ID doesn't need to be read from ARM since we know how to produce it. We could add a function to the `Deployment` type instead?

hack/generated/pkg/armclient/template_client.go
Porges: @matthchr the way the API works it can never produce an error since it doesn’t know what authorizer is being produced. Really the API needs reworking…

Porges: Reported at: https://github.com/Azure/go-autorest/issues/580

hack/generated/pkg/armclient/raw_client.go
Porges: Just using `GetEntityPath` in caller.

hack/generated/controllers/controller_test.go
Porges: Expanded comment

matthchr (approved):

PR #322, opened 2020-11-10
Add install script for those not using devcontainer

Porges:
This installs the tools to `$clone/hack/tools` instead. There is a `dev.sh` script to run to get into a shell with a `PATH` pointing to the right place.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -14,44 +14,21 @@ RUN apt-get update \
 # RUN curl -fsSL https://get.docker.com | sh -
 # RUN usermod -aG docker vscode
 
-# Go dependencies
-ENV GO111MODULE=on
-RUN mkdir -p /tmp/gotools \
-    && export GOPATH=/tmp/gotools \
-    && export GOCACHE=/tmp/gotools/cache \
-    && go get \
-        # go tools for vscode are preinstalled by base image (see first comment in Dockefile)
-        # golangci-lint is provided by base image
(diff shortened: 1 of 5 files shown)


.devcontainer/Dockerfile
```diff
@@ -14,44 +14,21 @@ RUN apt-get update \
...
 # note that the base image includes oh-my-bash, we are enabling plugins here
 RUN sed -i '/^plugins=/a kubectl\ngolang' "/home/vscode/.bashrc"
 RUN sed -i '/^completions=/a kubectl\ngo\ntask' "/home/vscode/.bashrc"
-# make kubectl completions work with 'k' alias
+
+# Make kubectl completions work with 'k' alias
 RUN echo 'complete -F __start_kubectl k' >> "/home/vscode/.bashrc"
 
+# Setup go-task completions
+RUN curl -sL "https://raw.githubusercontent.com/go-task/task/v3.0.0/completion/bash/task.bash" > "$HOME/.oh-my-bash/completions/task.completion.sh"
```
Porges, line 32: Revert to home/vscode

Porges (commented):
Needs a fix before merge

.devcontainer/Dockerfile
Porges:
```suggestion
RUN curl -sL "https://raw.githubusercontent.com/go-task/task/v3.0.0/completion/bash/task.bash" > "/home/vscode/.oh-my-bash/completions/task.completion.sh"
```

babbageclunk (approved):
Looks awesome!

PR #323, opened 2020-11-11
Automatically resolve import conflicts 

theunrepentantgeek:
When there are multiple imports with conflicting names, infer a name based on the second to last part of the package path.

For example, if you had

```
imports (
    "some/package/path/microsoft.email/v20180801"
    "some/package/path/microsoft.network/v20180801"
)
```
these would conflict on the inferred name of `v20180801`.

If this happens, the prior segment of the path is sliced up and the last part taken.

E.g.

* for "some/package/path/microsoft.email/v20180801" the result is `email`
* for "some/package/path/microsoft.network/v20180801" the result is `network`

If there are still conflicts after that, they're given as warnings as before:

```
File foo:
import 'collide' of 'microsoft.email/v20180801' conflicts with 1 other import(s) of the same name
import 'collide' of 'microsoft.network/v20180801' conflicts with 1 other import(s) of the same name
```


hack/generator/pkg/astmodel/file_definition.go
```diff
@@ -147,21 +147,10 @@ func (file *FileDefinition) generateImports() *PackageImportSet {
...
-					imp.packageReference,
-					imp.PackageName(),
-					otherImp.packageReference,
-					otherImp.PackageName())
-			}
-		}
+	// Resolve any conflicts and report any that couldn't be fixed up automatically
+	err := requiredImports.ResolveConflicts()
+	if err != nil {
+		klog.Warningf("File %s: %v", file.packageReference, err)
```
Porges, line 153: This should really be an error, right? The resulting code won't compile.

Porges (approved):

hack/generator/pkg/astmodel/package_import_set.go
```diff
@@ -135,3 +193,19 @@ func ByNameInGroups(left PackageImport, right PackageImport) bool {
...
+
+func (set *PackageImportSet) alternatePackageNameFor(imp PackageImport) string {
+	if imp.HasExplicitName() {
+		// Don't override explicit names
+		return imp.name
+	}
+
+	pathBits := strings.Split(imp.packageReference.PackagePath(), "/")
+	index := len(pathBits) - 1
+	if index > 0 {
```
matthchr, line 205: May want to leave a comment here as to what you're seeking?

I think this is: "Fall back to the package name (as opposed to the package version), for example Microsoft.Whatever in /foo/bar/baz/microsoft.whatever/v20150101

hack/generator/pkg/astmodel/package_import_set.go
```diff
@@ -108,6 +113,59 @@ func (set *PackageImportSet) ApplyName(ref PackageReference, name string) {
...
+	// Check for any remaining conflicts
+	var errs []error
+	for name, imports := range set.createMapByPackageName() {
+		if len(imports) <= 1 {
+			// Only one import, so no conflict
+			continue
+		}
+
+		for _, imp := range imports {
+			err := fmt.Errorf(
```
matthchr, line 142: prefer `errors.Errorf` rather than `fmt.Errorf`?

hack/generator/pkg/astmodel/package_import_set.go
matthchr: Also, this isn't sufficient to disambiguate multiple versions of the _same_ package, is it? We might want to combine version + `microsoft.whatever` above.

hack/generator/pkg/astmodel/package_import_set_test.go
```diff
@@ -14,6 +14,10 @@ var (
 	simpleTestRef PackageReference = MakeExternalPackageReference("simple")
 	pathTestRef   PackageReference = MakeExternalPackageReference("package/path")
 
+	// Important for these two package references to have the same version so that they conflict
+	emailTestRef   PackageReference = MakeExternalPackageReference("microsoft.email/v20180801")
```
matthchr, line 18: As mentioned above you may want to expand this test to 2x `microsoft.email` versions and 2x `microsoft.network` versions and I think they'll conflict.

matthchr (approved):
Approved, but I think there's an area for improvement in the conflict resolution

hack/generator/pkg/astmodel/package_import_set.go
babbageclunk: The multiple versions would use the version as their default import name though, so that's fine - the only problem would be if there was a package that was `somethingelse.whatever/v20150101`.

hack/generator/pkg/astmodel/package_import_set_test.go
babbageclunk: Oh, that's a good point - if the versions match across the two top-level packages then you'll still get a problem. Maybe the fallback should only be applied to one of the conflicting packages? Would mean you'd need to sort them first for repeatability though.

hack/generator/pkg/astmodel/package_import_set.go
matthchr: Yeah sorry, I meant both 2 versions of A and 2 of B, where 1A and 1B collide and 2A and 2B collide.

hack/generator/pkg/astmodel/package_import_set_test.go
matthchr: I personally think `emailv20100801` is basically _always_ better to read than `v20180801`, so I sorta think we should maybe just do this all the time for `LocalPackageReference` packages at least.

Porges: @matthchr I was thinking that too

hack/generator/pkg/astmodel/file_definition.go
theunrepentantgeek: Converted to use `klog.Errorf()` as discussed in today's meeting.

hack/generator/pkg/astmodel/package_import_set.go
theunrepentantgeek: Added comments, and also a fallback for further conflict resolution, using versioned identifiers.

hack/generator/pkg/astmodel/package_import_set.go
theunrepentantgeek: Changed to use `errors.Errorf()`

hack/generator/pkg/astmodel/package_import_set_test.go
theunrepentantgeek: Added fallback to versioned imports, and code to force named imports to be used for local package references.

PR #326, opened 2020-11-12
Properly support enum Names in resources

Porges:
### Naming changes

Don't convert all resource Spec `Name` properties to `string`, and instead allow the following cases:

- An enum with one value generates no `AzureName` property in the CRD as it cannot be changed. An `AzureName()` function is provided for `ConvertToArm` to invoke, which returns the fixed value.
  This only triggers for one resource currently:
    - Web `HostingEnvironmentsMultiRolePool` which must have a name of `default`

- An enum with multiple values generates an `AzureName` property with an enum value. An `AzureName()` function is provided for `ConvertToArm` which casts the enum value to string.
  Example resources for this are:
    - Keyvault `VaultsAccessPolicies` which allows names of `add`, `remove`, and `replace`
    - API Management `ServiceIdentityProvider` which allows names of `aad`, `aadB2C`, `facebook`, `google`, `microsoft`, and `twitter`

All Spec `Name` properties are still converted to `string` for the corresponding ARM type, as we need to insert a full ARM ID including owning resource IDs, etc.

### Serialization changes

Split `ArmTransformer` interface into two parts `FromArmConverter`/`ToArmConverter` and make it so that Status types only implement `FromArmConverter`.

## Other cases

There are other cases to be solved; some services (like Storage) require a Name ending in `default` (which is actually more correct, as the Name is a full URI including parent resources). At the schema level this is enforced via a regex like `^.*/default$`. We can recognize these cases and handle them; but that will be a different PR—we don’t yet support type validation properly.


hack/generator/pkg/astmodel/armconversion/convert_from_arm_function_builder.go
```diff
@@ -134,15 +134,52 @@ func (builder *convertFromArmBuilder) namePropertyHandler(
 	toProp *astmodel.PropertyDefinition,
 	fromType *astmodel.ObjectType) []ast.Stmt {
 
-	if !toProp.Equals(GetAzureNameProperty(builder.idFactory)) || !builder.isResource {
+	if !builder.isResource || !toProp.HasName(astmodel.AzureNameProperty) {
+		return nil
+	}
+
+	if typeName, ok := toProp.PropertyType().(astmodel.TypeName); ok {
+		// we are assigning to (presumbaly) an enum-typed AzureName property, (no way to check that here)
```
Porges, line 142:
```suggestion
		// we are assigning to (presumably) an enum-typed AzureName property, (no way to check that here)
```

hack/generator/pkg/astmodel/kubernetes_resource_interface.go
```diff
@@ -18,44 +20,95 @@ const (
...
+		if !namePropType.BaseType().Equals(StringType) {
+			return nil, errors.Errorf("unable to handle non-string enum base type in Name property")
+		}
+
+		options := namePropType.Options()
+		if len(options) == 1 {
+			// if there is only one possible value,
+			// we make an AzureName function that returns it, and do not
+			// provide an AzureName property on the spec
+			//r = r.WithSpec(spec.WithoutProperty(AzureNameProperty))
```
Porges, line 71: TODO: any ideas on how best to do this? this results in bad output as really we need to update the TypeDefinition for this Spec… and I can only come up with ugly solutions

Porges:
To fix: issue with optionality of `Name` 🤔

hack/generator/pkg/astmodel/kubernetes_resource_interface.go
Porges: going with ugly solution

hack/generator/pkg/codegen/pipeline_apply_kubernetes_resource_interface.go
```diff
@@ -19,32 +20,31 @@ func applyKubernetesResourceInterface(idFactory astmodel.IdentifierFactory) Pipe
...
-					spec, ok := types[specName]
-					if !ok {
-						return nil, errors.Errorf("couldn't find resource spec %q", specName)
-					}
-
-					specObj, err := astmodel.TypeAsObjectType(spec.Type())
+					newResource, err := resource.WithKubernetesResourceInterfaceImpl(idFactory, types)
 					if err != nil {
-						return nil, errors.Wrapf(err, "resource spec %q did not contain an object", specName)
+						return nil, errors.Wrapf(err, "Couldn't implement Kubernetes resource interface for %q", typeName)
```
matthchr, line 34: minor: start error with lowercase?

hack/generator/pkg/codegen/pipeline_apply_kubernetes_resource_interface.go
```diff
@@ -19,32 +20,31 @@ func applyKubernetesResourceInterface(idFactory astmodel.IdentifierFactory) Pipe
...
 					}
 
-					iface, err := astmodel.NewKubernetesResourceInterfaceImpl(idFactory, specObj)
-					if err != nil {
-						return nil, errors.Wrapf(err, "Couldn't implement Kubernetes resource interface for %q", spec.Name())
+					// this is really very ugly; a better way?
```
matthchr, line 37: You could make `WithKubernetesResourceInterfaceImpl` return a triplet with one return being an optional updated `SpecType`. It makes it a bit more obvious from the function signature that there's a possible modification? Or if the triplet result is too ugly you could make a little struct for it.

It's honestly not _lots_ better but might be a bit better.

hack/generator/pkg/codegen/pipeline_create_arm_types.go
```diff
@@ -153,14 +177,26 @@ func iterDefs(
...
 
-			actionedDefs[specTypeName] = struct{}{}
+			for _, def := range defs {
+				// don't add if already defined.
+				// some status types are shared by multiple resources…
+				if alreadyDef, ok := newDefs[def.Name()]; !ok {
+					newDefs[def.Name()] = def
+				} else {
+					if !alreadyDef.Type().Equals(def.Type()) {
+						panic("mismatch")
```
matthchr, line 196: minor: Improve text here a bit?

matthchr (approved):
Left some comments, but looks good

hack/generator/pkg/astmodel/kubernetes_resource_interface.go
```diff
@@ -18,44 +20,95 @@ const (
...
-		idFactory: idFactory,
-		asFunc:    ownerFunction,
+	namePropType := nameProp.PropertyType()
+	// resolve property type if it is a typename
+	namePropType, err = types.FullyResolve(namePropType)
+	if err != nil {
+		return nil, errors.Wrapf(err, "unable to resolve type of resource Name property: %s", namePropType.String())
+	}
+
+	var nameFunc asFuncType
```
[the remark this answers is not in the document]
Porges: Will defer this to another PR as I have updates coming in this area.

PR #330, opened 2020-11-19
Add !ignore_autogenerated build tag

Porges:
This appears to be officially supported by `gopls` and is also generated into the deepcopy files that `controller-gen` builds. Let’s use it as well.


theunrepentantgeek (approved):
Reluctantly, Approved.

PR #332, opened 2020-11-20
Remove StorageType in favour of using FlaggedType

theunrepentantgeek:
Removes `StorageType` and instead uses `FlaggedType` with the flag `StorageFlag` to do the required tracking.

2nd of 3 related PRs. #331 must merge first.


babbageclunk (approved):
Whee! This is my favourite kind of PR.

PR #336, opened 2020-11-22
Switch out go/ast for dst

Porges:
1. Mostly this is just replacing the `go/ast` import with `ast "github.com/dave/dst"`
2. Comments now go into a `Decs` (= decorations) field in the AST.
3. We can specify exactly where newlines are inserted so drop both the re-formatting step and the post-processing `addBlankLinesBeforeComments` function.
4. With `dst` AST nodes cannot be reused, so I had to do a bit of editing to add judicious calls to `ast.Clone` or otherwise remove used of shared AST nodes. This means going from `ast.Ident` to `string` in a bunch of places to remove the potential to accidentally share nodes.


hack/generator/pkg/astbuilder/calls.go
```diff
@@ -69,5 +64,5 @@ func InvokeQualifiedFunc(qualifier *ast.Ident, funcName *ast.Ident, arguments ..
 // <qualifier>.<funcName>(arguments...)
 // If you want to use the result of the function call as a value, use CallQualifiedFuncByName() instead
 func InvokeQualifiedFuncByName(qualifier string, funcName string, arguments ...ast.Expr) ast.Stmt {
-	return InvokeQualifiedFunc(ast.NewIdent(qualifier), ast.NewIdent(funcName), arguments...)
+	return InvokeQualifiedFunc(qualifier, funcName, arguments...)
```
theunrepentantgeek, line 67: Now that the parameters are simply being passed through without change, this method should be deleted.

theunrepentantgeek (approved):
Looks good!

Tidying up. 😁 
Tidying up

hack/generator/pkg/astbuilder/calls.go
Porges: it has gone

PR #343, opened 2020-12-01
Simplification of flagged types by removing flags in pipeline stage

theunrepentantgeek:
The simplification stage of the pipeline needs to also remove any flags that are still pending - this should have been done when flagged types were introduced in PR #331 

Also fixes a bug where `GetTypesInPackage()` needs to resolve the provided reference to a `LocalPackageReference` for consistency with the way packages are defined.

Once this is merged, the CI build for PR #333 should pass.

### DIFF
--- a/hack/generator/pkg/codegen/pipeline_simplify_definitions.go
+++ b/hack/generator/pkg/codegen/pipeline_simplify_definitions.go
@@ -51,6 +51,12 @@ func createSimplifyingVisitor() astmodel.TypeVisitor {
 		return tv.Visit(&ot, ctx)
 	}
 
+	// Unwrap flagged types, promoting the object within.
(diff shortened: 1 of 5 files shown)


hack/generator/pkg/codegen/pipeline_simplify_definitions.go
```diff
@@ -51,6 +51,12 @@ func createSimplifyingVisitor() astmodel.TypeVisitor {
 		return tv.Visit(&ot, ctx)
 	}
 
+	// Unwrap flagged types,, promoting the object within.
```
Porges, line 54:
```suggestion
	// Unwrap flagged types, promoting the object within.
```

Porges (approved):

hack/generator/pkg/codegen/pipeline_simplify_definitions.go
theunrepentantgeek: Fixed. Thanks.

babbageclunk (approved):

PR #345, opened 2020-12-02
Use Codecov

Porges:
Coveralls seems to keep having issues where only one of the reports is applied, giving results like "-77% code coverage". Let's see if Codecov is better.


babbageclunk (approved):
Wow, nice simple integration! Is it worth us committing their script rather than downloading it from them every time? I guess it's not like the script can do anything from inside an action.

Porges:
> Wow, nice simple integration! Is it worth us committing their script rather than downloading it from them every time? I guess it's not like the script can do anything from inside an action.

This is the recommended way to use it, and yes there are no creds or anything…

PR #351, opened 2020-12-11
Fix for removing tmp files

Porges:
go makes mod directories readonly so the files cannot be deleted, so make everything writable before deleting.


theunrepentantgeek (approved):
:shipit:

PR #358, opened 2020-12-15
Use a buffered writer to improve performance

theunrepentantgeek:
Improve the performance of the export files stage by buffering file writes using `buffio`.

Rudimentary/naïve benchmark:

Unbuffered - 193.7s
Buffered - 123.1s

Saved time ~70s (~35%)

(Benchmark method: one run under each condition without isolating the effect of other processes on the machine.)


babbageclunk (approved):
Very nice!

PR #362, opened 2021-01-06
ResourceGroup needs defaulting webhook

matthchr:


matthchr:
This fixes the nightly live tests which were failing

babbageclunk (approved):
Looks great! I guess this is because ResourceGroup is a special case that isn't generated, right? For all the other types we autogenerate the same webhook.

matthchr:
@babbageclunk - yea, and it only fails the live tests because the recording tests don't (currently) validate that the incoming request body is the same as it was before.

@Porges is there a way to make the recording tests check that?

Porges:
@matthchr we could but will need to check how much of the replay-tests it breaks 😉

matthchr:
@Porges Should I file an item for it? I would think it wouldn't break too many but maybe that's optimistic.

matthchr:
Filed #363

PR #365, opened 2021-01-07
Check request bodies during replay

Porges:
At the moment only URL/Method are compared which means you can break things in the request and you will still get a recorded response.


matthchr (approved):
Nice and simple, neat!

theunrepentantgeek (approved):

PR #366, opened 2021-01-11
Pipeline stage prerequisites

theunrepentantgeek:
Allow pipeline stages to declare their prerequisites so that we can catch issues caused by changes to the pipeline early, before they generate slightly dodgy CRD code.

* Adds `RequiresPrerequisiteStages()` to `PipelineStage` to allow prerequisites to be declared
* Adds `verifyPipeline()` to `CodeGenerator` to enforce them
* Adds some prerequisites to the existing pipeline

Enforcement is very simple at present, deliberately taking the simplest possible approach.

### DIFF
--- a/hack/generator/pkg/codegen/code_generator.go
+++ b/hack/generator/pkg/codegen/code_generator.go
@@ -7,6 +7,7 @@ package codegen
 
 import (
 	"context"
+	kerrors "k8s.io/apimachinery/pkg/util/errors"
 
 	"github.com/Azure/k8s-infra/hack/generator/pkg/astmodel"
 	"github.com/Azure/k8s-infra/hack/generator/pkg/config"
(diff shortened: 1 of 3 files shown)


hack/generator/pkg/codegen/code_generator.go
```diff
@@ -89,17 +95,21 @@ func createAllPipelineStages(idFactory astmodel.IdentifierFactory, configuration
...
 
 		// Strip out redundant type aliases:
 		removeTypeAliases(),
 
-		// De-pluralize resource types:
-		improveResourcePluralization(),
+		// De-pluralize resource types
+		// (Must come after type aliases are resolved)
+		improveResourcePluralization().
+			RequiresPrerequisiteStages("removeAliases"),
```
Porges, line 112: rather than coding in the names here it would probably be better to do e.g. 

```golang
removeAliases := removeTypeAliases()
// ... 
improveResourcePluralization().RequiresPrerequisiteStages(removeAliases.Name())
```

PR #367, opened 2021-01-11
Do not upload coverage for master-validation workflow and make pr-validation workflow run on pushes to master

Porges:


theunrepentantgeek (approved):

babbageclunk (approved):
Not very familiar with the workflows but I think this makes sense - we only track the coverage for tests without livetests, since we can't run those for PRs because of auth.

Porges:
@babbageclunk yes, exactly 👍

PR #368, opened 2021-01-12
Add ErroredType

Porges:
`ErroredType` lets us represent errors as associated with a particular type. This will (in future) allow us to attach fatal errors to types, and then _not_ fail the generation if those types are pruned from the output.

As part of this, if an `error` is `panic`’d from the `AsDeclarations` or `AsType` functions it is handled at file generation time; this is now done by `AllOfType` and `OneOfType`.


hack/generator/pkg/astmodel/type.go
```diff
@@ -51,3 +53,26 @@ func TypeEquals(left, right Type) bool {
...
+// CodeGenerationPanic is a panic thrown at codegen time
+// this should only be caused by bugs
+type CodeGenerationPanic struct {
+	errors []string
+}
+
+func (p CodeGenerationPanic) ToError() error {
+	var es []error
+	for _, e := range p.errors {
+		es = append(es, errors.New(e))
```
Porges, line 66: we should store errors in `CodeGenerationPanic` instead of strings, so the callstack is captured at point of panic creation

Porges: or just throw raw `error`, actually!

babbageclunk (approved):
Nice.

theunrepentantgeek (approved):

hack/generator/pkg/astmodel/errored_type.go
```diff
@@ -0,0 +1,99 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ */
+
+package astmodel
```
matthchr, line 6: You may want to do a test to ensure that this type makes it all the way through the pipeline. There are at least a few stages where wrapper types got dropped when I did this for `FlagType`.

See [here](https://github.com/matthchr/k8s-infra/blob/744a47fda37545c6e4ce6283e284f102dfc87c2c/hack/generator/pkg/codegen/pipeline_create_arm_types.go#L228) and [here](https://github.com/Azure/k8s-infra/pull/309/files#diff-ac2b16201a4dd7f4ffd0044fae998aba3391504a0a3d3d98d78ed08f2e5a1e1a) for example.

hack/generator/pkg/astmodel/gomega_extensions_test.go
```diff
@@ -0,0 +1,38 @@
...
+package astmodel
+
+import (
+	"fmt"
+
+	"github.com/onsi/gomega/format"
+	"github.com/onsi/gomega/types"
+)
+
+func MatchErrorString(value string) types.GomegaMatcher {
```
matthchr, line 15: Why doesn't this work?
https://onsi.github.io/gomega/#matcherrorexpected-interface

matthchr: It seems to support exactly this case already? Or am I missing something?

matthchr (approved):
Looks good but left a comment about the custom matcher - not sure if it's actually needed?

hack/generator/pkg/astmodel/gomega_extensions_test.go
Porges: what! I looked for this and somehow didn't find it. deleting my version ...

hack/generator/pkg/astmodel/errored_type.go
Porges: I'll deal with these cases when I start using it 🙂

PR #369, opened 2021-01-12
Move missing status info to a ErroredType warning

Porges:
Remove missing status info pipeline stage and put the warning into an `ErroredType`. This means it will be emitted if the type is not pruned. 

Various other fixes to ensure `ErroredType` works properly in the rest of the pipeline (at least for this scenario), e.g. `ErroredType` is allowed to contain a nil `Type`.


hack/generator/pkg/astmodel/errored_type.go
```diff
@@ -72,7 +100,7 @@ func (errored ErroredType) handleWarningsAndErrors() {
 		}
 
 		if len(es) == 1 {
-			panic(errored.errors[0])
+			panic(es[0])
```
Porges, line 103: missed this on last PR

hack/generator/pkg/astmodel/errored_type.go
```diff
@@ -53,11 +73,19 @@ func stringSlicesEqual(l []string, r []string) bool {
 }
 
 func (errored ErroredType) References() TypeNameSet {
-	return errored.inner.References()
+	if errored.inner != nil {
```
babbageclunk, line 76: Minor: these would read better (to me at least) with the sense flipped.

babbageclunk, line 76: Minor: these would read better (to me at least) with the sense flipped.

hack/generator/pkg/astmodel/errored_type.go
```diff
@@ -81,15 +109,27 @@ func (errored ErroredType) handleWarningsAndErrors() {
 
 func (errored ErroredType) AsDeclarations(cgc *CodeGenerationContext, dc DeclarationContext) []ast.Decl {
 	errored.handleWarningsAndErrors()
-	return errored.inner.AsDeclarations(cgc, dc)
+	if errored.inner != nil {
```
babbageclunk, line 112: I think I see one of the reasons you were wishing for interface receivers the other day! I don't think the workaround I was going to suggest of wrapping inner in a struct with nil-checking Type methods would buy much here though. :(

hack/generator/pkg/astmodel/errored_type_test.go
```diff
@@ -0,0 +1,34 @@
...
+
+import (
+	"testing"
+
+	"github.com/leanovate/gopter"
+	"github.com/leanovate/gopter/gen"
+	"github.com/leanovate/gopter/prop"
+)
+
+func TestErroredTypeProperties(t *testing.T) {
```
babbageclunk, line 16: Ooh nice - I need to find more ways to do property testing! I've mostly only used it in parsing-type scenarios.

hack/generator/pkg/codegen/pipeline_augment_status.go
```diff
@@ -70,15 +71,18 @@ func augmentResourcesWithStatus(idFactory astmodel.IdentifierFactory, config *co
...
+						newStatus = statusDef
 						found++
 					} else {
-						// missing status is caught later in pipeline (checkForMissingStatusInformation)
-						// as we only want to report this for non-pruned types
-						newTypes.Add(typeDef)
+						klog.V(4).Infof("Swagger information missing for %v", typeName)
+						// add a warning that the status is missing
+						// this will be reported if the type is not pruned
+						newStatus = astmodel.MakeErroredType(nil, nil, []string{fmt.Sprintf("missing status information for %v", typeName)})
```
babbageclunk, line 83: Why not just put the error on the resource rather than allowing ErroredType to wrap nil? Is it because we want to allow generation of the resource even if the status is missing? Although I guess that's what warnings are for. Some other reason?

babbageclunk (approved):
This is neat, some minor comments. Also I'm curious why ErroredType needs to be able to wrap nil?

hack/generator/pkg/codegen/pipeline_augment_status.go
Porges: Wrapping the ResourceType would introduce a lot more churn/changes since we pattern-match on ResourceType in many places. (part of the problem with having wrapper types instead of some kind of annotation/attribute)

theunrepentantgeek (approved):

hack/generator/pkg/codegen/pipeline_augment_status.go
babbageclunk: Ah right - makes sense, thanks.

hack/generator/pkg/astmodel/errored_type.go
Porges: @babbageclunk I actually wanted it for the `IgnoringErrors` function 🙂

PR #370, opened 2021-01-16
Make LocalPathPrefix configurable

matthchr:


hack/generator/pkg/config/type_transformer.go
```diff
@@ -48,9 +48,13 @@ type TypeTransformer struct {
 	// type. This is only usable with Property. Target and Remove are
 	// mutually exclusive.
 	Remove bool `yaml:",omitempty"`
+
+	// makeLocalPackageReferenceFunc is a function creating a local package reference
+	makeLocalPackageReferenceFunc func(group string, version string) astmodel.LocalPackageReference
 }
 
-func produceTargetType(target TransformTarget, descriptor string) (astmodel.Type, error) {
+// TODO: passing this here is a bit icky
```
matthchr, line 56: Open to a cleaner way to do this -- _could_ change the function to be non-pure so that it can just use the lambda on the `Transformer`? I don't love that solution either as it's awkward to call a function on self during the `Initialize()` pass - open to changing in that direction though if folks prefer it that way

Porges:
Ha, I have just run into wanting this. Thanks for preëmpting me 🙂

hack/generator/pkg/astmodel/local_package_reference_test.go
```diff
@@ -11,6 +11,10 @@ import (
 	. "github.com/onsi/gomega"
 )
 
+func makeTestLocalPackageReference(group string, version string) LocalPackageReference {
+	return MakeLocalPackageReference("github.com/Azure/k8s-infra/hack/generated/apis", group, version)
```
Porges, line 15:
```suggestion
	return MakeLocalPackageReference("github.com/Azure/k8s-infra/hack/generated/_apis", group, version)
```

needs updating after #372

matthchr: Fixed

Porges (approved):

PR #371, opened 2021-01-18
Remove gcov2lcov

Porges:
This is not needed with codecov.io and it is very slow when running locally.


Porges:
@codecov-io uhoh

Porges:
Ok I think that the difference is that the plain gcov information has info on "partial" lines which are branches where only one path is taken.

babbageclunk (approved):
Oh nice - does this mean we can get branch coverage as well?

Porges:
@babbageclunk a little bit, but maybe not full branch coverage, it sounds like [the go tooling doesn't support it](https://github.com/golang/go/issues/28888)

PR #372, opened 2021-01-18
Rename "apis" dir to exclude it from go tools by default

Porges:
Rename "apis" to "_apis" which makes go tooling ignore the files inside it. Without this, things like `gopls` will try to load all the generated files and die. This lets the IDE tooling be more useful, but to have e.g. controller-tools scan every file the `./...` path must be used from the `_apis` directory or it will be ignored.

Bonus: With this change we can simply use `./...` in the Taskfile where before we had to filter out the generated files specifically.


Porges:
Need to restore some files before review

babbageclunk (approved):

hack/generated/.gitignore
```diff
@@ -1,3 +1,4 @@
 apis/
```
matthchr, line 1: minor: Could remove this?

PR #373, opened 2021-01-19
Add a test for creating/deleting a CosmosDB Account

Porges:


hack/generated/controllers/crd_cosmosdb_test.go
```diff
@@ -0,0 +1,79 @@
...
+
+	expectedKind := documentdb.DatabaseAccountStatusKindGlobalDocumentDB
+	g.Expect(*acct.Status.Kind).To(Equal(expectedKind))
+
+	g.Expect(acct.Status.Id).ToNot(BeNil())
+	armId := *acct.Status.Id
+
+	// Run sub-tests
+	/*
+		t.Run("Blob Services CRUD", func(t *testing.T) {
```
babbageclunk, line 65: Just left around from another test?

hack/generated/controllers/suite_test.go
```diff
@@ -19,7 +19,7 @@ import (
 
 const (
 	TestNamespace          = "k8s-infra-test-ns"
-	DefaultResourceTimeout = 2 * time.Minute
+	DefaultResourceTimeout = 10 * time.Minute
```
babbageclunk, line 22: Is there a way to make this 10 min just for CosmosDB and keep other resources 2 min?

babbageclunk (approved):
Looks nice! Line count threw me for a moment. :)

theunrepentantgeek (approved):

hack/generated/controllers/suite_test.go
Porges: @babbageclunk we already had some timeouts on master with the live tests so perhaps better to make it 10 minutes for everything? That's about the maximum we can do, BTW, go kills the tests after 11 minutes 😆

PR #375, opened 2021-01-22
Begin to obey Retry-After

Porges:
I feel like this is rather messy. Unsure how to improve it at the moment.

The `Applier` now always returns the value of the `Retry-After` header (or a zero duration). It is up to the caller to interpret that, as `Retry-After` can be used in both success (202 Accepted) and error (non-200) situations. 

Also removed the default requeue delay we were using, now relying on the automatic backoff that controller-runtime gives us.

### DIFF
--- a/hack/generated/controllers/crd_resourcegroup_test.go
+++ b/hack/generated/controllers/crd_resourcegroup_test.go
@@ -42,7 +42,7 @@ func Test_ResourceGroup_CRUD(t *testing.T) {
 
 	// Ensure that the resource group was really deleted in Azure
 	// TODO: Do we want to just use an SDK here? This process is quite icky as is...
-	exists, err, _ := testContext.AzureClient.HeadResource(
+	exists, _, err := testContext.AzureClient.HeadResource(
 		ctx,
 		armId,
 		"2020-06-01")
--- a/hack/generated/controllers/crd_resourcegroup_test.go
+++ b/hack/generated/controllers/crd_resourcegroup_test.go
@@ -42,7 +42,7 @@ func Test_ResourceGroup_CRUD(t *testing.T) {
 
 	// Ensure that the resource group was really deleted in Azure
 	// TODO: Do we want to just use an SDK here? This process is quite icky as is...
-	exists, err := testContext.AzureClient.HeadResource(
+	exists, err, _ := testContext.AzureClient.HeadResource(
 		ctx,
 		armId,
 		"2020-06-01")
--- a/hack/generated/controllers/generic_controller.go
+++ b/hack/generated/controllers/generic_controller.go
@@ -443,7 +443,7 @@ func (gr *GenericReconciler) MonitorDeployment(ctx context.Context, action Recon
 			return ctrl.Result{}, errors.Wrap(err, "getting resource ID from resource")
 		}
 
-		s, statusErr, _ := gr.getStatus(ctx, resourceID, data)
+		s, _, statusErr := gr.getStatus(ctx, resourceID, data)
 		if statusErr != nil {
 			return ctrl.Result{}, errors.Wrap(statusErr, "getting status from ARM")
 		}
@@ -563,16 +563,18 @@ func (gr *GenericReconciler) constructArmResource(ctx context.Context, data *Rec
 	return resource, nil
 }
 
-func (gr *GenericReconciler) getStatus(ctx context.Context, id string, data *ReconcileMetadata) (genruntime.FromArmConverter, error, time.Duration) {
+var zeroDuration time.Duration = 0
+
+func (gr *GenericReconciler) getStatus(ctx context.Context, id string, data *ReconcileMetadata) (genruntime.FromArmConverter, time.Duration, error) {
 	deployableSpec, err := reflecthelpers.ConvertResourceToDeployableResource(ctx, gr.ResourceResolver, data.metaObj)
 	if err != nil {
-		return nil, err, 0
+		return nil, zeroDuration, err
 	}
 
 	// TODO: do we tolerate not exists here?
 	armStatus, err := reflecthelpers.NewEmptyArmResourceStatus(data.metaObj)
 	if err != nil {
-		return nil, errors.Wrapf(err, "constructing ARM status for resource: %q", id), 0
+		return nil, zeroDuration, errors.Wrapf(err, "constructing ARM status for resource: %q", id)
 	}
 
 	// Get the resource
@@ -580,20 +582,20 @@ func (gr *GenericReconciler) getStatus(ctx context.Context, id string, data *Rec
 	if data.log.V(4).Enabled() {
 		statusBytes, marshalErr := json.Marshal(armStatus)
 		if marshalErr != nil {
-			return nil, errors.Wrapf(err, "serializing ARM status to JSON for debugging"), 0
+			return nil, zeroDuration, errors.Wrapf(err, "serializing ARM status to JSON for debugging")
 		}
 
 		data.log.V(4).Info("Got ARM status", "status", string(statusBytes))
 	}
 
 	if err != nil {
-		return nil, errors.Wrapf(err, "getting resource with ID: %q", id), retryAfter
+		return nil, retryAfter, errors.Wrapf(err, "getting resource with ID: %q", id)
 	}
 
 	// Convert the ARM shape to the Kube shape
 	status, err := reflecthelpers.NewEmptyStatus(data.metaObj)
 	if err != nil {
-		return nil, errors.Wrapf(err, "constructing Kube status object for resource: %q", id), 0
+		return nil, zeroDuration, errors.Wrapf(err, "constructing Kube status object for resource: %q", id)
 	}
 
 	owner := data.metaObj.Owner()
@@ -608,10 +610,10 @@ func (gr *GenericReconciler) getStatus(ctx context.Context, id string, data *Rec
 	// TODO: The owner parameter here should be optional
 	err = status.PopulateFromArm(knownOwner, reflecthelpers.ValueOfPtr(armStatus)) // TODO: PopulateFromArm expects a value... ick
 	if err != nil {
-		return nil, errors.Wrapf(err, "converting ARM status to Kubernetes status"), 0
+		return nil, zeroDuration, errors.Wrapf(err, "converting ARM status to Kubernetes status")
 	}
 
-	return status, nil, 0
+	return status, 0, nil
 }
 
 func (gr *GenericReconciler) resourceSpecToDeployment(ctx context.Context, data *ReconcileMetadata) (*armclient.Deployment, error) {
--- a/hack/generated/controllers/generic_controller.go
+++ b/hack/generated/controllers/generic_controller.go
@@ -350,7 +350,7 @@ func (gr *GenericReconciler) MonitorDelete(
 	found, err, retryAfter := gr.ARMClient.HeadResource(ctx, resource.GetId(), resource.Spec().GetApiVersion())
 	if err != nil {
 		if retryAfter != 0 {
-			data.log.V(3).Info("Error performing HEAD on resource, will retry", "delay", retryAfter/time.Second)
+			data.log.V(3).Info("Error performing HEAD on resource, will retry", "delaySec", retryAfter/time.Second)
 			return ctrl.Result{RequeueAfter: retryAfter}, nil
 		}
 
@@ -424,7 +424,7 @@ func (gr *GenericReconciler) MonitorDeployment(ctx context.Context, action Recon
 	deployment, err, retryAfter := gr.ARMClient.GetDeployment(ctx, deployment.Id)
 	if err != nil {
 		if retryAfter != 0 {
-			data.log.V(3).Info("Error performing GET on deployment, will retry", "delay", retryAfter/time.Second)
+			data.log.V(3).Info("Error performing GET on deployment, will retry", "delaySec", retryAfter/time.Second)
 			return ctrl.Result{RequeueAfter: retryAfter}, nil
 		}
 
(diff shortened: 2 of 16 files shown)


hack/generated/controllers/generic_controller.go
```diff
@@ -100,10 +101,6 @@ func (options *Options) setDefaults() {
 		options.RequeueDelay = 5 * time.Second
 	}
 
-	if options.RequeueDelayFast == 0 {
```
Porges, line 103: Removed this “fast delay”: we can simply set `Requeue: true`

hack/generated/controllers/generic_controller.go
```diff
@@ -530,21 +550,17 @@ func (gr *GenericReconciler) ManageOwnership(ctx context.Context, action Reconci
 			return ctrl.Result{}, errors.Wrap(err, "patching resource error")
 		}
 
-		return ctrl.Result{
-			// TODO: We should consider a scaling backoff here, see: https://github.com/Azure/k8s-infra/issues/263
```
Porges, line 534: Removed this comment: controller-runtime already implements a scaling backoff, see: https://stuartleeks.com/posts/error-back-off-with-controller-runtime/

hack/generated/controllers/generic_controller.go
```diff
@@ -333,9 +332,13 @@ func (gr *GenericReconciler) StartDeleteOfResource(
 	}
 
 	// delete has started, check back to seen when the finalizer can be removed
-	return ctrl.Result{
-		RequeueAfter: gr.RequeueDelay,
-	}, nil
+	requeueDelay := gr.RequeueDelay
+	if retryAfter > requeueDelay {
+		data.log.Info("DELETE started, will check again", "delay", retryAfter/time.Second)
```
Porges, line 337: TODO should probably remove these logs

hack/generated/controllers/generic_controller.go
```diff
@@ -355,8 +358,13 @@ func (gr *GenericReconciler) MonitorDelete(
 	}
 
 	// already deleting, just check to see if it still exists and if it's gone, remove finalizer
-	found, err := gr.ARMClient.HeadResource(ctx, resource.GetId(), resource.Spec().GetApiVersion())
+	found, err, retryAfter := gr.ARMClient.HeadResource(ctx, resource.GetId(), resource.Spec().GetApiVersion())
 	if err != nil {
+		if retryAfter != 0 {
+			data.log.Info("Error HEADing resource, will retry", "delay", retryAfter/time.Second)
```
Porges, line 364: need to log error as we are hiding it?

hack/generated/controllers/generic_controller.go
```diff
@@ -426,32 +433,38 @@ func (gr *GenericReconciler) MonitorDeployment(ctx context.Context, action Recon
 		return ctrl.Result{}, err
 	}
 
+	deployment, err, retryAfter := gr.ARMClient.GetDeployment(ctx, deployment.Id)
+	if err != nil {
+		if retryAfter != 0 {
+			data.log.Info("Error GETting deployment, will retry", "delay", retryAfter/time.Second)
```
Porges, line 439: need to log error as we are hiding it?

theunrepentantgeek (approved):
I don't think it looks very messy. Not sure how to do it better, TBH.

hack/generated/controllers/generic_controller.go
matthchr: Or at least make their log level really high

hack/generated/controllers/generic_controller.go
matthchr: Error performing HEAD on resource, will retry? (HEADing reads sorta weird?)

matthchr, line 364: `delaySec` rather than `delay` to clarify the units?

hack/generated/controllers/generic_controller.go
matthchr, line 337: `delaySec` rather than `delay` to clarify the units if you keep this?

hack/generated/controllers/generic_controller.go
```diff
@@ -312,16 +308,19 @@ func (gr *GenericReconciler) StartDeleteOfResource(
...
+	var retryAfter time.Duration // ARM can tell us how long to wait for a DELETE
+
 	err = gr.Patch(ctx, data, func(ctx context.Context, mutData *ReconcileMetadata) error {
 		emptyStatus, err := reflecthelpers.NewEmptyArmResourceStatus(mutData.metaObj)
 		if err != nil {
 			return errors.Wrapf(err, "creating empty status for %q", resource.GetId())
 		}
 
-		err = gr.ARMClient.BeginDeleteResource(ctx, resource.GetId(), resource.Spec().GetApiVersion(), emptyStatus)
+		err, retryAfter = gr.ARMClient.BeginDeleteResource(ctx, resource.GetId(), resource.Spec().GetApiVersion(), emptyStatus)
```
matthchr, line 319: Do we care to be able to distinguish "there was no retryAfter header" from "there was a retryAfter header and it was zero"?

I'm thinking it'd be sorta easy to call this and think that the service was saying to retry immediately when really it just didn't give any guidance.

hack/generated/controllers/generic_controller.go
matthchr: Same comments as above -- `performing GET deployment` and `delaySec`

hack/generated/controllers/generic_controller.go
matthchr: Or maybe we're ok with saying that even if the service gave `0` as their retryAfter we just wouldn't honor it.

hack/generated/controllers/generic_controller.go
```diff
@@ -498,14 +514,18 @@ func (gr *GenericReconciler) MonitorDeployment(ctx context.Context, action Recon
...
-		result = ctrl.Result{
-			RequeueAfter: gr.RequeueDelay,
-		}
+	if deployment.IsTerminalProvisioningState() {
+		// we are done
+		return ctrl.Result{}, nil
 	}
-	return result, err
+	// need to check back; still creating resources or deleting the deployment itself
+	requeueDelay := gr.RequeueDelay
```
matthchr, line 522: Wondering if we want a helper method that just does this and returns the actual delay? It could have common logging (a few of these if's have had logs) and it saves us missing a case if we ever decide to change the logic?

hack/generated/controllers/generic_controller.go
```diff
@@ -562,36 +578,37 @@ func (gr *GenericReconciler) constructArmResource(ctx context.Context, data *Rec
 	return resource, nil
 }
 
-func (gr *GenericReconciler) getStatus(ctx context.Context, id string, data *ReconcileMetadata) (genruntime.FromArmConverter, error) {
+func (gr *GenericReconciler) getStatus(ctx context.Context, id string, data *ReconcileMetadata) (genruntime.FromArmConverter, error, time.Duration) {
 	deployableSpec, err := reflecthelpers.ConvertResourceToDeployableResource(ctx, gr.ResourceResolver, data.metaObj)
 	if err != nil {
-		return nil, err
+		return nil, err, 0
```
matthchr, line 584: minor: Might be clearer to use `time.Duration(0)` here? Not sure if that's idiomatic or not though

matthchr: Same thoughts below

hack/generated/controllers/generic_controller.go
```diff
@@ -562,36 +578,37 @@ func (gr *GenericReconciler) constructArmResource(ctx context.Context, data *Rec
...
 	// TODO: do we tolerate not exists here?
 	armStatus, err := reflecthelpers.NewEmptyArmResourceStatus(data.metaObj)
 	if err != nil {
-		return nil, errors.Wrapf(err, "constructing ARM status for resource: %q", id)
+		return nil, errors.Wrapf(err, "constructing ARM status for resource: %q", id), 0
 	}
 
 	// Get the resource
-	err = gr.ARMClient.GetResource(ctx, id, deployableSpec.Spec().GetApiVersion(), armStatus)
+	err, retryAfter := gr.ARMClient.GetResource(ctx, id, deployableSpec.Spec().GetApiVersion(), armStatus)
```
matthchr, line 594: wait, shouldn't err be the last result of the response tuple?

hack/generated/pkg/armclient/raw_client.go
```diff
@@ -110,60 +111,72 @@ func (c *Client) PutDeployment(ctx context.Context, deployment *Deployment) erro
...
-func (c *Client) GetResource(ctx context.Context, resourceID string, resource interface{}) error {
+func (c *Client) GetResource(ctx context.Context, resourceID string, resource interface{}) (error, time.Duration) {
 
 	preparer := autorest.CreatePreparer(
 		autorest.AsContentType("application/json"))
 
 	req, err := c.newRequest(ctx, http.MethodGet, resourceID)
 	if err != nil {
-		return err
+		return err, 0
```
matthchr, line 121: Same question about `time.Duration(0)` vs just `0`

hack/generated/pkg/armclient/template_client.go
```diff
@@ -24,18 +24,18 @@ import (
...
 	NewSubscriptionDeployment(location string, deploymentName string, resourceSpec genruntime.ArmResourceSpec) *Deployment
 
 	SubscriptionID() string
 
 	// TODO: These functions take an empty status and fill it out with the response from Azure (rather than as
 	// TODO: the return type. I don't love that pattern but don't have a better one either.
-	BeginDeleteResource(ctx context.Context, id string, apiVersion string, status genruntime.ArmResourceStatus) error
-	GetResource(ctx context.Context, id string, apiVersion string, status genruntime.ArmResourceStatus) error
-	HeadResource(ctx context.Context, id string, apiVersion string) (bool, error)
+	BeginDeleteResource(ctx context.Context, id string, apiVersion string, status genruntime.ArmResourceStatus) (error, time.Duration)
```
matthchr, line 36: Same comment on return order -- `error` should be last I think.

hack/generated/pkg/armclient/raw_client.go
```diff
@@ -110,60 +111,72 @@ func (c *Client) PutDeployment(ctx context.Context, deployment *Deployment) erro
...
+		}
+	}
+
+	return 0
 }
 
 // DeleteResource will make an HTTP DELETE call to the resourceId and attempt to fill the resource with the response.
 // If the body of the response is empty, the resource will be nil.
-func (c *Client) DeleteResource(ctx context.Context, resourceID string, resource interface{}) error {
+func (c *Client) DeleteResource(ctx context.Context, resourceID string, resource interface{}) (error, time.Duration) {
```
matthchr, line 167: Returns are backwards? (error should be last?)

matthchr (commented):
Had some minor comments

hack/generated/pkg/armclient/raw_client.go
matthchr: I repeated this a few times, sorry =)

hack/generated/controllers/generic_controller.go
Porges: @matthchr that was my thought 🙂

hack/generated/controllers/generic_controller.go
Porges: @matthchr I’m going to be updating this to remove our requeue delay in most (all?) places, given that (per discussion), controller-runtime does its own backoff if we simply specify `Requeue: true`

matthchr: Ah ok - that makes sense.

You doing that in this PR or a separate one?

hack/generated/controllers/generic_controller.go
matthchr: Fine by me

hack/generated/controllers/generic_controller.go
Porges: Will update this PR.

hack/generated/pkg/armclient/raw_client.go
Porges: I might define a constant…

hack/generated/controllers/generic_controller.go
Porges: Moved to log level 3, since at 4 we dump JSON 🙂

hack/generated/controllers/crd_resourcegroup_test.go
```diff
@@ -42,7 +42,7 @@ func Test_ResourceGroup_CRUD(t *testing.T) {
 
 	// Ensure that the resource group was really deleted in Azure
 	// TODO: Do we want to just use an SDK here? This process is quite icky as is...
-	exists, err := testContext.AzureClient.HeadResource(
+	exists, err, _ := testContext.AzureClient.HeadResource(
```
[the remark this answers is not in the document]
Porges: We need to have retries in the tests in general, I think. But we should address that in a holistic manner 🙂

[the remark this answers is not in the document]
Porges: Added: https://github.com/Azure/k8s-infra/issues/381

hack/generated/controllers/generic_controller.go
Porges: done

hack/generated/pkg/armclient/raw_client.go
```diff
@@ -110,60 +111,72 @@ func (c *Client) PutDeployment(ctx context.Context, deployment *Deployment) erro
...
 
-	return nil
+	return nil, retryAfter
+}
+
+func getRetryAfter(resp *http.Response) time.Duration {
+	if retryAfterStr := resp.Header.Get("Retry-After"); retryAfterStr != "" {
+		if retryAfterVal, parseErr := strconv.ParseInt(retryAfterStr, 10, 64); parseErr == nil {
+			return time.Duration(retryAfterVal) * time.Second
+		}
```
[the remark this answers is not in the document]
Porges: I’m not sure that Azure ever returns this… but I’ll look into it.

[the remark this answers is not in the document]
Porges: Added!

theunrepentantgeek (approved):

PR #376, opened 2021-01-24
Performance improvements during generation

Porges:
This cuts about a minute off the runtime.

* cache identifiers so that identifier factory doesn't keep recreating the same ones
  * some parts of the code were spending most of their time building identifiers! 
* parallelize file emitting (this makes the biggest difference)
* make `AllOfType`, `ErroredType`, `OneOfType` into pointers so that Type can be `==` compared for short-circuiting, and use this in `TypeVisitor`
  * should help reduce the amount of garbage made by visitors  

Before/after profiles (GitHub won’t let me attach SVG): [profiles.zip](https://github.com/Azure/k8s-infra/files/5882743/profiles.zip)
The baseline one might be after adding the identifier cache, can’t recall — but it shows a lot of allocation/GC time being spent in maps generated by `dst`. Potentially we could contribute improvements upstream for this.

### DIFF
--- a/hack/generator/pkg/astmodel/allof_type.go
+++ b/hack/generator/pkg/astmodel/allof_type.go
@@ -29,10 +29,10 @@ type AllOfType struct {
 
 var _ Type = &AllOfType{}
 
-// MakeAllOfType is a smart constructor for AllOfType,
+// BuildAllOfType is a smart constructor for AllOfType,
 // maintaining the invariants. If only one unique type
 // is passed, then the result will be that type, not an AllOf.
-func MakeAllOfType(types ...Type) Type {
+func BuildAllOfType(types ...Type) Type {
 	uniqueTypes := MakeTypeSet()
 	for _, t := range types {
 		if allOf, ok := t.(*AllOfType); ok {
@@ -75,10 +75,10 @@ func MakeAllOfType(types ...Type) Type {
 
 		var ts []Type
 		oneOfs[0].types.ForEach(func(t Type, _ int) {
-			ts = append(ts, MakeAllOfType(append(notOneOfs, t)...))
+			ts = append(ts, BuildAllOfType(append(notOneOfs, t)...))
 		})
 
-		return MakeOneOfType(ts...)
+		return BuildOneOfType(ts...)
 	} else if len(oneOfs) > 1 {
 		// emit a warning if this ever comes up
 		// (it doesn't at the moment)
--- a/hack/generator/pkg/astmodel/allof_type.go
+++ b/hack/generator/pkg/astmodel/allof_type.go
@@ -89,7 +89,6 @@ func MakeAllOfType(types ...Type) Type {
 	return &AllOfType{uniqueTypes}
 }
 
-
 // Types returns what types the AllOf can be.
 // Exposed as ReadonlyTypeSet so caller can't break invariants.
 func (allOf *AllOfType) Types() ReadonlyTypeSet {
--- a/hack/generator/pkg/astmodel/allof_type.go
+++ b/hack/generator/pkg/astmodel/allof_type.go
@@ -27,13 +27,15 @@ type AllOfType struct {
 	types TypeSet
 }
 
+var _ Type = &AllOfType{}
+
 // MakeAllOfType is a smart constructor for AllOfType,
 // maintaining the invariants. If only one unique type
 // is passed, then the result will be that type, not an AllOf.
 func MakeAllOfType(types ...Type) Type {
 	uniqueTypes := MakeTypeSet()
 	for _, t := range types {
-		if allOf, ok := t.(AllOfType); ok {
+		if allOf, ok := t.(*AllOfType); ok {
 			allOf.types.ForEach(func(t Type, _ int) {
 				uniqueTypes.Add(t)
 			})
@@ -52,10 +54,10 @@ func MakeAllOfType(types ...Type) Type {
 	}
 
 	// see if there are any OneOfs inside
-	var oneOfs []OneOfType
+	var oneOfs []*OneOfType
 	var notOneOfs []Type
 	uniqueTypes.ForEach(func(t Type, _ int) {
-		if oneOf, ok := t.(OneOfType); ok {
+		if oneOf, ok := t.(*OneOfType); ok {
 			oneOfs = append(oneOfs, oneOf)
 		} else {
 			notOneOfs = append(notOneOfs, t)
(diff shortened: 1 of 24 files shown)


Porges:
Not sure what to do about the logging here (@bearps)

hack/generator/pkg/astmodel/errored_type.go
```diff
@@ -20,20 +20,24 @@ type ErroredType struct {
...
 		errors:   errors,
 		warnings: warnings,
-	}.WithType(t) // using WithType ensures warnings and errors get merged if needed
+	}
+
+	return result.WithType(t) // using WithType ensures warnings and errors get merged if needed
 }
 
-func (e ErroredType) InnerType() Type {
+func (e *ErroredType) InnerType() Type {
```
theunrepentantgeek, line 35: Suggest being consistent with receiver name
```suggestion
func (errored *ErroredType) InnerType() Type {
```

hack/generator/pkg/astmodel/identifier_factory.go
```diff
@@ -45,11 +56,29 @@ func NewIdentifierFactory() IdentifierFactory {
 	return &identifierFactory{
 		renames:       createRenames(),
 		reservedWords: createReservedWords(),
+		idCache:       make(idCache),
```
theunrepentantgeek, line 59: I didn't know you could pass the variable instead of the type - nice!

hack/generator/pkg/codegen/pipeline_export_generated_code.go
```diff
@@ -87,33 +88,58 @@ func writeFiles(ctx context.Context, packages map[astmodel.PackageReference]*ast
 	globalProgress := newProgressMeter()
 	groupProgress := newProgressMeter()
 
+	var wg sync.WaitGroup
```
theunrepentantgeek, line 91: How do you limit having too many of these running at once?

If there are too many, I'd expect output congestion, probably need to limit it to #CPUs or something like that.

hack/generator/pkg/astmodel/errored_type.go
```diff
@@ -20,20 +20,24 @@ type ErroredType struct {
 	warnings []string
 }
 
-func MakeErroredType(t Type, errors []string, warnings []string) ErroredType {
-	return ErroredType{
+var _ Type = &ErroredType{}
+
+func MakeErroredType(t Type, errors []string, warnings []string) *ErroredType {
```
matthchr, line 25: minor: `NewErroredType` if it returns a ptr is I think the style we've been using

hack/generator/pkg/astmodel/allof_type.go
```diff
@@ -27,13 +27,15 @@ type AllOfType struct {
 	types TypeSet
 }
 
+var _ Type = &AllOfType{}
+
 // MakeAllOfType is a smart constructor for AllOfType,
 // maintaining the invariants. If only one unique type
 // is passed, then the result will be that type, not an AllOf.
 func MakeAllOfType(types ...Type) Type {
```
matthchr, line 35: minor: `NewAllOfType` if it returns a ptr I think... although this is a bit of a special case because the return type is not actually `AllOfType` so maybe it doesn't make sense?

theunrepentantgeek (commented):
A couple of questions, but otherwise looks good.

hack/generator/pkg/astmodel/oneof_type.go
```diff
@@ -24,13 +24,15 @@ type OneOfType struct {
 	types TypeSet
 }
 
+var _ Type = &OneOfType{}
+
 // MakeOneOfType is a smart constructor for a  OneOfType,
 // maintaining the invariants. If only one unique type
 // is passed, the result will be that type, not a OneOf.
 func MakeOneOfType(types ...Type) Type {
```
matthchr, line 32: Same comment as the `AllOf` type - unsure if changing this name makes sense or not - although I do note that in other instances such as the IdFactory when we return a ptr even if the actual specified type is an interface we still say `New`

hack/generator/pkg/astmodel/type_visitor.go
```diff
@@ -28,7 +28,7 @@ type TypeVisitor struct {
 	VisitResourceType  func(this *TypeVisitor, it *ResourceType, ctx interface{}) (Type, error)
 	VisitFlaggedType   func(this *TypeVisitor, it *FlaggedType, ctx interface{}) (Type, error)
 	VisitValidatedType func(this *TypeVisitor, it ValidatedType, ctx interface{}) (Type, error)
```
matthchr, line 30: minor: Wondering if we want `ValidatedType` to be a ptr since basically everything else is?

hack/generator/pkg/astmodel/type_visitor.go
```diff
@@ -273,17 +293,36 @@ func IdentityVisitOfAllOfType(this *TypeVisitor, it AllOfType, ctx interface{})
...
 	}
 
+	if typeSlicesFastEqual(newTypes, it.types.types) {
+		return it, nil // short-circuit
+	}
+
 	return MakeAllOfType(newTypes...), nil
 }
 
+// just checks reference equality of Types
```
matthchr, line 303: Comment for why we have this since it's "weird" at first glance?

hack/generator/pkg/codegen/pipeline_export_generated_code.go
```diff
@@ -87,33 +88,58 @@ func writeFiles(ctx context.Context, packages map[astmodel.PackageReference]*ast
...
+
+	wgDone := make(chan bool)
```
matthchr, line 93: minor: Consider a comment about why parallel here - possible link to issue/issues in `dst` (if that's the primary cause of problems you noticed)

matthchr (approved):

hack/generator/pkg/astmodel/errored_type.go
matthchr: FWIW I think @babbageclunk had argued against long (> a few letters) receiver names in the past as non-idiomatic.

hack/generator/pkg/astmodel/identifier_factory.go
Porges: that is a type :)

hack/generator/pkg/astmodel/errored_type.go
theunrepentantgeek: I don't mind short or long (within reason) but all the other methods in this file are using `errored` already, hence the suggestion. Could rename them all to `e` or `et`.

hack/generator/pkg/astmodel/identifier_factory.go
theunrepentantgeek: Suddenly I don't like it so much. 😁 🤔

hack/generator/pkg/codegen/pipeline_export_generated_code.go
Porges: Limited to 8; it's going to be IO bound as well so just going by CPU count isn't super useful. 8 is a nice round number :D

hack/generator/pkg/astmodel/type_visitor.go
Porges: Yes it should be, as it will blow up in the same way with `==`. Not sure how I missed it.

hack/generator/pkg/astmodel/allof_type.go
Porges: Yeah, it's a smart constructor so doesn't always. I might rename these to `Build*`

Issue #124, opened 2020-05-27
Handle apiVersion property specially


matthchr:

matthchr:
@theunrepentantgeek - this should probably play into our thinking about how we're going to do API versioning in general so tagging you here.

matthchr:
Alternatives are something like _each_ version being store-able in etcd and not attempting to convert between them... not sure that's even possible in k8s-land.

matthchr:
This was partially resolved by #213 - the last remaining bit is how we remember/consistently use the right APIVersion.

Porges:
I’ve done this and it seems very straightforward, at least for the conversion to ARM. The remaining bit is for conversion to the storage types.

Porges:
Mostly done in #377 but needs to be considered for conversion to storage types so leaving this open as a reminder.

PR #377, opened 2021-01-25
Make APIVersion auto-inserted so user does not need to specify it

Porges:
This completes most of #124, but still need to insert the property for the storage type. Also need to double-check how we read from Azure…


matthchr (approved):
This came out nicely

Porges:
We don’t yet generate conversions for storage types, so can leave that unsolved.

PR #378, opened 2021-01-25
Generate functions for storage conversions

theunrepentantgeek:


hack/generator/pkg/astmodel/storage_conversion_function.go
```diff
@@ -0,0 +1,345 @@
...
+	}
+
+	if fn.conversionDirection == ConvertFrom {
+		return "source"
+	}
+
+	panic(fmt.Sprintf("Unexpected conversion conversionType %v", fn.conversionDirection))
+}
+
+func (fn *StorageConversionFunction) createConversions(receiver TypeDefinition) []error {
```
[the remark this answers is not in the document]
matthchr: It might be worth chatting about this in person so I understand more fully how the `StoragePropertyConversion`s are used. I think they're set up the way they are because at construction time you don't have/know the final reciever name, so you need to have them be lambdas so that you can pick/choose what the receiver name actually is? The actual conversion though (say, optional int to int) is encoded at function construction time in the `StoragePropertyConversion` is that right?

hack/generator/pkg/astmodel/storage_conversion_factories.go
```diff
@@ -0,0 +1,360 @@
...
+		return "0"
+	case UInt32Type:
+		return "0"
+	case UInt64Type:
+		return "0"
+	case BoolType:
+		return "false"
+	}
+
+	return "##DOESNOTCOMPUTE##"
```
matthchr, line 359: This feels like it should just be a panic to me? Better that than generating invalid code it feels like.

hack/generator/pkg/astmodel/storage_conversion_endpoint.go
```diff
@@ -0,0 +1,96 @@
...
+func (endpoint *StorageConversionEndpoint) WithType(theType Type) *StorageConversionEndpoint {
+	return NewStorageConversionEndpoint(
+		theType,
+		endpoint.name,
+		endpoint.knownLocals)
+}
+
+// createLocal creates a new unique local with the specified suffix
+// Has to be deterministic, so we use an incrementing number to make them unique
+func (locals KnownLocalsSet) createLocal(nameHint string) string {
```
matthchr, line 65: It feels like there's some duplication or overlap here and with `IdentifierFactory`?

Obviously `IdentifierFactory` doesn't have any of the locals-tracking stuff, but wondering if it makes sense to do some or all of the following:
1. Promote this concept of the `KnownLocalsSet` to its own file (possibly throwing in a rename to use `Identifier` somewhere in there? `KnownLocalIdentifierSet`?)
2. Take `IdentifierFactory` as an input to a `NewSet` constructor.
3. Move any functionality missing from `IdentifierFactory` there (the plural/singular bits?)

That has the added advantage (at least I think it's an advantage?) that the renames that we set up in `idFactory` apply to the locals as well? Although I guess maybe they already do because the namehint is coming from something which has already been renamed by the idFactory? Also allows this to make use of @porges caching stuff as well.

matthchr: Definitely at least the `toPrivate` bit of this is exactly what `IdentifierFactory` `CreateIdentifier` with visibility `notexported` already does.

hack/generator/pkg/astmodel/storage_conversion_factories.go
```diff
@@ -0,0 +1,360 @@
...
+	"github.com/pkg/errors"
+	"go/token"
+)
+
+// createPropertyConversion tries to create a property conversion between the two provided properties, using all of the
+// available conversion functions in priority order to do so.
+func createPropertyConversion(
+	sourceProperty *PropertyDefinition,
+	destinationProperty *PropertyDefinition,
+	knownLocals map[string]struct{}) (StoragePropertyConversion, error) {
```
matthchr, line 21:
```suggestion
	knownLocals KnownLocalsSet) (StoragePropertyConversion, error) {
```

hack/generator/pkg/astmodel/storage_conversion_factories.go
```diff
@@ -0,0 +1,360 @@
...
+		sourceProperty.propertyType, string(sourceProperty.propertyName), knownLocals)
+	destinationEndpoint := NewStorageConversionEndpoint(
+		destinationProperty.propertyType, string(destinationProperty.propertyName), knownLocals)
+
+	conversion, err := createTypeConversion(sourceEndpoint, destinationEndpoint)
+
+	if err != nil {
+		return nil, errors.Wrapf(
+			err,
+			"trying to assign %s from %s",
```
matthchr, line 33:
```suggestion
			"trying to assign %q from %q",
```

hack/generator/pkg/astmodel/storage_conversion_factories.go
```diff
@@ -0,0 +1,360 @@
...
+		return conversion(reader, writer, ctx)
+	}, nil
+}
+
+// StorageTypeConversion generates the AST for a given conversion.
+// source is an expression to read the original value
+// destination is an expression to write the converted value
+// The parameters source and destination are funcs because AST fragments can't be reused, and in
+// some cases we need to reference source and destination multiple times in a single fragment.
+type StorageTypeConversion func(reader dst.Expr, writer dst.Expr, ctx *CodeGenerationContext) []dst.Stmt
```
matthchr, line 52: minor: same comment about `ctx` being confusing in the go context when not referring to `context.Context`

hack/generator/pkg/astmodel/storage_conversion_factories.go
```diff
@@ -0,0 +1,360 @@
...
+
+		var reader = astbuilder.Selector(source, string(sourceProperty.PropertyName()))
+		var writer = astbuilder.Selector(destination, string(destinationProperty.PropertyName()))
+
+		return conversion(reader, writer, ctx)
+	}, nil
+}
+
+// StorageTypeConversion generates the AST for a given conversion.
+// source is an expression to read the original value
```
matthchr, line 48: comments for parameter names don't match actual parameter names.

I do kinda prefer `source` + `destination` more than `reader`/`writer` but see how you got to `reader`/`writer` given the ubiquity of `source`/`destination`.

hack/generator/pkg/astmodel/storage_conversion_factories.go
```diff
@@ -0,0 +1,360 @@
...
+// destination is an expression to write the converted value
+// The parameters source and destination are funcs because AST fragments can't be reused, and in
```
matthchr, line 50: They aren't funcs?

hack/generator/pkg/astmodel/storage_conversion_factories.go
```diff
@@ -0,0 +1,360 @@
...
+// <destination> = <source>
+//
+func assignPrimitiveTypeFromPrimitiveType(
+	sourceEndpoint *StorageConversionEndpoint,
+	destinationEndpoint *StorageConversionEndpoint) StorageTypeConversion {
+
+	srcOpt := AsOptionalType(sourceEndpoint.Type())
+	dstOpt := AsOptionalType(destinationEndpoint.Type())
+
+	if srcOpt != dstOpt {
```
matthchr, line 105: Should this be `.Equals`?

I do think you mentioned rewriting this function though.

hack/generator/pkg/astmodel/storage_conversion_factories.go
```diff
@@ -0,0 +1,360 @@
...
+
+	return func(reader dst.Expr, writer dst.Expr, ctx *CodeGenerationContext) []dst.Stmt {
+		return []dst.Stmt{
+			astbuilder.SimpleAssignment(writer, token.ASSIGN, astbuilder.AddrOf(reader)),
+		}
+	}
+}
+
+// assignPrimitiveTypeFromOptionalPrimitiveType will generate a direct assignment if both types
+// have the same underlying primitive type and have the same optionality
```
matthchr, line 158: Comment doesn't match what method does?

hack/generator/pkg/astmodel/storage_conversion_factories.go
```diff
@@ -0,0 +1,360 @@
...
+
+	return func(reader dst.Expr, writer dst.Expr, ctx *CodeGenerationContext) []dst.Stmt {
+		return []dst.Stmt{
+			astbuilder.SimpleAssignment(writer, token.ASSIGN, reader),
+		}
+	}
+}
+
+// assignOptionalPrimitiveTypeFromPrimitiveType will generate a direct assignment if both types
+// have the same underlying primitive type and have the same optionality
```
matthchr, line 127: Comment doesn't match what the method does?

hack/generator/pkg/astmodel/storage_conversion_factories.go
```diff
@@ -0,0 +1,360 @@
...
+	dt := AsArrayType(destinationEndpoint.Type())
+
+	if st == nil || dt == nil {
+		// One or other type is not an array
+		return nil
+	}
+
+	srcEp := sourceEndpoint.WithType(st.element)
+	dstEp := destinationEndpoint.WithType(dt.element)
+	conversion, _ := createTypeConversion(srcEp, dstEp)
```
matthchr, line 238: minor: a comment explaining that this error is ignored because the intent of this function is to return nil if no conversion matches, and `createTypeConversion` returns `err` in that case might be worth it?

matthchr (commented):
Some stylistic comments and questions but looks good!

hack/generator/pkg/astmodel/storage_conversion_function.go
```diff
@@ -0,0 +1,345 @@
...
+	"github.com/dave/dst"
+	"sort"
+)
+
+// StoragePropertyConversion generates the AST for a given conversion.
+// source is an expression for the source value that will be read.
+// destination is an expression the target value that will be written.
+type StoragePropertyConversion func(source dst.Expr, destination dst.Expr, ctx *CodeGenerationContext) []dst.Stmt
+
+// Represents a function that performs conversions for storage versions
```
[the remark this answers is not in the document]
theunrepentantgeek: Fixed - let me know if I missed any.

hack/generator/pkg/astmodel/storage_conversion_function.go
```diff
@@ -0,0 +1,345 @@
...
+package astmodel
+
+import (
+	"fmt"
+	"github.com/Azure/k8s-infra/hack/generator/pkg/astbuilder"
+	"github.com/dave/dst"
+	"sort"
+)
+
+// StoragePropertyConversion generates the AST for a given conversion.
```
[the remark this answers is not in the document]
theunrepentantgeek: Rephrased to:
> StoragePropertyConversion represents a function that generates the correct AST to convert a single property value
> Different functions will be used, depending on the types of the properties to be converted.

hack/generator/pkg/astmodel/storage_conversion_function.go
```diff
@@ -0,0 +1,345 @@
...
+)
+
+var _ Function = &StorageConversionFunction{}
+
+func NewStorageConversionFromFunction(
+	receiver TypeDefinition,
+	source TypeName,
+	staging TypeDefinition,
+	idFactory IdentifierFactory,
+) (*StorageConversionFunction, []error) {
```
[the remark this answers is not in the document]
theunrepentantgeek: Agreed. I'd already considered changing that in my followup branch, will do so here.

hack/generator/pkg/astmodel/storage_conversion_function.go
```diff
@@ -0,0 +1,345 @@
...
+		conversionDirection: ConvertFrom,
+		conversions:         make(map[string]StoragePropertyConversion),
+		knownLocals:         make(KnownLocalsSet),
+	}
+
+	errs := result.createConversions(receiver)
+	return result, errs
+}
+
+func NewStorageConversionToFunction(
```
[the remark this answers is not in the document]
theunrepentantgeek: Fixed.

hack/generator/pkg/astmodel/storage_conversion_factories.go
theunrepentantgeek: Oh, they were - before I changed the `astbuilder` code to take care of the cloning. Fixed.

hack/generator/pkg/astmodel/storage_conversion_factories.go
theunrepentantgeek: Fixed

hack/generator/pkg/astmodel/storage_conversion_factories.go
theunrepentantgeek: Fixed.

hack/generator/pkg/astmodel/storage_conversion_factories.go
theunrepentantgeek: Added at the start of the block

hack/generator/pkg/astmodel/storage_conversion_function.go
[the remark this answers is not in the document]
matthchr: Yeah that I think helps clarify

hack/generator/pkg/astmodel/storage_conversion_function.go
```diff
@@ -0,0 +1,345 @@
...
+// source is an expression for the source value that will be read.
+// destination is an expression the target value that will be written.
+type StoragePropertyConversion func(source dst.Expr, destination dst.Expr, ctx *CodeGenerationContext) []dst.Stmt
+
+// Represents a function that performs conversions for storage versions
+type StorageConversionFunction struct {
+	// Name of this conversion function
+	name string
+	// Name of the ultimate hub type to which we are converting, passed as a parameter
+	parameter TypeName
```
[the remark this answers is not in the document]
matthchr: FWIW the `staging` throws me a bit more than `parameter` does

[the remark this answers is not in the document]
theunrepentantgeek: As discussed offline, have made three changes:
* renamed `parameter` to `hubType` 
* renamed `staging` to `intermediateType` with associated type change to `*TypeDefinition`
* Leave `intermediateType` as **nil** when directly converting

hack/generator/pkg/astmodel/storage_conversion_function.go
```diff
@@ -0,0 +1,345 @@
...
+}
+
+func (fn *StorageConversionFunction) RequiredPackageReferences() *PackageReferenceSet {
+	return NewPackageReferenceSet(
+		fn.parameter.PackageReference,
+		fn.staging.name.PackageReference)
+}
+
+func (fn *StorageConversionFunction) References() TypeNameSet {
+	return NewTypeNameSet(fn.parameter, fn.staging.name)
```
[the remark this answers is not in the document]
theunrepentantgeek: We've tried to consistently use `.Name()` elsewhere, I'll update these usages to match.

hack/generator/pkg/astmodel/storage_conversion_function.go
```diff
@@ -0,0 +1,345 @@
...
+}
+
+func (fn *StorageConversionFunction) Equals(f Function) bool {
+	if other, ok := f.(*StorageConversionFunction); ok {
+		// Only check name for now
```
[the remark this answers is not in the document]
theunrepentantgeek: It wasn't, but I've expanded `.Equals()` anyway

hack/generator/pkg/astmodel/storage_conversion_function.go
```diff
@@ -0,0 +1,345 @@
...
+// parameter is an expression for access to our parameter passed to the function, also used for field access
+// ctx is our code generation context, passed to allow resolving of identifiers in other packages
+func (fn *StorageConversionFunction) generateBody(receiver string, parameter string, ctx *CodeGenerationContext) []dst.Stmt {
+
+	if fn.parameter.Equals(fn.staging.name) {
+		// Last step of conversion, directly to the parameter type we've been given
+		if fn.conversionDirection == ConvertFrom {
+			return fn.generateDirectConversionFrom(receiver, parameter, ctx)
+		} else {
+			// fn.conversionType == ConvertTo
```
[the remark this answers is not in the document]
theunrepentantgeek: Wasn't a comment, was instead documenting which branch this was. Have converted to a **switch** for clarity, including other places.

hack/generator/pkg/astmodel/storage_conversion_function.go
```diff
@@ -0,0 +1,345 @@
...
+			return fn.generateDirectConversionTo(receiver, parameter, ctx)
+		}
+	}
+
+	// Intermediate step of conversion, not working directly with the parameter type we've been given
```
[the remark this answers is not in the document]
theunrepentantgeek: Have expanded; along with other changes in this method, I think it's now clear. Please verify!

hack/generator/pkg/astmodel/storage_conversion_function.go
```diff
@@ -0,0 +1,345 @@
...
+	sort.Slice(properties, func(i, j int) bool {
+		return properties[i] < properties[j]
+	})
+
+	// Accumulate all the statements required for conversions, in alphabetical order
+	for _, prop := range properties {
+		conversion := fn.conversions[prop]
+		block := conversion(source, destination, ctx)
+		if len(block) > 0 {
+			//TODO: Tidy
```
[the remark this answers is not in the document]
theunrepentantgeek: This was a lot messier earlier; removed the comment.

hack/generator/pkg/astmodel/storage_conversion_function.go
```diff
@@ -0,0 +1,345 @@
...
+	result = append(result, callConvertTo)
+	return result
+}
+
+func (fn *StorageConversionFunction) generateAssignments(source dst.Expr, destination dst.Expr, ctx *CodeGenerationContext) []dst.Stmt {
+	var result []dst.Stmt
+
+	// Find all the properties for which we have a conversion
+	var properties []string
+	for p := range fn.conversions {
```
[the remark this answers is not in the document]
theunrepentantgeek: Yes - we generate errors for any properties left out.

hack/generator/pkg/astmodel/storage_conversion_function.go
```diff
@@ -0,0 +1,345 @@
...
+
+	// Sort the properties into alphabetical order to ensure deterministic generation
+	sort.Slice(properties, func(i, j int) bool {
+		return properties[i] < properties[j]
+	})
+
+	// Accumulate all the statements required for conversions, in alphabetical order
+	for _, prop := range properties {
+		conversion := fn.conversions[prop]
+		block := conversion(source, destination, ctx)
```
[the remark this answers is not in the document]
theunrepentantgeek: Renaming to `generationContext` throughout.

hack/generator/pkg/astmodel/storage_conversion_function.go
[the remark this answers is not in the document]
theunrepentantgeek: There **is** an assumption that the list of properties on both the receiver and other types won't change from the point where the analysis is done. 

I did it this way partly to keep the amount of work done in the code-generation phase as low as possible, but you do raise a possible risk.

Do you think we need to enforce this by doing a cross check at the codegen stage?

hack/generator/pkg/astmodel/storage_conversion_function.go
```diff
@@ -0,0 +1,345 @@
...
+			// fn.conversionType == ConvertTo
+			return fn.generateDirectConversionTo(receiver, parameter, ctx)
+		}
+	}
+
+	// Intermediate step of conversion, not working directly with the parameter type we've been given
+	if fn.conversionDirection == ConvertFrom {
+		return fn.generateIndirectConversionFrom(receiver, parameter, ctx)
+	} else {
+		// fn.conversionType == ConvertTo
```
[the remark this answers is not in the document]
theunrepentantgeek: Not commented, documented - converted to a switch for clarity.

hack/generator/pkg/astmodel/storage_conversion_function.go
```diff
@@ -0,0 +1,345 @@
...
+		if ok {
+			var conv StoragePropertyConversion
+			var err error
+			if fn.conversionDirection == ConvertFrom {
+				conv, err = createPropertyConversion(otherProperty, receiverProperty, fn.knownLocals)
+			} else {
+				conv, err = createPropertyConversion(receiverProperty, otherProperty, fn.knownLocals)
+			}
+
+			if conv != nil {
```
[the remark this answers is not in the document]
theunrepentantgeek: I've partially adopted your suggestion, but want to keep the test for `conv != nil` for the case where there are properties we ignore.

hack/generator/pkg/astmodel/storage_conversion_factories.go
```diff
@@ -0,0 +1,360 @@
...
+	sourceEndpoint *StorageConversionEndpoint,
+	destinationEndpoint *StorageConversionEndpoint) (StorageTypeConversion, error) {
+	for _, f := range typeConversionFactories {
+		result := f(sourceEndpoint, destinationEndpoint)
+		if result != nil {
+			return result, nil
+		}
+	}
+
+	err := fmt.Errorf(
```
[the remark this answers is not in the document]
theunrepentantgeek: Fixed.

hack/generator/pkg/astmodel/storage_conversion_factories.go
theunrepentantgeek: Agreed.

hack/generator/pkg/astmodel/storage_conversion_endpoint.go
theunrepentantgeek: I promoted `KnownLocalsSet` from an alias to a full type, added an `IdentifierFactory` as a member and created a factory method `NewKnownLocalsSet()`. Nice and tidy.

hack/generator/pkg/astmodel/storage_conversion_factories.go
theunrepentantgeek: No longer needed - I changed this into a method on `StorageConversionFunction`, which removes debris from the parameter list making the code more declarative.

hack/generator/pkg/astmodel/storage_conversion_factories.go
theunrepentantgeek: Fixed.

hack/generator/pkg/astmodel/storage_conversion_function.go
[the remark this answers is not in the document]
matthchr: Ok - that's what I was uncertain about with my question: "is `conv == nil` and `err == nil` a valid situation, and it sounds like you're saying it is in the case we're bypassing a property, in which case the code you had also made sense.

hack/generator/pkg/astmodel/storage_conversion_endpoint.go
matthchr: Cool, I'll make a note when this merges to go see if I can find what I did in the ARM functions and maybe update to use `KnownLocalsSet`, as there's certainly a similar problem there and we could probably benefit from reuse.

hack/generator/pkg/astmodel/storage_conversion_endpoint.go
```diff
@@ -0,0 +1,100 @@
...
+	name string,
+	knownLocals *KnownLocalsSet) *StorageConversionEndpoint {
+	return &StorageConversionEndpoint{
+		theType:     theType,
+		name:        strings.ToLower(name),
+		knownLocals: knownLocals,
+	}
+}
+
+// Expr returns a clone of the expression for this endpoint
```
matthchr, line 35: minor: Says `Expr` but function is called `Type`.

Also the comment doesn't seem to match the method

hack/generator/pkg/astmodel/storage_conversion_endpoint.go
```diff
@@ -0,0 +1,100 @@
...
+// (returning a clone avoids issues with reuse of fragments within the dst)
+func (endpoint *StorageConversionEndpoint) Type() Type {
+	return endpoint.theType
+}
+
+// Create an identifier for a local variable that represents a single item
```
matthchr, line 41: minor: should start with function name

hack/generator/pkg/astmodel/storage_conversion_factories.go
```diff
@@ -0,0 +1,361 @@
...
+import (
+	"fmt"
+	"github.com/Azure/k8s-infra/hack/generator/pkg/astbuilder"
+	"github.com/dave/dst"
+	"github.com/pkg/errors"
+	"go/token"
+)
+
+// createPropertyConversion tries to create a property conversion between the two provided properties, using all of the
+// available conversion functions in priority order to do so.
```
matthchr, line 17:
```suggestion
// available conversion functions in priority order to do so. If no valid conversion could be created an error is returned.
```

hack/generator/pkg/astmodel/storage_conversion_factories.go
```diff
@@ -0,0 +1,361 @@
...
+	destinationEndpoint *StorageConversionEndpoint) (StorageTypeConversion, error) {
+	for _, f := range typeConversionFactories {
+		result := f(sourceEndpoint, destinationEndpoint)
+		if result != nil {
+			return result, nil
+		}
+	}
+
+	err := errors.Errorf(
+		"no conversion found to assign %s from %s",
```
matthchr, line 84:
```suggestion
		"no conversion found to assign %q from %q",
```

hack/generator/pkg/astmodel/storage_conversion_factories.go
```diff
@@ -0,0 +1,361 @@
...
+		// Keys are different types
+		return nil
+	}
+
+	srcEp := sourceEndpoint.WithType(st.value)
+	dstEp := destinationEndpoint.WithType(dt.value)
+	conversion, _ := createTypeConversion(srcEp, dstEp)
+
+	if conversion == nil {
+		// No conversion between the elements of the map
```
matthchr, line 308:
```suggestion
		// No conversion between the elements of the map, so we can't do the conversion
```

hack/generator/pkg/astmodel/storage_conversion_function.go
```diff
@@ -0,0 +1,414 @@
...
+// StoragePropertyConversion represents a function that generates the correct AST to convert a single property value
+// Different functions will be used, depending on the types of the properties to be converted.
+// source is an expression for the source value that will be read.
+// destination is an expression the target value that will be written.
+type StoragePropertyConversion func(
+	source dst.Expr, destination dst.Expr, generationContext *CodeGenerationContext) []dst.Stmt
+
+// StorageConversionFunction represents a function that performs conversions for storage versions
+type StorageConversionFunction struct {
+	// Name of this conversion function
```
matthchr, line 25: minor: I think these are supposed to start with the name of the property too although not 100% sure here if the rule applies to non-exported fields as well as exported fields.

really minor though so feel free to ignore this if you like.

hack/generator/pkg/astmodel/storage_conversion_function.go
```diff
@@ -0,0 +1,414 @@
...
+			// Different count of conversions means not-equal
+			return false
+		}
+
+		for name := range fn.conversions {
+			if _, found := other.conversions[name]; !found {
+				// Missing conversion means not-equal
+				return false
+			}
+		}
```
matthchr, line 142: I think you're missing a `return true` here - without it this function never returns true

hack/generator/pkg/astmodel/storage_conversion_function.go
```diff
@@ -0,0 +1,414 @@
...
+}
+
+// generateDirectConversionFrom returns the method body required to directly copy information from
+// the parameter instance onto our receiver
+func (fn *StorageConversionFunction) generateDirectConversionFrom(
+	receiver string,
+	parameter string,
+	generationContext *CodeGenerationContext,
+) []dst.Stmt {
+	receiverIdent := dst.NewIdent(receiver)
```
matthchr, line 226: really minor: You could possibly just change this whole function to a oneliner:

```
	return fn.generateAssignments(dst.NewIdent(parameter), dst.NewIdent(receiver), generationContext)
```

matthchr: Same with the below `generateDirectConversionTo`

hack/generator/pkg/astmodel/storage_conversion_factories.go
```diff
@@ -0,0 +1,361 @@
...
+	"fmt"
+	"github.com/Azure/k8s-infra/hack/generator/pkg/astbuilder"
+	"github.com/dave/dst"
+	"github.com/pkg/errors"
+	"go/token"
+)
+
+// createPropertyConversion tries to create a property conversion between the two provided properties, using all of the
+// available conversion functions in priority order to do so.
+func (fn StorageConversionFunction) createPropertyConversion(
```
matthchr, line 18: This isn't a ptr receiver but others are.

matthchr, line 18: It feels weird to me that this function is on `StorageConversionFunction` but it's defined in a different file than `StorageConversionFunction` is? Maybe the result of a refactoring at some point in the past?

Consider moving this alongside the other functions on `StorageConversionFunction`

hack/generator/pkg/astmodel/storage_conversion_function.go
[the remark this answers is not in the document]
matthchr: Ah yes, I see that.

hack/generator/pkg/astmodel/testdata/NastyTest-Direct.golden
```diff
@@ -0,0 +1,52 @@
+// Code generated by k8s-infra. DO NOT EDIT.
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+package vCurrent
+
+import "github.com/Azure/k8s-infra/hack/generated/_apis/Verification/vHub"
+
+type Person struct {
+	nasty map[string][]map[string]bool `json:"my-oh-my"`
```
matthchr, line 9: minor: json name mismatching `nasty` is a little weird (allowed but abnormal?)

matthchr: Applies to other files but not repeating comment there

hack/generator/pkg/astmodel/testdata/NastyTest-ViaStaging.golden
```diff
@@ -0,0 +1,65 @@
...
+import (
+	"github.com/Azure/k8s-infra/hack/generated/_apis/Verification/vHub"
+	"github.com/Azure/k8s-infra/hack/generated/_apis/Verification/vNext"
+)
+
+type Person struct {
+	nasty map[string][]map[string]bool `json:"my-oh-my"`
+}
+
+func (person Person) ConvertFrom(source *vHub.Person) error {
```
matthchr, line 15: Should these generated functions have a comment?

matthchr (approved):

Porges (approved):

hack/generator/pkg/astmodel/storage_conversion_factories.go
theunrepentantgeek: I've made this change manually.

hack/generator/pkg/astmodel/testdata/SetOptionalStringFromOptionalString-Direct.golden
```diff
@@ -0,0 +1,24 @@
...
+
+type Person struct {
+	name *string `json:"name"`
+}
+
+// ConvertFrom populates our Person from the provided source Person
+func (person Person) ConvertFrom(source *vHub.Person) error {
+
+	// name
+	person.name = source.name
```
babbageclunk, line 16: This leaves the source and destination linked together, doesn't it? After this point the code could assign to `*source.name` and the value of `*person.name` would also be updated. Should there be a local variable to prevent this?

hack/generator/pkg/astmodel/storage_conversion_factories.go
```diff
@@ -0,0 +1,331 @@
...
+// <destination> = <source>
+//
+func assignPrimitiveTypeFromPrimitiveType(
+	sourceEndpoint *StorageConversionEndpoint,
+	destinationEndpoint *StorageConversionEndpoint) StorageTypeConversion {
+
+	srcOpt := AsOptionalType(sourceEndpoint.Type())
+	dstOpt := AsOptionalType(destinationEndpoint.Type())
+
+	if srcOpt != dstOpt {
```
babbageclunk, line 73: Shouldn't this be `if (srcOpt == nil) != (dstOpt == nil)`? Oh, I guess not since you're going to punt if they're different primitive types anyway further down. Maybe update the comment though?

babbageclunk (commented):
This looks good, but I think optional handling needs to be extended to avoid source and destination being linked in the case of primitive types.

At the moment there isn't any handling of conversions where a field changes primitive type between versions (eg float to int) - how is that going to be done? Will it require a user-provided conversion function (I guess it's always going to be lossy in one direction or the other)?

hack/generator/pkg/astmodel/storage_conversion_endpoint.go
```diff
@@ -0,0 +1,102 @@
...
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ */
+
+package astmodel
+
+import (
+	"github.com/gobuffalo/flect"
+	"strconv"
+	"strings"
```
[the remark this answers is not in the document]
theunrepentantgeek: Fixed.

hack/generator/pkg/astmodel/storage_conversion_endpont_test.go
```diff
@@ -0,0 +1,28 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ */
+
+package astmodel
+
+import (
+	. "github.com/onsi/gomega"
+	"testing"
```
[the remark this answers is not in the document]
theunrepentantgeek: Fixed. Other source files in this package too, courtesy of `goimports`.

hack/generator/pkg/astmodel/storage_conversion_endpoint.go
```diff
@@ -0,0 +1,102 @@
...
+// CreateSingularLocal creates an identifier for a local variable that represents a single item
+// Each call will return a unique identifier
+func (endpoint *StorageConversionEndpoint) CreateSingularLocal() string {
+	singular := flect.Singularize(endpoint.name)
+	return endpoint.knownLocals.createLocal(singular)
+}
+
+// CreatePluralLocal creates an identifier for a local variable that represents multiple items
+// Each call will return a unique identifier
+func (endpoint *StorageConversionEndpoint) CreatePluralLocal(suffix string) string {
```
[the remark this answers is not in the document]
theunrepentantgeek: I originally only pluralized, but the variables weren't distinct enough when reading, hence the suffix. It's important that the results don't collide with `CreateSingularLocal()`, but the suffix takes care of that. I'll switch to using `flect.Singularize()`.

hack/generator/pkg/astmodel/storage_conversion_factories.go
```diff
@@ -0,0 +1,331 @@
...
+ */
+
+package astmodel
+
+import (
+	"fmt"
+	"github.com/Azure/k8s-infra/hack/generator/pkg/astbuilder"
+	"github.com/dave/dst"
+	"github.com/pkg/errors"
+	"go/token"
```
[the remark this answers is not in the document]
theunrepentantgeek: Fixed.

hack/generator/pkg/astmodel/testdata/SetStringFromString-Direct.golden
```diff
@@ -0,0 +1,24 @@
...
+package vCurrent
+
+import "github.com/Azure/k8s-infra/hack/generated/_apis/Verification/vHub"
+
+type Person struct {
+	name string `json:"name"`
+}
+
+// ConvertFrom populates our Person from the provided source Person
+func (person Person) ConvertFrom(source *vHub.Person) error {
```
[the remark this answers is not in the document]
theunrepentantgeek: Done.

hack/generator/pkg/astmodel/testdata/SetStringFromString-ViaStaging.golden
```diff
@@ -0,0 +1,37 @@
...
+	name string `json:"name"`
+}
+
+// ConvertFrom populates our Person from the provided source Person
+func (person Person) ConvertFrom(source *vHub.Person) error {
+	// personTemp is our intermediate for conversion
+	var personTemp vNext.Person
+
+	// Populate personTemp from source
+	personTemp.ConvertFrom(source)
```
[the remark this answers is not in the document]
theunrepentantgeek: I was going to come back to these - but have now propagated the errors properly.

hack/generator/pkg/astmodel/testdata/SetStringFromString-ViaStaging.golden
```diff
@@ -0,0 +1,37 @@
...
+// ConvertTo populates the provided destination Person from our Person
+func (person Person) ConvertTo(destination *vHub.Person) error {
+	// personTemp is our intermediate for conversion
+	var personTemp vNext.Person
+
+	// name
+	personTemp.name = person.name
+
+	// Populate destination from personTemp
+	personTemp.ConvertTo(destination)
```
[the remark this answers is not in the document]
theunrepentantgeek: Fixed.

hack/generator/pkg/astmodel/testdata/SetOptionalStringFromString-Direct.golden
```diff
@@ -0,0 +1,28 @@
...
+
+type Person struct {
+	name *string `json:"name"`
+}
+
+// ConvertFrom populates our Person from the provided source Person
+func (person Person) ConvertFrom(source *vHub.Person) error {
+
+	// name
+	person.name = &source.name
```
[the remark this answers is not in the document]
theunrepentantgeek: Don't want to run the risk of leaving debris in the destination object, so I want to definitively assign _something_ every time. A zero value seems the best choice.

hack/generator/pkg/astmodel/testdata/SetOptionalStringFromOptionalString-Direct.golden
theunrepentantgeek: Agreed. I'm splitting neither-optional from both-optional in a later PR, so I'll put the fix there.

hack/generator/pkg/astmodel/storage_conversion_factories.go
theunrepentantgeek: I restructure this in my next PR and it's much clearer.

PR #384, opened 2021-02-02
Improve header check

matthchr:
It was very slow before. Rewrote in Python for speed (which isn't something you hear every day)


scripts/check_headers.py
```diff
@@ -0,0 +1,61 @@
...
+
+# Returns true if the file passes validation, false if it fails
+def is_file_compliant(regex, path):
+    if not path.endswith('.go'):
+        return True
+
+    f = open(path, 'r')
+    combined = ''
+    for i in range(number_of_lines):
+        combined += f.readline()
```
babbageclunk, line 32: I'm only suggesting this because seeing string += makes me twitchy (although I realise it's just 4 lines), and also I really miss Python sometimes.
```suggestion
    combined = ''.join(itertools.islice(f, number_of_lines))
```

theunrepentantgeek (approved):
Looks good.

babbageclunk (approved):
Looks good to me, some super minor nitpicky comments.

scripts/check_headers.py
```diff
@@ -0,0 +1,61 @@
...
+    global reset
+    print('{}{}{}'.format(header, text, reset))
+
+
+# Returns true if the file passes validation, false if it fails
+def is_file_compliant(regex, path):
+    if not path.endswith('.go'):
+        return True
+
+    f = open(path, 'r')
```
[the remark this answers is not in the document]
matthchr: Bah this is what I get for doing it outside of pycharm! My Python is rusty

scripts/check_headers.py
matthchr: Good suggestion

matthchr:
Thanks for the review @babbageclunk - great comments!

PR #388, opened 2021-02-04
Add new FileWriter abstraction

matthchr:
Reduces duplication as we support more and more "file formats" such as test files, etc.

This is a stepping stone to adding a new file type for registering the CRD types with the controller.

### DIFF
--- a/hack/generator/pkg/astmodel/package_definition.go
+++ b/hack/generator/pkg/astmodel/package_definition.go
@@ -30,12 +30,12 @@ func NewPackageDefinition(groupName string, packageName string, genVersion strin
 	return &PackageDefinition{groupName, packageName, genVersion, make(Types)}
 }
 
-func (pkgDef *PackageDefinition) Definitions() Types {
-	return pkgDef.definitions
+func (p *PackageDefinition) Definitions() Types {
+	return p.definitions
 }
 
-func (pkgDef *PackageDefinition) GetDefinition(typeName TypeName) (TypeDefinition, error) {
-	for _, def := range pkgDef.definitions {
+func (p *PackageDefinition) GetDefinition(typeName TypeName) (TypeDefinition, error) {
+	for _, def := range p.definitions {
 		if def.Name().Equals(typeName) {
 			return def, nil
 		}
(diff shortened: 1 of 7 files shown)


hack/generator/pkg/astmodel/file_writer.go
```diff
@@ -0,0 +1,72 @@
...
+	"bufio"
+	"io"
+	"os"
+
+	"github.com/dave/dst"
+	"github.com/dave/dst/decorator"
+)
+
+// FileAst represents a file of Go code
+type FileAst interface { // TODO: Is there a better name? Something about CodeGeneratedFile?
```
theunrepentantgeek, line 18: What about `GoSourceFile` since that's what is generated?

hack/generator/pkg/astmodel/file_writer.go
```diff
@@ -0,0 +1,72 @@
...
+	// AsAst transforms the file into a dst.File (you can think of this as the AST representation of the file)
+	AsAst() (*dst.File, error)
+}
+
+// This is just a grouping mechanism currently -- we could
+// possibly have some configuration here but we don't right now
+type FileWriter struct{}
```
theunrepentantgeek, line 25: If you made this `GoSourceFileWriter` with a single member like this:
```
type GoSourceFileWriter struct {
    file GoSourceFile
}
```
and a factory method like this:
``` go
func NewGoSourceFileWriter(file GoSourceFile) { ... }
```
then you could avoid always having to pass the file into the methods.

hack/generator/pkg/astmodel/package_definition.go
```diff
@@ -23,19 +23,20 @@ type PackageDefinition struct {
 	PackageName      string
 	GeneratorVersion string
 	definitions      Types
+	fileWriter       *FileWriter
 }
 
 // NewPackageDefinition constructs a new package definition
 func NewPackageDefinition(groupName string, packageName string, genVersion string) *PackageDefinition {
-	return &PackageDefinition{groupName, packageName, genVersion, make(Types)}
+	return &PackageDefinition{groupName, packageName, genVersion, make(Types), NewFileWriter()}
```
theunrepentantgeek, line 31: We almost want a way for each file type to return the right kind of file writer. Maybe?

This looks as though you're assuming a package may only contain AST files - is that right? What if we want to create documentation or YAML files that go along as part of the package?

matthchr: I think that in this case the package is the go construct, so it seems reasonable to think that the contents of the package are valid go code?

I'm not sure if reality bears that out in my head things such as YAML files, etc aren't part of the package really - they might be in the folder but they're not part of the logical code-package?

hack/generator/pkg/astmodel/file_writer.go
matthchr: I was originally thinking that the writer would be usable for multiples files and might hold some configuration state around things like if it should `go fmt` the files before it wrote them or whatnot. In the abscence of any actual need for that though, I'll move to the model you suggested for now and we can always refactor back to this if it turns out we have some cross-cutting configuration we want to store.

hack/generator/pkg/astmodel/file_writer.go
```diff
@@ -0,0 +1,74 @@
...
+	"bufio"
+	"io"
+	"os"
+
+	"github.com/dave/dst"
+	"github.com/dave/dst/decorator"
+)
+
+// GoSourceFile represents a file of Go code
+type GoSourceFile interface {
```
theunrepentantgeek, line 18: Should the file be called `go_source_file.go`? #nitpick

theunrepentantgeek (approved):

hack/generator/pkg/astmodel/file_writer.go
theunrepentantgeek: Sounds good.

hack/generator/pkg/astmodel/package_definition.go
theunrepentantgeek: Heh. Different definitions. To me, the package is the directory - any files in the directory that aren't Go code had better still be related to what the package does; if not, they should be moved.

hack/generator/pkg/astmodel/file_writer.go
matthchr: Fixed

hack/generator/pkg/astmodel/package_definition.go
matthchr: Yeah I see how you got there. I'm not actually sure the technical meaning of `package` in the go context - is it just the code in the folder or everything? For example if I import an package that has some YAML files in it do the YAML files come along with the import or just the .go files?

In any case though I think in our context today the `PackageDefinition` is purely dealing with Go files and isn't capable of emitting non-go files (since `definitions` is the "set of things to generate" and it is just a collection of `Types`). If/when we add support for other types of generations we can revisit this and include a type->writer mapping?

PR #391, opened 2021-02-08
Improve Type Unwrapping 

theunrepentantgeek:
Expanding on the set of utility unwrapping functions to cover additional cases

* Changes the signature of all the methods to better fit the Go idiom of returning a bool to indicate whether the conversion works. 
* Removes IsObjectType() from `object_type.go` in favour of `AsObjectType()`
* Simplifies implementation of `extractEmbeddedTypeName()` in `object_type.go` by using `AsTypeName()`

Background: When adding the conversion `AsTypeName()`, I realised that it couldn't return a **nil** value to indicate failure; the obvious solution was to use the usual Go idiom and return `(TypeName, bool)` - but then that method would have been inconsistent with the existing methods. Solution: change all of them!

### DIFF
--- a/hack/generator/pkg/astmodel/meta_type.go
+++ b/hack/generator/pkg/astmodel/meta_type.go
@@ -77,3 +77,16 @@ func AsOptionalType(aType Type) *OptionalType {
 
 	return nil
 }
+
+// AsEnumType unwraps any wrappers around the provided type and returns either the underlying EnumType and true, or nil and false.
+func AsEnumType(aType Type) (*EnumType, bool) {
+	if enm, ok := aType.(*EnumType); ok {
+		return enm, true
+	}
+
+	if wrapper, ok := aType.(MetaType); ok {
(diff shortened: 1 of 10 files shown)


hack/generator/pkg/astmodel/meta_type.go
```diff
@@ -12,68 +12,97 @@ type MetaType interface {
...
-// PrimitiveType or nil
-func AsPrimitiveType(aType Type) *PrimitiveType {
+// PrimitiveType and true, or nil and false.
+func AsPrimitiveType(aType Type) (*PrimitiveType, bool) {
 	if primitive, ok := aType.(*PrimitiveType); ok {
-		return primitive
+		return primitive, true
 	}
 
 	if wrapper, ok := aType.(MetaType); ok {
```
babbageclunk, line 21: Ooh, I hadn't seen the MetaType interface, I like it!

babbageclunk (approved):
Nice!

PR #392, opened 2021-02-08
Add support for storage conversion of enumeration properties

theunrepentantgeek:
Extends the framework already in place to handle copying enumeration values across. Handles all the usual cases of properties being required/optional.

Includes a significant refactoring of the testing, to allow additional types to be includes as required. This was driven by the way that Enumerations are specified by `TypeName`, which needs to be resolved to the actual underlying type.

Introduces a `StorageconversionContext` as a carrier of additional information that is only needed by some conversions.

Also fixes some bugs to do with pointers to shared values potentially opening us up to aliasing issues

~Note: PR #391 must be merged first.~ PR #391 has been merged


hack/generator/pkg/astmodel/storage_conversion_context.go
```diff
@@ -0,0 +1,44 @@
...
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ */
+
+package astmodel
+
+// StorageConversionContext captures additional supporting information that may be needed when a
+// storage conversion factory creates a conversion
+type StorageConversionContext struct {
+	// types is a map of all known type definitions, used to resolve TypeNames to actual types
```
Porges, line 11:
```suggestion
	// Types is a map of all known type definitions, used to resolve TypeNames to actual types
```

Porges (approved):

hack/generator/pkg/astmodel/storage_conversion_factories.go
```diff
@@ -287,31 +309,42 @@ func assignArrayFromArray(
...
 		return nil
 	}
 
 	return func(reader dst.Expr, writer dst.Expr, generationContext *CodeGenerationContext) []dst.Stmt {
-		itemId := sourceEndpoint.CreateSingularLocal()
-		indexId := itemId + "Index"
-		tempId := sourceEndpoint.CreatePluralLocal("List")
+		// We create three obviously related identifiers to use for the conversion
+		id := sourceEndpoint.CreateLocal()
+		itemId := id + "Item"
```
babbageclunk, line 322: Should these be added to the known locals to avoid subsequent conversions from clobbering them?

hack/generator/pkg/astmodel/storage_conversion_function.go
```diff
@@ -73,10 +76,15 @@ func NewStorageConversionFromFunction(
 		conversionDirection: ConvertFrom,
 		conversions:         make(map[string]StoragePropertyConversion),
 		knownLocals:         NewKnownLocalsSet(idFactory),
+		conversionContext:   conversionContext,
 	}
 
 	err := result.createConversions(receiver)
-	return result, err
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to create '%s()'", result.name)
```
babbageclunk, line 84: In general the wrapping message should just provide the local context, rather than saying things like "failed" or "error" - that will be clear from the rest of the error message, and if there are multiple levels of wrapping then the "failed/error" messages get very noisy.
```suggestion
		return nil, errors.Wrapf(err, "creating '%s()'", result.name)
```

hack/generator/pkg/astmodel/storage_conversion_function.go
```diff
@@ -94,10 +103,15 @@ func NewStorageConversionToFunction(
 		conversionDirection: ConvertTo,
 		conversions:         make(map[string]StoragePropertyConversion),
 		knownLocals:         NewKnownLocalsSet(idFactory),
+		conversionContext:   conversionContext,
 	}
 
 	err := result.createConversions(receiver)
-	return result, err
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to create '%s()'", result.name)
```
babbageclunk, line 111: Same here.

hack/generator/pkg/astmodel/testdata/SetArrayOfOptionalFromArrayOfRequired-Direct.golden
```diff
@@ -13,12 +13,14 @@ type Person struct {
 func (person Person) ConvertFrom(source *vHub.Person) error {
 
 	// scores
-	scoreList := make([]*int, len(source.scores))
-	for scoreIndex, score := range source.scores {
-		scoreList[scoreIndex] = &score
+	score1List := make([]*int, len(source.scores))
+	for score1Index, score1Item := range source.scores {
+		score1Item := score1Item
+		score := score1Item
```
babbageclunk, line 19: This is a bit nasty - is there any way we can detect that we've already made a local and just use that? Or is that difficult because of how the code is organised?

hack/generator/pkg/astmodel/testdata/SetOptionalStringFromOptionalString-Direct.golden
```diff
@@ -13,7 +13,8 @@ type Person struct {
 func (person Person) ConvertFrom(source *vHub.Person) error {
 
 	// name
-	person.name = source.name
+	name := *source.name
```
babbageclunk, line 16: This will panic if `source.name` is nil - wrap it in an if? (Same for all of the optional-optional conversions I think, it's just clearest here.)

hack/generator/pkg/astmodel/testdata/SetStringFromOptionalString-Direct.golden
```diff
@@ -27,7 +27,8 @@ func (person Person) ConvertFrom(source *vHub.Person) error {
 func (person Person) ConvertTo(destination *vHub.Person) error {
 
 	// name
-	destination.name = &person.name
+	name := person.name
+	destination.name = &name
```
babbageclunk, line 31: If the value of `person.name` is "", should this explicitly set `destination.name` to nil?

hack/generator/pkg/astmodel/types.go
```diff
@@ -40,9 +40,9 @@ func (types Types) FullyResolve(t Type) (Type, error) {
 	return t, nil
 }
 
-// AddAll adds multiple types to the set, with the same safety check as Add() to panic if a duplicate is included
-func (types Types) AddAll(otherTypes []TypeDefinition) {
-	for _, t := range otherTypes {
+// AddAll adds multiple definitions to the set, with the same safety check as Add() to panic if a duplicate is included
+func (types Types) AddAll(otherDefinitions []TypeDefinition) {
```
babbageclunk, line 44: Nice

babbageclunk (changes requested):
This looks mostly great, but the optional-optional code generated has a nil-reference error at the moment.

hack/generator/pkg/astmodel/testdata/SetOptionalStringFromOptionalString-Direct.golden
theunrepentantgeek: Good catch! Thanks.

hack/generator/pkg/astmodel/testdata/SetStringFromOptionalString-Direct.golden
theunrepentantgeek: I'm not sure that gains us anything, we'll always have a mismatch because the non-optional field can't handle all the values. We'll end up with one of these two scenarios:

| Scenario | `*string` | `string` | `*string` |
| -- | -- | -- | -- | 
| A | **nil** | `""` | `""` |
| A | `""` | `""` | `""` |
| B | **nil** | `""` | **nil** |
| B | `""` | `""` | **nil** |

Either `""` always goes to **nil** (and we lose any original `""`) or it always goes to `""`" (and we lose any original **nil**)

babbageclunk: Yeah, that makes sense - it's inherently lossy essentially.

babbageclunk (approved):
Looks awesome, thanks!

PR #393, opened 2021-02-09
Improve conversions involving properties of simple types

theunrepentantgeek:
This is a follow-up to PR #378 to address comments raised by @babbageclunk. Refer to that PR to see his comments.

Most of the changes show up in the golden test files, so this should be quick to review.

### DIFF
--- a/hack/generator/pkg/astmodel/package_definition.go
+++ b/hack/generator/pkg/astmodel/package_definition.go
@@ -9,9 +9,10 @@ import (
 	"bytes"
 	"fmt"
 	"io/ioutil"
-	kerrors "k8s.io/apimachinery/pkg/util/errors"
 	"path/filepath"
 	"text/template"
+
+	kerrors "k8s.io/apimachinery/pkg/util/errors"
 
 	"github.com/pkg/errors"
 	"k8s.io/klog/v2"
--- a/hack/generator/pkg/astmodel/storage_conversion_function_test.go
+++ b/hack/generator/pkg/astmodel/storage_conversion_function_test.go
@@ -7,9 +7,10 @@ package astmodel
 
 import (
 	"bytes"
-	"github.com/sebdah/goldie/v2"
 	"strings"
 	"testing"
+
+	"github.com/sebdah/goldie/v2"
 
 	. "github.com/onsi/gomega"
 )
--- a/hack/generator/pkg/astmodel/testdata/SetIntFromInt-ViaStaging.golden
+++ b/hack/generator/pkg/astmodel/testdata/SetIntFromInt-ViaStaging.golden
@@ -18,7 +18,10 @@ func (person Person) ConvertFrom(source *vHub.Person) error {
 	var personTemp vNext.Person
 
 	// Populate personTemp from source
-	personTemp.ConvertFrom(source)
+	err := personTemp.ConvertFrom(source)
+	if err != nil {
+		return errors.Wrap(err, "for person, calling personTemp.ConvertFrom(source)")
+	}
 
 	// age
 	person.age = personTemp.age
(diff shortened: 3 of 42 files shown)


hack/generator/pkg/astmodel/package_definition.go
```diff
@@ -9,10 +9,11 @@ import (
 	"bytes"
 	"fmt"
 	"io/ioutil"
-	kerrors "k8s.io/apimachinery/pkg/util/errors"
 	"path/filepath"
 	"text/template"
 
+	kerrors "k8s.io/apimachinery/pkg/util/errors"
+
```
babbageclunk, line 16: You shouldn't have a break between `kerrors` and `github.com/pkg/errors`. Generally the pattern is to have (up to) 3 import stanzas: stdlib, external packages, this-module packages.

hack/generator/pkg/astmodel/storage_conversion_function_test.go
```diff
@@ -7,10 +7,11 @@ package astmodel
 
 import (
 	"bytes"
-	"github.com/sebdah/goldie/v2"
 	"strings"
 	"testing"
 
+	"github.com/sebdah/goldie/v2"
```
babbageclunk, line 13: Same here.

hack/generator/pkg/astmodel/testdata/SetIntFromInt-ViaStaging.golden
```diff
@@ -18,10 +18,16 @@ func (person Person) ConvertFrom(source *vHub.Person) error {
 	var personTemp vNext.Person
 
 	// Populate personTemp from source
-	personTemp.ConvertFrom(source)
+	err := personTemp.ConvertFrom(source)
+	if err != nil {
+		return errors.Wrap(err, "for person, calling personTemp.ConvertFrom(source)")
```
babbageclunk, line 23: Is this wrapping going to be too noisy if the chain of conversions has several links? In that case it seems like the message would be the same for each level. It might be better to have the package name in the message if possible, since that would be the thing that changes between each step in the chain, right? (Although that might be fiddly if it's not being passed down at the moment.)

hack/generator/pkg/astmodel/testdata/SetOptionalStringFromOptionalString-Direct.golden
```diff
@@ -14,11 +14,17 @@ func (person Person) ConvertFrom(source *vHub.Person) error {
 
 	// name
 	person.name = source.name
```
babbageclunk, line 16: This will still have source and destination sharing the name field - is that being fixed in a different change?

hack/generator/pkg/astmodel/type_definition_test.go
```diff
@@ -6,9 +6,10 @@
 package astmodel
 
 import (
-	"github.com/pkg/errors"
 	"testing"
 
+	"github.com/pkg/errors"
+
```
babbageclunk, line 12: Superfluous line. :)

babbageclunk (approved):
Looks good thanks! Mostly **super nitpicky** comments about imports, a couple of other things.

hack/generator/pkg/astmodel/testdata/SetOptionalStringFromOptionalString-Direct.golden
theunrepentantgeek: Yes; I have a different PR (will put it out soon) that splits neither-optional from both-optional, and will fix the shared pointer issue there.

hack/generator/pkg/astmodel/testdata/SetIntFromInt-ViaStaging.golden
theunrepentantgeek: Yes, it will be verbose, but it seems better than not wrapping. Don't currently have the package available at this point in the code (but I think will have in a later PR), so perhaps we improve it then?

PR #394, opened 2021-02-09
Add support for include-transitive

Porges:
This allows us to include a type and everything it references, meaning we can add types on an opt-in basis.

We now filter out everything except what is required to run the CI tests.


hack/generator/pkg/config/configuration.go
```diff
@@ -184,24 +184,77 @@ func (config *Configuration) initialize(configPath string) error {
...
+
+			return "", ""
+		}
+
+	default:
+		panic(errors.Errorf("unknown exportfilter directive: %s", f.Action))
+	}
+}
+
+func getAllReferencedTypes(allTypes astmodel.Types, root astmodel.TypeName, output astmodel.TypeNameSet) {
```
babbageclunk, line 230: Maybe `collectAllReferencedTypes`? A `get` function that doesn't return anything seems weird.

babbageclunk (approved):
This is cool!

hack/generator/pkg/config/configuration.go
matthchr, line 230: You possibly could make this a type w/ `allTypes` and `output` as struct variables...

```
type TransitiveTypeFinder struct {
    allTypes astmodel.Types
    applicableTypes astmodel.TypeNameSet
    filter *ExportFilter
}

func (tf *TransitiveTypeFinder) ShouldExport(typeName astmodel.TypeName) (ShouldExportResult, string) {
    if tf.applicableTypes.Contains(typeName) {
        return Export, tf.filter.Because
    }
    return "", ""
}

func (tf *TransitiveTypeFinder) Initialize() { // Or whatever name is good
    for tn := range tf.allTypes {
        if tf.filter.AppliestoType(tn) {
            tf.findReferencedTypes(tn)
        }
    }
}

func (tf *TransitiveTypeFinder) findReferencedTypes(root astmodel.TypeName) {
    tf.applicableTypes.Add(root)
    for referenced := range tf.allTypes[root].Type().References() {
        if !tf.applicableTypes.Contains(referenced) {
            tf.findReferencedTypes(referenced)
        }
    }
}
```

Then usage becomes:
```
case ExportFilterIncludeTransitive:
    typeFinder := &TransitiveTypeFinder{allTypes: allTypes, applicableTypes: make(astmodel.TypeNameSet), filter: f}
    typeFinder.Initialize()
    return typeFinder.ShouldExport // I think this works?
```

matthchr (approved):

matthchr:
Oh also you may want to take a look at the `Crossplane.yaml` file I added and see if we should make similar config file modifications to it

hack/generator/pkg/astmodel/type_name_set.go
```diff
@@ -59,12 +59,17 @@ func (ts TypeNameSet) Equals(set TypeNameSet) bool {
...
-	var result TypeNameSet
-	for val := range s1 {
-		result = result.Add(val)
+	return make(TypeNameSet).AddAll(s1).AddAll(s2)
+}
+
+func (ts TypeNameSet) AddAll(other TypeNameSet) TypeNameSet {
+	if ts == nil {
+		ts = make(TypeNameSet)
 	}
```
theunrepentantgeek, line 68: I'm not a fan of this idiom of creating the object on the fly like this. 

I've seen this in code written by C++ devs as a fertile source of bugs - consumer thinks there's a reference to an existing object there, a new one gets silently created, hijinks ensues.

Porges: However it is how the type works already. I won't change it for this PR. Go also seems to value "uninitialized objects should still work properly"…

Porges: Actually I can drop this from the PR as I didn't end up using the method.

hack/generator/pkg/config/configuration.go
Porges: I don't think this makes it much clearer, so I'll leave it as-is.

Porges:
@matthchr updated Crossplane config too

PR #395, opened 2021-02-11
Add support for conversion of complex objects

theunrepentantgeek:
Adds support for property conversions of properties with complex object types

Includes a small number of additional helper methods for generating AST fragments.

~PR #392 must be merged first~ PR #392 has been merged


hack/generator/pkg/astbuilder/assignments.go
```diff
@@ -42,3 +42,50 @@ func SimpleAssignmentWithErr(lhs dst.Expr, tok token.Token, rhs dst.Expr) *dst.A
...
+//     <trueBranch>
+// } else {
+//     <falseBranch>
+// }
+//
+func SimpleIf(condition dst.Expr, trueBranch dst.Stmt, falseBranch dst.Stmt) *dst.IfStmt {
+	result := &dst.IfStmt{
+		Cond: condition,
+		Body: StatementBlock(trueBranch),
+		Else: StatementBlock(falseBranch),
```
babbageclunk, line 58: These seem redundant since you're overwriting `Body` and `Else` below.
```suggestion
		Body: StatementBlock(trueBranch),
		Else: StatementBlock(falseBranch),
```

hack/generator/pkg/astbuilder/assignments.go
```diff
@@ -42,3 +42,50 @@ func SimpleAssignmentWithErr(lhs dst.Expr, tok token.Token, rhs dst.Expr) *dst.A
...
+	}
+
+	if block, ok := trueBranch.(*dst.BlockStmt); ok {
```
babbageclunk, line 61: How about adding an `EnsureStatementBlock` function that wraps a Stmt if it's not already a BlockStmt? Then you can just call that in the IfStmt literal.

hack/generator/pkg/astbuilder/assignments.go
babbageclunk, line 58: These seem redundant since you're overwriting `Body` and `Else` below.
```suggestion
```

hack/generator/pkg/astbuilder/lists.go
```diff
@@ -67,8 +65,6 @@ func IterateOverListWithIndex(index string, item string, list dst.Expr, statemen
 		Value: dst.NewIdent(item),
 		Tok:   token.DEFINE,
 		X:     list,
-		Body: &dst.BlockStmt{
-			List: cloneStmtSlice(statements),
-		},
+		Body:  StatementBlock(statements...),
```
babbageclunk, line 68: Quite nice that this clones the list!

hack/generator/pkg/astmodel/storage_conversion_context.go
```diff
@@ -42,3 +50,27 @@ func (c *StorageConversionContext) ResolveEnum(t Type) (TypeName, *EnumType, boo
 
 	return name, enumType, true
 }
+
+// ResolveObject takes a Type and resolves it into the name of an object type,
+// returning true if found or false if not.
+func (c *StorageConversionContext) ResolveObject(t Type) (TypeName, bool) {
```
babbageclunk, line 56: This name confused me - I expected it to take a TypeName and return the object type, but this returns the name after checking that the referent type is an object type. Maybe call it `ObjectTypeName`?

hack/generator/pkg/astmodel/storage_conversion_factories.go
```diff
@@ -79,30 +87,30 @@ func assignPrimitiveTypeFromPrimitiveType(
 	destinationEndpoint *StorageConversionEndpoint,
 	_ *StorageConversionContext) StorageTypeConversion {
 
+	// Require source to be non-optional
```
babbageclunk, line 90: These are much clearer, thanks!

hack/generator/pkg/astmodel/storage_conversion_factories.go
```diff
@@ -727,6 +702,374 @@ func assignOptionalEnumTypeFromOptionalEnumType(
...
+// }
+// <destination> = &<local>
+//
+// For ConvertTo:
+//
+// var <local> <destinationType>
+// if <source> != nil {
+//     <source>.ConvertTo(&<local>)
+// }
+// <destination> = &<local>
```
babbageclunk, line 988: Shouldn't nil be assigned to destination if source is nil? This always sets destination.

```suggestion
// 
// if <source> != nil {
//     var <local> <destinationType>
//    <local>.ConvertFrom(<source>)
//    <destination> = &<local>
// }
//
// For ConvertTo:
//
// if <source> != nil {
//     var <local> <destinationType>
//     <source>.ConvertTo(&<local>)
//     <destination> = &<local>
// }
```

hack/generator/pkg/astmodel/storage_conversion_factories.go
```diff
@@ -727,6 +702,374 @@ func assignOptionalEnumTypeFromOptionalEnumType(
...
+}
+
+// assignOptionalObjectTypeFromObjectType will generate a conversion if both properties are
+// TypeNames referencing ObjectType definitions and only the destination is optional
+//
+// For ConvertFrom:
+//
+// var <local> <destinationType>
+// <local>.ConvertFrom(<source>)
+// <destination> = &<local>
```
babbageclunk, line 895: These are missing error handling (although as far as I can see the generated code is doing it).

hack/generator/pkg/astmodel/testdata/NastyTest-Direct.golden
```diff
@@ -21,13 +21,10 @@ func (person Person) ConvertFrom(source *vHub.Person) error {
 			for nasty2Key, nasty2Value := range nasty1Item {
 				nasty2Map[nasty2Key] = nasty2Value
 			}
-
```
babbageclunk, line 24: Yeah, whitespace is for losers!
(Just kidding - I think removing these is a good thing.)

hack/generator/pkg/astmodel/testdata/SetOptionalObjectFromOptionalObject-Direct.golden
```diff
@@ -0,0 +1,48 @@
...
+	// role
+	if source.role != nil {
+		var role Release
+		err := role.ConvertFrom(source.role)
+		if err != nil {
+			return errors.Wrap(err, "populating role from role, calling ConvertFrom()")
+		}
+		person.role = &role
+	} else {
+		person.role = nil
```
babbageclunk, line 24: Ah, of course you need to explicitly assign nil here in case the destination wasn't empty.

hack/generator/pkg/astmodel/storage_conversion_factories.go
babbageclunk: (Confirmed after looking at the golden files.)

hack/generator/pkg/astmodel/storage_conversion_factories.go
babbageclunk: (Again, the golden files are right for this, it's just that the comment is misleading.)

babbageclunk (approved):
This looks great - I like the clarifying you've done of some of the comments too. A couple of doc comments for the new object conversions are wrong (probably just hadn't been updated after you updated the code).

hack/generator/pkg/astbuilder/assignments.go
theunrepentantgeek: Good catch - I've switched them to use an `EnsureStatementBlock()` method as you suggested elsewhere.

hack/generator/pkg/astbuilder/assignments.go
theunrepentantgeek: Done.

hack/generator/pkg/astmodel/storage_conversion_context.go
theunrepentantgeek: Good point. This started as a similar method to `ResolveEnum` but was returning values never used, so I simplified it, but without reconsidering the name.

Starting to suspect it's now the wrong abstraction as I'm having trouble naming it. 🤔

hack/generator/pkg/astmodel/storage_conversion_factories.go
```diff
@@ -123,41 +131,42 @@ func assignOptionalPrimitiveTypeFromPrimitiveType(
...
 	}
 
+	// Require both properties to have the same primitive type
 	if !srcPrim.Equals(dstPrim) {
-		// Not the same primitive type
 		return nil
 	}
 
-	local := destinationEndpoint.CreateLocal("Value")
+	copyvar := destinationEndpoint.CreateLocal("Value")
```
matthchr, line 163: minor: `copyVar`?

hack/generator/pkg/astbuilder/assignments.go
```diff
@@ -42,3 +42,50 @@ func SimpleAssignmentWithErr(lhs dst.Expr, tok token.Token, rhs dst.Expr) *dst.A
...
+
+// SimpleIf creates a simple if statement with a single statement in each branch
+//
+// if <condition> {
+//     <trueBranch>
+// } else {
+//     <falseBranch>
+// }
+//
+func SimpleIf(condition dst.Expr, trueBranch dst.Stmt, falseBranch dst.Stmt) *dst.IfStmt {
```
matthchr, line 54: Maybe `SimpleIfElse`? (In contrast to a `SimpleIf` which wouldn't have an `else` clause)?

hack/generator/pkg/astmodel/storage_conversion_factories.go
matthchr: Might be ok to omit the error handling here for clarity of comment, but possibly worth noting that it's been done?

```
For ConvertFrom (error handling omitted for brevity):
...
```

?

hack/generator/pkg/astmodel/storage_conversion_factories.go
```diff
@@ -727,6 +702,374 @@ func assignOptionalEnumTypeFromOptionalEnumType(
...
+		return nil
+	}
+
+	// Require destination to be an object
+	dstName, dstIsObject := conversionContext.ResolveObject(destinationEndpoint.Type())
+	if !dstIsObject {
+		return nil
+	}
+
+	copyvar := destinationEndpoint.CreateLocal()
```
matthchr, line 929: minor: `copyVar`? (Applies elsewhere as well)

matthchr (approved):
Had a few really minor comments.

LGTM!

hack/generator/pkg/astbuilder/builder.go
```diff
@@ -282,6 +282,15 @@ func ReturnNoError() dst.Stmt {
 	return result
 }
 
+// WrappedErrorf returns the err local, wrapped with additional information
```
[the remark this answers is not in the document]
theunrepentantgeek: Done.

PR #397, opened 2021-02-12
Split creation of ARM types from conversion interface

matthchr:
- Improve code clarity some.
 - Crossplane will likely make use of the first phase but likely will not make use of the second.
 - Fix a bug with how status types were handled in ARM code generation which would cause invalid code to be generated.

### DIFF
--- a/hack/generator/pkg/codegen/pipeline_create_arm_types.go
+++ b/hack/generator/pkg/codegen/pipeline_create_arm_types.go
@@ -13,256 +13,98 @@ import (
 	"k8s.io/klog/v2"
 
 	"github.com/Azure/k8s-infra/hack/generator/pkg/astmodel"
-	"github.com/Azure/k8s-infra/hack/generator/pkg/astmodel/armconversion"
 )
 
-// createArmTypesAndCleanKubernetesTypes walks the type graph and builds new types for communicating
-// with ARM, as well as removes ARM-only properties from the Kubernetes types.
-func createArmTypesAndCleanKubernetesTypes(idFactory astmodel.IdentifierFactory) PipelineStage {
+// createARMTypes walks the type graph and builds new types for communicating
+// with ARM
+func createARMTypes(idFactory astmodel.IdentifierFactory) PipelineStage {
 	return MakePipelineStage(
 		"createArmTypes",
-		"Create ARM types and remove ARM-only properties from Kubernetes types",
+		"Creates ARM types",
 		func(ctx context.Context, definitions astmodel.Types) (astmodel.Types, error) {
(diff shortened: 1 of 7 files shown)


hack/generator/pkg/codegen/pipeline_create_arm_types.go
```diff
@@ -13,275 +13,118 @@ import (
...
-
-	// TODO: a better way to do all this following would be if we had Spec/Status tags
-	// for Spec/Status types. Then we would only do one pass over all types and
-	// just examine the flags, rather than walking all resources first and
-	// the remaining definitions in a different pass… and we wouldn’t have to deal with
-	// multiple uses of the same type (see the panic below), etc, etc…
-
-	// Do all the resources first - this ensures we avoid handling a spec before we've processed
-	// its associated resource.
+// TODO: Generic method on astmodel.Types?
```
babbageclunk, line 44: They're not going to be available that soon! ;)
How do you mean generic here? Is it just a matter of moving it somewhere shared?

hack/generator/pkg/codegen/pipeline_create_arm_types.go
```diff
@@ -13,275 +13,118 @@ import (
...
-			}
-
-			modifiedDef, err := addArmConversionInterface(def, armDef, idFactory, armconversion.OrdinaryType)
-			if err != nil {
-				return astmodel.TypeDefinition{}, errors.Wrapf(err, "failed to add ARM conversion interface to %q", def.Name())
-			}
-
-			return modifiedDef, nil
-		})
+type armTypeCreator struct {
```
babbageclunk, line 39: This works out much nicer I think!

babbageclunk (approved):
Nice - the two pipeline stages read a bit cleaner after this change in my opinion.

hack/generator/pkg/codegen/pipeline_create_arm_types.go
matthchr: I just mean that this could be rephrased into a more official sounding name and moved to `astmodel.Types` as a built-in filter, as it seems to be an operation that we do quite frequently ("Find me the specs of all of the resources").

Sometimes though we need the spec and the resource (or the name of the resource), and sometimes we just need the spec, so we'd need to think about what we wanted the return type to be

matthchr: This TODO was just me attempting to ask the question: Do folks think that this operation is common enough for it to be promoted to a method someplace?

hack/generator/pkg/codegen/pipeline_create_arm_types.go
matthchr: @Porges actually suggested it a while back - agree it's significantly cleaner.

PR #398, opened 2021-02-13
Generate kustomization.yaml and controller type registration

### LINKED ISSUE #252: Autogenerate files required to register types with generic controller
Locations:
1. generated/controllers/controller_resources.go
2. generated/config/crd/kustomization.yaml 
3. generated/main.go (Scheme registration at the top)

matthchr:
Add gen-kustomize command
  - Use gen-kustomize to generate the Kustomization.yaml
     used by the controller for resource registration.

Export controller type registration
  - Allow exporting of type registration for the controller.

Closes #252


hack/generated/controllers/controller_resources.go
```diff
@@ -1,34 +1,37 @@
...
-*/
-
+// Code generated by k8s-infra. DO NOT EDIT.
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
 package controllers
 
-// TODO: Code generate this file
 import (
+	batch "github.com/Azure/k8s-infra/hack/generated/_apis/microsoft.batch/v20170901storage"
```
matthchr, line 7: Should I .gitignore this now since it's generated?

matthchr:
For reference, currently the generated kustomization.yaml in `config/crd/bases` looks like this:
```
resources:
  - bases/microsoft.batch.infra.azure.com_batchaccounts.yaml
  - bases/microsoft.batch.infra.azure.com_batchaccountsapplications.yaml
  - bases/microsoft.batch.infra.azure.com_batchaccountsapplicationsversions.yaml
  - bases/microsoft.batch.infra.azure.com_batchaccountscertificates.yaml
  - bases/microsoft.batch.infra.azure.com_batchaccountspools.yaml
  - bases/microsoft.documentdb.infra.azure.com_databaseaccounts.yaml
  - bases/microsoft.resources.infra.azure.com_resourcegroups.yaml
  - bases/microsoft.storage.infra.azure.com_storageaccounts.yaml
  - bases/microsoft.storage.infra.azure.com_storageaccountsblobservices.yaml
  - bases/microsoft.storage.infra.azure.com_storageaccountsblobservicescontainers.yaml
  - bases/microsoft.storage.infra.azure.com_storageaccountsblobservicescontainersimmutabilitypolicies.yaml
  - bases/microsoft.storage.infra.azure.com_storageaccountsmanagementpolicies.yaml

```

It's in `.gitignore` though so you don't see it in the PR.

matthchr:
I think `controller_resources_gen` is coming out not `go fmt`'ed - need to fix that before merge

matthchr:
> 
> 
> I think `controller_resources_gen` is coming out not `go fmt`'ed - need to fix that before merge

Fixed in a separate PR here: #399

hack/generated/controllers/controller_resources_gen.go
```diff
@@ -0,0 +1,33 @@
...
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+package controllers
+
+import (
+	batchv20170901 "github.com/Azure/k8s-infra/hack/generated/_apis/microsoft.batch/v20170901"
+	documentdbv20150408 "github.com/Azure/k8s-infra/hack/generated/_apis/microsoft.documentdb/v20150408"
+	storagev20190401 "github.com/Azure/k8s-infra/hack/generated/_apis/microsoft.storage/v20190401"
+	"k8s.io/apimachinery/pkg/runtime"
+	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
```
matthchr, line 11: This file is generated -- should it be .gitignored?

hack/generator/pkg/astmodel/resource_registration_file.go
```diff
@@ -0,0 +1,327 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ */
+
+// TODO: I am not sure that it makes sense to have this file in astmodel at all... maybe just in code_generator?
```
matthchr, line 6: Thoughts?

hack/generated/main.go
```diff
@@ -12,50 +12,32 @@ import (
...
-	// TODO: Need to generate this
-	_ = clientgoscheme.AddToScheme(scheme)
-	_ = batch.AddToScheme(scheme)
-	_ = storage.AddToScheme(scheme)
-	_ = resources.AddToScheme(scheme)
-	// +kubebuilder:scaffold:scheme
-}
-
 func main() {
+	flagSet := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
```
babbageclunk, line 26: Yay nice!

hack/generated/controllers/controller_resources.go
```diff
@@ -5,30 +5,23 @@ Licensed under the MIT license.
...
-
-var KnownTypes = []runtime.Object{
-	new(batch.BatchAccount),
-	// new(batch.BatchAccountsPool),
-	new(documentdb.DatabaseAccount),
-	new(resources.ResourceGroup),
-	new(storage.StorageAccount),
-	new(storage.StorageAccountsBlobService),
-	new(storage.StorageAccountsBlobServicesContainer),
+func GetKnownTypes() []runtime.Object {
```
babbageclunk, line 14: Would it be clearer to change the generators for known types and schemes to always add resources? I guess it's much of a muchness. (Not sure whether you have that expression in the US?)

hack/generator/pkg/astmodel/resource_registration_file.go
babbageclunk: Reading it I agree with the comment, this should probably be in code_generator - it's building some AST parts, but it's not really part of the model.

babbageclunk (approved):

hack/generated/controllers/controller_resources.go
matthchr: No won't don't have that phrase in the US. I looked it up though and it sound sorta similar to "6 of one, half-dozen of the other"?

I'm not actually sure what you're getting at here ("change the generators for known types and schemes to always add resources"). Can you explain more?

hack/generator/pkg/astmodel/resource_registration_file.go
matthchr: I agree - moved it

hack/generated/controllers/controller_resources.go
babbageclunk: Oh sorry! Yes, "6 of one, half-a-dozen of the other" is the same sense. I just meant, rather than generate private `getKnownTypes` and then wrap a `GetKnownTypes` around it to add the resource type, it might be better to push the code that adds it into the generator? Or might not.

hack/generated/controllers/controller_resources_gen.go
matthchr: Added to .gitignore

hack/generated/controllers/controller_resources.go
matthchr: Yeah my thinking was that it was easier to add other logic if the wrapper was handcrafted (as then it doesn't require changing the generator). So the generated code is the "building block" and then the handcrafted wrapper makes use of that building block.

Easy enough to change down the road if we determine we don't like that but I think I'll leave it for now.

PR #403, opened 2021-02-26
Code Gardening

theunrepentantgeek:
A few things that I've been collecting

* Fixing up documentation comments
* Adding/improving logging

### DIFF
--- a/hack/generator/pkg/astmodel/package_import_set.go
+++ b/hack/generator/pkg/astmodel/package_import_set.go
@@ -64,7 +64,7 @@ func (set *PackageImportSet) Remove(packageImport PackageImport) {
 	delete(set.imports, packageImport.packageReference)
 }
 
-// Contains allows checking to see if an import is included
+// ContainsImport allows checking to see if an import is included
 func (set *PackageImportSet) ContainsImport(packageImport PackageImport) bool {
 	if imp, ok := set.imports[packageImport.packageReference]; ok {
 		return imp.Equals(packageImport)
@@ -82,7 +82,7 @@ func (set *PackageImportSet) ImportFor(ref PackageReference) (PackageImport, boo
 	return PackageImport{}, false
 }
 
-// AsSlice() returns a slice containing all the imports
+// AsSlice returns a slice containing all the imports
 func (set *PackageImportSet) AsSlice() []PackageImport {
 	var result []PackageImport
 	for _, imp := range set.imports {
@@ -92,7 +92,7 @@ func (set *PackageImportSet) AsSlice() []PackageImport {
 	return result
 }
 
-// AsSortedSlice() return a sorted slice containing all the imports
+// AsSortedSlice return a sorted slice containing all the imports
 func (set *PackageImportSet) AsSortedSlice() []PackageImport {
 	result := set.AsSlice()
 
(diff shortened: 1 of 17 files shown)


hack/generator/pkg/astmodel/package_import_set.go
```diff
@@ -6,6 +6,7 @@
 package astmodel
 
 import (
+	"k8s.io/klog/v2"
```
matthchr, line 9: minor: Generally ordering is done such that non-stdlib imports are in the second section, not the first

hack/generator/pkg/codegen/pipeline_remove_type_aliases.go
```diff
@@ -49,7 +49,8 @@ func removeTypeAliases() PipelineStage {
 func resolveTypeName(visitor *astmodel.TypeVisitor, name astmodel.TypeName, types astmodel.Types) (astmodel.Type, error) {
 	def, ok := types[name]
 	if !ok {
-		panic(fmt.Sprintf("Couldn't find type for type name %s", name))
+		klog.Warningf("Couldn't find type for type name %s", name)
```
matthchr, line 52: Why isn't this a panic or error? Isn't a name that points to a type not in the collection "bad" (bound to cause failures eventually?)

matthchr (commented):
Had one question

hack/generator/pkg/codegen/pipeline_remove_type_aliases.go
matthchr: Thinking more on this, are there cases where this isn't an error because it's a reference to a non-local type? If so possibly we should specifically check for that here by examining the `name.PackageReference`?

theunrepentantgeek: Making it an error makes sense.

hack/generator/pkg/astmodel/package_import_set.go
theunrepentantgeek: Fixed.

hack/generator/pkg/codegen/pipeline_remove_type_aliases.go
matthchr: I don't think it hurts too much for this to be a `panic` (though I have no problem with it as an error either).

Applying the "can the caller handle this" litmus test, I am not sure (in the context of our application) they can, which would leave a `panic` ok? (@porges has more thoughts on this than I do though)

hack/generator/pkg/codegen/pipeline_remove_type_aliases.go
```diff
@@ -49,7 +50,7 @@ func removeTypeAliases() PipelineStage {
 func resolveTypeName(visitor *astmodel.TypeVisitor, name astmodel.TypeName, types astmodel.Types) (astmodel.Type, error) {
 	def, ok := types[name]
 	if !ok {
-		panic(fmt.Sprintf("Couldn't find type for type name %s", name))
+		return nil, errors.Errorf("Couldn't find definition for type name %s", name)
```
matthchr, line 53: minor: lowercase first letter of error because Go hates us

matthchr (approved):

hack/generator/pkg/codegen/pipeline_remove_type_aliases.go
theunrepentantgeek: Fixed.

PR #404, opened 2021-02-26
Some fixes

matthchr:
- ObjectType should consider embedded structures in References().
- panic if include-transitive doesn't match any types.
- VisitTypeDefinition shouldn't return a pointer.

### DIFF
--- a/hack/generator/pkg/astmodel/object_type.go
+++ b/hack/generator/pkg/astmodel/object_type.go
@@ -208,17 +208,13 @@ func (objectType *ObjectType) RequiredPackageReferences() *PackageReferenceSet {
 
 // References returns the set of all the types referred to by any property.
 func (objectType *ObjectType) References() TypeNameSet {
-	var results TypeNameSet
+	results := NewTypeNameSet()
 	for _, property := range objectType.properties {
-		for ref := range property.PropertyType().References() {
-			results = results.Add(ref)
-		}
+		results = results.AddAll(property.PropertyType().References())
 	}
 
 	for _, property := range objectType.embedded {
-		for ref := range property.PropertyType().References() {
-			results = results.Add(ref)
-		}
+		results = results.AddAll(property.PropertyType().References())
 	}
 
 	// Not collecting types from functions deliberately.
--- a/hack/generator/pkg/astmodel/object_type.go
+++ b/hack/generator/pkg/astmodel/object_type.go
@@ -215,6 +215,12 @@ func (objectType *ObjectType) References() TypeNameSet {
 		}
 	}
 
+	for _, property := range objectType.embedded {
+		for ref := range property.PropertyType().References() {
+			results = results.Add(ref)
+		}
+	}
+
 	// Not collecting types from functions deliberately.
 	return results
 }
--- a/hack/generator/pkg/codegen/codegen_test.go
+++ b/hack/generator/pkg/codegen/codegen_test.go
@@ -67,6 +67,14 @@ func makeTestLocalPackageReference(group string, version string) astmodel.LocalP
 	return astmodel.MakeLocalPackageReference(goModulePrefix, group, version)
 }
 
+func makeEmbeddedTestTypeDefinition() astmodel.TypeDefinition {
+	name := astmodel.MakeTypeName(makeTestLocalPackageReference("test", "v20200101"), "EmbeddedTestType")
+	t := astmodel.NewObjectType()
+	t.WithProperty(astmodel.NewPropertyDefinition("FancyProp", "fancyProp", astmodel.IntType))
+
+	return astmodel.MakeTypeDefinition(name, t)
+}
+
 func injectEmbeddedStructType() PipelineStage {
 	return MakePipelineStage(
 		"injectEmbeddedStructType",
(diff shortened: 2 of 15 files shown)


hack/generator/pkg/codegen/codegen_test.go
```diff
@@ -67,20 +67,29 @@ func makeTestLocalPackageReference(group string, version string) astmodel.LocalP
 	return astmodel.MakeLocalPackageReference(goModulePrefix, group, version)
 }
 
+func makeEmbeddedTestTypeDefinition() astmodel.TypeDefinition {
+	name := astmodel.MakeTypeName(makeTestLocalPackageReference("test", "v20200101"), "EmbeddedTestType")
+	t := astmodel.NewObjectType()
+	t.WithProperty(astmodel.NewPropertyDefinition("FancyProp", "fancyProp", astmodel.IntType))
```
theunrepentantgeek, line 73: The `WithProperty()` method returns a new instance, which you're discarding. 
```suggestion
	t = t.WithProperty(astmodel.NewPropertyDefinition("FancyProp", "fancyProp", astmodel.IntType))
```

hack/generator/pkg/codegen/testdata/EmbeddedTypes/Embedded_type_simple_resource.golden
```diff
@@ -78,6 +81,34 @@ func (fakeResourceSpecArm FakeResource_SpecArm) GetType() string {
...
+
+var _ genruntime.ArmTransformer = &EmbeddedTestType{}
+
+// ConvertToArm converts from a Kubernetes CRD object to an ARM object
+func (embeddedTestType *EmbeddedTestType) ConvertToArm(name string) (interface{}, error) {
+	if embeddedTestType == nil {
+		return nil, nil
+	}
+	var result EmbeddedTestTypeArm
+	return result, nil
```
theunrepentantgeek, line 95: Shouldn't this method include an assignment or two to copy data across? This does nothing.

theunrepentantgeek: Or is this related to the missing property that I noted earlier?

hack/generator/pkg/config/configuration.go
```diff
@@ -243,12 +244,19 @@ func buildExportFilterFunc(f *ExportFilter, allTypes astmodel.Types) ExportFilte
...
 		applicableTypes := make(astmodel.TypeNameSet)
+		foundMatch := false
 		for tn := range allTypes {
 			if f.AppliesToType(tn) {
+				foundMatch = true
 				collectAllReferencedTypes(allTypes, tn, applicableTypes)
 			}
 		}
 
+		// TODO: should we consider somehow turning this into an err?
```
theunrepentantgeek, line 255: As long as an error here results in the generator aborting due to the useless filter, then _**yes**_ this should be an error.

theunrepentantgeek (changes requested):
I think I found one bug to fix.

Also, I don't understand why the change in `TypeVisitor` to return a blank `TypeDefinition` instead of **nil**, can you please share your motivation?

hack/generator/pkg/codegen/codegen_test.go
matthchr: oops!

matthchr:
Replying to this one here since I can't do it inline:

> Also, I don't understand why the change in TypeVisitor to return a blank TypeDefinition instead of nil, can you please share your motivation?

Because in all other contexts `TypeDefinition` is returned as a struct directly rather than a ptr type (similar to how we always have `TypeName` as a struct). For example if you look at `package_definition.go` `GetDefinition` or `code_generation_context.go` `GetImportedDefinition` they both return `TypeDefinition, error` not `*TypeDefinition, error`.

Moving away from `*TypeDefinition` saves us having to use the `*` operator to dereference the result when called - which is a safe operation since we're always checking the resulting `error` -- but since we're always checking the resulting `error` the optionality of the `TypeDefinition` also is conveying no extra information (that bit is already conveyed and checked in the `error` part of the return).

It's possible that I wrote all of the methods that are doing it the way I changed it to here though - I've asked about it in the Go slack channel in the hopes that somebody can tell me if I'm wrong here. I do think we should be consistent though so either we change the other calls to be this pattern, or we change this call to be the other pattern.

hack/generator/pkg/codegen/testdata/EmbeddedTypes/Embedded_type_simple_resource.golden
matthchr: Yes, `EmbeddedTestType` was empty before so there was nothing to copy.

hack/generator/pkg/codegen/codegen_test.go
matthchr: Fixed as part of another commit

hack/generator/pkg/codegen/testdata/EmbeddedTypes/Embedded_type_simple_resource.golden
matthchr: Really good catch here

hack/generator/pkg/astmodel/object_type.go
```diff
@@ -215,6 +215,12 @@ func (objectType *ObjectType) References() TypeNameSet {
 		}
 	}
 
+	for _, property := range objectType.embedded {
+		for ref := range property.PropertyType().References() {
+			results = results.Add(ref)
+		}
+	}
+
```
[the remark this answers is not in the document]
matthchr: I've just called `SetUnion` instead.

hack/generator/pkg/config/configuration.go
matthchr: Fixed.

hack/generator/pkg/astmodel/object_type.go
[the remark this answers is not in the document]
Porges: @matthchr we should add an `AddAll` instead of `SetUnion` so we don't copy the entire set on every single iteration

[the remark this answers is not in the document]
Porges: already implemented one here: https://github.com/Azure/k8s-infra/pull/394/files/4b6af574c0ce35f6a0c641b36e189b953c4de8ce#diff-9816f6d90fe771dbafda26064926a9c3b4f1b67b903092fac792a37aa8cd2d53

theunrepentantgeek (approved):

PR #406, opened 2021-03-04
Flatten properties

Porges:


hack/generator/pkg/astmodel/property_definition.go
```diff
@@ -55,11 +56,26 @@ func (property *PropertyDefinition) PropertyType() Type {
 
 // SetRequired sets if the property is required or not
 func (property *PropertyDefinition) SetRequired(required bool) *PropertyDefinition {
+	if required == property.isRequired {
+		return property
+	}
+
 	result := *property
```
theunrepentantgeek, line 63: This should use `.copy()` to safely copy the `tags` map as well.

hack/generator/pkg/astmodel/property_definition.go
```diff
@@ -55,11 +56,26 @@ func (property *PropertyDefinition) PropertyType() Type {
...
 	return &result
 }
 
+// SetFlatten sets if the property should be flattened or not
+func (property *PropertyDefinition) SetFlatten(flatten bool) *PropertyDefinition {
+	if flatten == property.flatten {
+		return property
+	}
+
+	result := *property
```
theunrepentantgeek, line 74: Ditto should use .copy() to safely copy the tags map as well.

hack/generator/pkg/astmodel/property_definition.go
Porges: do we still need to copy it even though we aren't modifying the tags?

Porges:
Reworking into different PRs.

PR #409, opened 2021-03-09
Submodule schemas

Porges:
Load schemas from submodule to avoid any issues with unexpected updates or CDN/network.

Achieved by adding a layer to rewrite schema URLs; tested by unplugging ethernet cable 😆

### DIFF
--- a/hack/generator/pkg/codegen/pipeline_load_schema.go
+++ b/hack/generator/pkg/codegen/pipeline_load_schema.go
@@ -9,6 +9,7 @@ import (
 	"context"
 	"net/http"
 	"os"
+	"strings"
 
 	"github.com/Azure/k8s-infra/hack/generator/pkg/astmodel"
 	"github.com/Azure/k8s-infra/hack/generator/pkg/config"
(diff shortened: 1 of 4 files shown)


hack/generator/pkg/codegen/pipeline_load_schema.go
```diff
@@ -45,6 +46,40 @@ func (factory *cancellableJSONLoaderFactory) New(source string) gojsonschema.JSO
...
+
+type rewritingJSONLoaderFactory struct {
+	from, to string
+	inner    gojsonschema.JSONLoaderFactory
+}
+
+var _ gojsonschema.JSONLoaderFactory = &rewritingJSONLoaderFactory{}
+
+func (factory *rewritingJSONLoaderFactory) New(source string) gojsonschema.JSONLoader {
+	return factory.inner.New(strings.Replace(source, factory.from, factory.to, 1))
```
babbageclunk, line 80: At first I thought source was the contents of the schema, but rereading it's the URL, right?

babbageclunk: 'Cause the former would be pretty blunt! Although I guess it would work, so :shrug: still would've been better than what we have now.

babbageclunk (approved):
This looks awesome!

PR #410, opened 2021-03-10
Simplify storage conversions

theunrepentantgeek:
This PR simplifies the generation of storage conversion functions so each one is exactly a single step in the conversion graph; we'll implement the whole graph in the `ConvertTo()` and `ConvertFrom()` methods as needed.

The names of the methods are also changed to avoid clashes with the `Convertible` interface.

Background:
I realized that the conversion methods were overly complex because they were all focused on conversion to/from the hub type, which doesn't actually work when we have nested types.

To illustrate, assume we have a `Person` type that contains a nested `Address`:

```
type Person struct {
    Address Address
}
```

Further, assume we have three versions in flight: `v1`, `v2` and `v3`
where `v3` is our "hub" storage type

When we are converting from `v1.Person` to `v2.Person`, we need to convert from a `v1.Address` to a `v2.Address`, _**not**_ from a `v1.Address` to a `v3.Address`.  Simplifying things so that each function converts just one step gives us the flexibility we need.

But, the `v1.Address` to a `v3.Address` conversion is what we were previously generating.


hack/generator/pkg/astmodel/storage_conversion_function.go
```diff
@@ -27,13 +26,9 @@ type StoragePropertyConversion func(
...
 	// name of this conversion function
 	name string
-	// hubType is the ultimate hub type to which (or from which) we are converting, passed as a
-	// parameter to our function
-	hubType TypeDefinition
-	// intermediateType optionally identifies a type which is "closer" to the hubType through which
-	// we can achieve our conversion. Will be nil if we are converting to/from the hub type
-	// directly, otherwise we use this as an intermediate form.
-	intermediateType *TypeDefinition
+	// otherType is the type we are converting to (or from). This will be a type which is "closer"
```
matthchr, line 29: I like this description

hack/generator/pkg/astmodel/storage_conversion_function.go
```diff
@@ -220,25 +205,12 @@ func (fn *StorageConversionFunction) generateBody(
...
-		case ConvertTo:
-			return fn.generateDirectConversionTo(receiver, parameter, generationContext)
-		default:
-			panic(fmt.Sprintf("unexpected conversion direction %q", fn.conversionDirection))
-		}
-	}
-
-	// Intermediate step of conversion, not working directly with the hubType type we've been given
-	// Instead we convert to/from our intermediate type (which is one step closer to the hub type in our conversion graph)
+	// Last step of conversion, directly working with the hubType type we've been given
```
matthchr, line 208: is `hubType` in this comment correct?

hack/generator/pkg/astmodel/testdata/NastyTest.golden
```diff
@@ -31,8 +31,8 @@ func (person Person) ConvertFrom(source *vHub.Person) error {
 	return nil
 }
 
-// ConvertTo populates the provided destination Person from our Person
-func (person Person) ConvertTo(destination *vHub.Person) error {
+// ConvertToVNext populates the provided destination Person from our Person
+func (person Person) ConvertToVNext(destination *vNext.Person) error {
```
matthchr, line 35: minor: I generally favor full words in methods. Would `ConvertToNextVersion` and `ConvertFromNextVersion` (or _possibly_ `ConvertToHigherVersion` and `ConvertFromHigherVersion`, assuming that my assumption that your "vnext" is always a "bigger" version (orders after)) be clearer?

matthchr (approved):
LGTM - had one comment on method naming

hack/generator/pkg/astmodel/testdata/NastyTest.golden
theunrepentantgeek: The `VNext` in the method name comes from the `vNext` namespace of the parameter. 

In actual use, the namespace will be something like `v1api20180101` so the method name will be `ConvertToV1api20180101` which is really explicit.

Porges (approved):

hack/generator/pkg/astmodel/storage_conversion_function.go
theunrepentantgeek: No it's not, and I don't think we need any comment here now. Removed.

hack/generator/pkg/astmodel/testdata/NastyTest.golden
matthchr: Totally missed that

PR #418, opened 2021-03-18
Fix controller_resources_gen.go generation

matthchr:
- Only storage types should be used to create GenericReconciler's.
  - All types should be used to register webhooks.
  - Improve resolver to have a better chance of finding the storage
    version.

### DIFF
--- a/hack/generated/controllers/generic_controller.go
+++ b/hack/generated/controllers/generic_controller.go
@@ -22,6 +22,7 @@ import (
 	"k8s.io/apimachinery/pkg/conversion"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	kerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/client-go/tools/record"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -107,20 +108,48 @@ func (options *Options) setDefaults() {
 	}
 }
 
-func RegisterAll(mgr ctrl.Manager, applier armclient.Applier, objs []runtime.Object, log logr.Logger, options Options) []error {
-	options.setDefaults()
-
+func RegisterWebhooks(mgr ctrl.Manager, objs []runtime.Object) error {
 	var errs []error
+
 	for _, obj := range objs {
-		if err := register(mgr, applier, obj, log, options); err != nil {
+		if err := registerWebhook(mgr, obj); err != nil {
 			errs = append(errs, err)
 		}
 	}
 
-	return errs
+	return kerrors.NewAggregate(errs)
 }
 
-func register(mgr ctrl.Manager, applier armclient.Applier, obj runtime.Object, log logr.Logger, options Options) error {
+func registerWebhook(mgr ctrl.Manager, obj runtime.Object) error {
+	_, err := conversion.EnforcePtr(obj)
+	if err != nil {
+		return errors.Wrap(err, "obj was expected to be ptr but was not")
+	}
+
+	return ctrl.NewWebhookManagedBy(mgr).
+		For(obj).
+		Complete()
+}
+
+func RegisterAll(mgr ctrl.Manager, applier armclient.Applier, objs []runtime.Object, log logr.Logger, options Options) error {
+	options.setDefaults()
+
+	reconciledResourceLookup, err := MakeResourceGVKLookup(mgr, objs)
+	if err != nil {
+		return err
+	}
+
+	var errs []error
+	for _, obj := range objs {
+		if err := register(mgr, reconciledResourceLookup, applier, obj, log, options); err != nil {
+			errs = append(errs, err)
+		}
+	}
+
+	return kerrors.NewAggregate(errs)
+}
+
+func register(mgr ctrl.Manager, reconciledResourceLookup map[schema.GroupKind]schema.GroupVersionKind, applier armclient.Applier, obj runtime.Object, log logr.Logger, options Options) error {
 	v, err := conversion.EnforcePtr(obj)
 	if err != nil {
 		return errors.Wrap(err, "obj was expected to be ptr but was not")
@@ -144,7 +173,7 @@ func register(mgr ctrl.Manager, applier armclient.Applier, obj runtime.Object, l
 	reconciler := &GenericReconciler{
 		ARMClient:            applier,
 		KubeClient:           kubeClient,
-		ResourceResolver:     armresourceresolver.NewResolver(kubeClient),
+		ResourceResolver:     armresourceresolver.NewResolver(kubeClient, reconciledResourceLookup),
 		Name:                 t.Name(),
 		Log:                  log.WithName(controllerName),
 		Recorder:             mgr.GetEventRecorderFor(controllerName),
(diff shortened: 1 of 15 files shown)


hack/generated/controllers/generic_controller.go
```diff
@@ -164,9 +192,27 @@ func register(mgr ctrl.Manager, applier armclient.Applier, obj runtime.Object, l
...
 
-	return ctrl.NewWebhookManagedBy(mgr).
-		For(obj).
-		Complete()
+	return nil
+}
+
+// MakeGroupVersionMap creates a map of group to version. This returns an error if there is more than one group to version mapping
+// for a given group.
+func MakeGroupVersionMap(mgr ctrl.Manager, objs []runtime.Object) (map[string]schema.GroupVersion, error) {
```
matthchr, line 200: This pattern won't work if we have a situation where a group has some resources which were removed in an old version, as then the storage type is that old version. I should change this to be based on Kind -- you can look up the Group-Version for each Kind.

babbageclunk: Do you mean group-kind? We're modelling the different RPs as different groups, right?

matthchr: Yeah I mean group-kind

babbageclunk (approved):
This looks good, other than the problem you point out with only allowing one version per group.

hack/generated/pkg/reflecthelpers/reflect_helpers_test.go
```diff
@@ -58,23 +58,38 @@ func createDummyResource() *batch.BatchAccount {
 	}
 }
 
+func MakeGroupToVersionMap(scheme *runtime.Scheme) (map[string]schema.GroupVersion, error) {
+	result := make(map[string]schema.GroupVersion)
+	for gvk := range scheme.AllKnownTypes() {
+		if gvk.Group == "" {
+			continue
+		}
+		result[gvk.Group] = schema.GroupVersion{Group: gvk.Group, Version: gvk.Version}
```
theunrepentantgeek, line 67: Does this need to prevent overwriting a group if it's there twice?

hack/generated/pkg/util/armresourceresolver/resolver_test.go
```diff
@@ -26,9 +26,22 @@ import (
...
+	return NewResolver(kubeclient.NewClient(fakeClient, s), reconciledGroupVersionsMap)
+}
+
+func MakeGroupToVersionMap(scheme *runtime.Scheme) (map[string]schema.GroupVersion, error) {
+	result := make(map[string]schema.GroupVersion)
+	for gvk := range scheme.AllKnownTypes() {
+		if gvk.Group == "" {
+			continue
+		}
+		result[gvk.Group] = schema.GroupVersion{Group: gvk.Group, Version: gvk.Version}
```
theunrepentantgeek, line 41: Ditto here - do we need to avoid overwrites?

hack/generated/pkg/reflecthelpers/reflect_helpers_test.go
matthchr: Since this is used purely in test it's not a big deal, but I've fixed this as part of the refactor to solve the group-kind issue mentioned above.

hack/generated/pkg/util/armresourceresolver/resolver_test.go
matthchr: Since this is used purely in test it's not a big deal, but I've fixed this as part of the refactor to solve the group-kind issue mentioned above.

hack/generated/controllers/generic_controller.go
```diff
@@ -107,20 +107,48 @@ func (options *Options) setDefaults() {
 	}
 }
 
+func RegisterWebhooks(mgr ctrl.Manager, objs []runtime.Object) []error {
```
theunrepentantgeek, line 110: Is the signature for this fixed, or could you return just `error` and use `kerrors.NewAggregate()` as we have elsewhere to combine the `[]error` into one?

hack/generated/controllers/generic_controller.go
```diff
@@ -107,20 +107,48 @@ func (options *Options) setDefaults() {
...
+	if err != nil {
+		return errors.Wrap(err, "obj was expected to be ptr but was not")
+	}
+
+	return ctrl.NewWebhookManagedBy(mgr).
+		For(obj).
+		Complete()
+}
+
 func RegisterAll(mgr ctrl.Manager, applier armclient.Applier, objs []runtime.Object, log logr.Logger, options Options) []error {
```
theunrepentantgeek, line 133: Same comment about using `error` instead of `[]error` here.

hack/generator/pkg/codegen/resource_registration_file.go
```diff
@@ -99,7 +108,7 @@ func (r *ResourceRegistrationFile) AsAst() (*dst.File, error) {
 func (r *ResourceRegistrationFile) generateImports() (*astmodel.PackageImportSet, error) {
 	requiredImports := astmodel.NewPackageImportSet()
 
-	for _, typeName := range r.resources {
+	for _, typeName := range append(r.resources, r.storageVersionResources...) {
```
theunrepentantgeek, line 111: Perhaps use `PackageImportSet.versionedNameForImport()` to create the unique names, instead of reinventing the wheel?

hack/generator/pkg/codegen/resource_registration_file.go
```diff
@@ -142,7 +151,24 @@ func makeClientGoKubernetesSchemePackageReference() astmodel.PackageReference {
...
+//			result = append(result, new(<package>.<resource>))
+//			result = append(result, new(<package>.<resource>))
+//			result = append(result, new(<package>.<resource>))
+//			...
+//			return result
+//		}
+func (r *ResourceRegistrationFile) createGetKnownStorageTypesFunc(codeGenerationContext *astmodel.CodeGenerationContext) (dst.Decl, error) {
+	return createKnownTypesFuncImpl(
+		codeGenerationContext,
+		r.storageVersionResources,
```
theunrepentantgeek, line 166: Where do we ensure these are always in the same order, so that codegen is deterministic?

hack/generated/controllers/generic_controller.go
matthchr: Yeah we can move to use `error` in both. Have done so.

hack/generated/controllers/generic_controller.go
matthchr: Yeah we can move to use `error` in both. Have done so.

hack/generator/pkg/codegen/resource_registration_file.go
matthchr: Inside the function itself, line 206

hack/generator/pkg/codegen/resource_registration_file.go
matthchr: Promoted that method to public and used it instead. I tried to use the `ResolveConflicts` method but that didn't get the same formatting as none of the names conflicted -- since we know the form of all of the imports in this file it made sense to me to force them to be named "nicely" (which is what the `VersionedNameForImport` does).

Take a look and see if you like it.

matthchr: I moved the `VersionedNameForImport` and `ServiceNameForImport` to `PackageImport` rather than the set to make this easier for me to do (I think they make more sense there anyway as it's an operation on the import not on a set)

theunrepentantgeek: Looks good. :grin:

theunrepentantgeek (approved):

PR #419, opened 2021-03-18
Improve Integration tests by adding headers to each stage

theunrepentantgeek:
Adds banner headings into the log output for each integration test, allowing the different stages/phases of the test to be easily identified while scanning the test log. 

Each banner is numbered for reference:
```
[controller:test-integration-envtest-cover] 2021/03/18 23:55:16 Webhook server running at: :34365
[controller:test-integration-envtest-cover] 2021/03/18 23:55:16 http: TLS handshake error from 127.0.0.1:46798: EOF
[controller:test-integration-envtest-cover] -----------------------
[controller:test-integration-envtest-cover] 1 Create resource group
[controller:test-integration-envtest-cover] -----------------------
[controller:test-integration-envtest-cover] 2021/03/18 23:55:16 Creating test resource group "k8sinfratest-rg-dnhfns"
[controller:test-integration-envtest-cover] 2021/03/18 23:55:17 Creating & starting controller-runtime manager
[controller:test-integration-envtest-cover] 2021/03/18 23:55:17 Minimizing requeue delay
[controller:test-integration-envtest-cover] 2021/03/18 23:55:17 Registering custom controllers
```

Nested headings are supported too:
```
[controller:test-integration-envtest-cover] === RUN   Test_StorageAccount_CRUD/Blob_Services_CRUD
[controller:test-integration-envtest-cover] -----------------------
[controller:test-integration-envtest-cover] 2.1 Create blob service
[controller:test-integration-envtest-cover] -----------------------
[controller:test-integration-envtest-cover] I0318 23:55:29.118536   11911 generic_controller.go:390] StorageAccountsBlobServiceController "msg"="Starting new deployment to Azure" "name"="k8sinfratest-blobservice-nayxdk" "namespace"="k8s-infra-test-ns" "action"="BeginDeployment"
```

### DIFF
--- a/hack/generated/controllers/crd_resourcegroup_test.go
+++ b/hack/generated/controllers/crd_resourcegroup_test.go
@@ -12,17 +12,22 @@ import (
 	. "github.com/onsi/gomega"
 
 	"github.com/Azure/k8s-infra/hack/generated/pkg/armclient"
+	util "github.com/Azure/k8s-infra/hack/generated/pkg/util"
 )
 
 func Test_ResourceGroup_CRUD(t *testing.T) {
 	t.Parallel()
 
 	g := NewGomegaWithT(t)
+	log := util.NewBannerLogger()
+	log.Header(t.Name())
+
 	ctx := context.Background()
 	testContext, err := testContext.ForTest(t)
 	g.Expect(err).ToNot(HaveOccurred())
 
 	// Create a resource group
+	log.Subheader("Create resource group")
 	rg := testContext.NewTestResourceGroup()
 	err = testContext.KubeClient.Create(ctx, rg)
 	g.Expect(err).ToNot(HaveOccurred())
@@ -36,6 +41,7 @@ func Test_ResourceGroup_CRUD(t *testing.T) {
 	armId := rg.Status.ID
 
 	// Delete the resource group
+	log.Subheader("Delete resource group")
 	err = testContext.KubeClient.Delete(ctx, rg)
 	g.Expect(err).ToNot(HaveOccurred())
 	g.Eventually(rg).Should(testContext.Match.BeDeleted(ctx))
--- a/hack/generated/controllers/crd_cosmosdb_test.go
+++ b/hack/generated/controllers/crd_cosmosdb_test.go
@@ -12,6 +12,8 @@ import (
 	. "github.com/onsi/gomega"
 
 	documentdb "github.com/Azure/k8s-infra/hack/generated/_apis/microsoft.documentdb/v20150408"
+	util "github.com/Azure/k8s-infra/hack/generated/pkg/util"
+
 	"github.com/Azure/k8s-infra/hack/generated/pkg/testcommon"
 )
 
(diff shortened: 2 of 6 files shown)


hack/generated/controllers/crd_resourcegroup_test.go
```diff
@@ -12,17 +12,22 @@ import (
...
 	"github.com/Azure/k8s-infra/hack/generated/pkg/armclient"
+	util "github.com/Azure/k8s-infra/hack/generated/pkg/util"
 )
 
 func Test_ResourceGroup_CRUD(t *testing.T) {
 	t.Parallel()
 
 	g := NewGomegaWithT(t)
+	log := util.NewBannerLogger()
+	log.Header(t.Name())
```
matthchr, line 23: Discussion here: https://github.com/golang/go/issues/19280 suggests a few things to me:

Firstly, `== CONT <TESTNAME>` is always logged as the first line of a test already. It's not clear to me that another line directly below that is all that helpful (even if it is a bit easier to read it also requires test authors to manually write that long line in their tests which they're likely to forget).

Secondly (from the linked issue above): 
> In order to see logging interleaved in a more structured way with test output, one would use t.Log, not println, log, or fmt.
> ...
> If you use t.Log, the logs will be grouped with the respective tests. In this case, tests collect the logs for this test and flush it to the parent on completion. Top-level tests flush to stdout.

The `BannerLogger` is not using `t.Log` which means that it will be outputting logs that may end up interleaved or separated from the tests in question (I think? @babbageclunk maybe knows more about this?).

Additionally, I am not sure that the log output from the tests themselves (which is really coming from the controllers hosted by envtest) is going to be located in the right place, because it _also_ isn't using `t.Log` as far as I can tell.

The `controllers.RegisterAll` function already takes a `log logr.Logger` parameter, and I am wondering if all we really need to do to make the test output more clear is to supply one of those that ends up routing to `t.Log`. That should remove any interleaving going on in test logs as well. We call `controllers.RegisterAll` in `kube_test_context_envtest.go`.

hack/generated/controllers/crd_resourcegroup_test.go
```diff
@@ -12,17 +12,22 @@ import (
...
+
 	ctx := context.Background()
 	testContext, err := testContext.ForTest(t)
 	g.Expect(err).ToNot(HaveOccurred())
 
 	// Create a resource group
+	log.Subheader("Create resource group")
```
matthchr, line 30: I'm somewhat concerned about needing to remember to add these logs for each interesting place in a test. It feels like it adds difficulty in authoring a test and means I can get it "wrong" (but my test still passes).

I'm wondering if a better pattern would be to hook `testContext.KubeClient.Create` (and `Delete` and possibly `Update` too) to do this automatically by inspecting the type in question. So you'd get something like:
`"Creating %q", obj.Kind`.

matthchr (changes requested):

hack/generated/controllers/crd_resourcegroup_test.go
Porges: Agree that we should be using the facilities provided by `testing.T` and not `fmt`.

theunrepentantgeek:
Abandoning in favour of work to properly isolate the logs for each test

PR #420, opened 2021-03-22
Add TypeWalker for walking graph of types

matthchr:


matthchr:
You don't see any usage of this in a pipeline yet as the pipeline that will use this (removing recursive resource references) will come in a future PR.

hack/generator/pkg/astmodel/type_walker.go
```diff
@@ -0,0 +1,195 @@
...
+	"github.com/pkg/errors"
+)
+
+// TODO: This is conceptually kinda close to ReferenceGraph except more powerful
+
+// TypeWalkerRemoveType is a special TypeName that informs the type walker to remove the property containing this TypeName
+// entirely.
+var TypeWalkerRemoveType = MakeTypeName(LocalPackageReference{}, "TypeWalkerRemoveProperty")
+
+// TODO: it's awkward to have so much configuration on this thing (3 separate funcs that apply at different places in the walking process?)
```
matthchr, line 20: Bringing this up in code review.

If anybody has a cleaner way to make this thing less configurable while maintaining the power I'm 100% open to it. I'm basically trying to make a "super-visitor" that does a depth-first walk of a whole graph of types (including following the references between types). So the result of `Walk` is going to be the `Types` collection that is reachable from the root definition you provided, with members of that collection modified by the appropriate rules.

The two that feel most awkward to me are `MakeContextFunc` and `AfterVisitFunc`.
`AfterVisitFunc` for example is really only adding a _tiny_ bit on top of the visitor anyway, as it's allowing a modification of the `TypeDefinition` produced after the visitor has been called (which effectively boils down to the extra ability to fuss with the name of the type). It almost feels like this could be embedded into the visitor somehow and we could then remove this extra bit of configuration... but I poked at it and couldn't come up with a clean way.

matthchr: Those 2 are in contrast to `WalkCycle` which I think is inevitably here - the `TypeWalker` tracks what constitutes a cycle and so you need to have a hook to say: "When you find a cycle what do you want me to do" - there's no other place that bit of information could go I don't think.

babbageclunk: I think I follow what you're saying about pushing the before and after hooks down into the visitor, but it's hard to visualise a better way to do it without seeing the usage. Presumably the visitor that gets passed in when making the walker has already had some some custom visit functions applied? One argument for keeping them as TypeWalker configuration is if it would ever make sense to use a customised visitor with two different TypeWalkers with different impls for before/after visit and cycle handling? That might be something that isn't ever really done though.

I don't think this seems bad as it is - definitely ok to land like this and then tweak it as clean ups occur to us.

hack/generator/pkg/astmodel/type_walker.go
```diff
@@ -0,0 +1,195 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	visitedTypeName, err := t.originalVisitTypeName(this, it, updatedCtx)
+	if err != nil {
+		return nil, err
+	}
+	var ok bool
+	it, ok = visitedTypeName.(TypeName)
```
babbageclunk, line 80: Why not
```suggestion
	it, ok := visitedTypeName.(TypeName)
```
?

hack/generator/pkg/astmodel/type_walker.go
```diff
@@ -0,0 +1,195 @@
...
+		panic(fmt.Sprintf("TypeWalker visitor VisitTypeName must return a TypeName, instead returned %T", visitedTypeName))
+	}
+
+	def, ok := t.allTypes[it]
+	if !ok {
+		return nil, errors.Errorf("couldn't find type %q", it)
+	}
+
+	// Prevent loops by bypassing this type if it's currently being processed. The processing
+	// slice is basically the "path" taken to get to the current type.
```
babbageclunk, line 91: Looks like this comment was written before processing was converted to a map?

hack/generator/pkg/astmodel/type_walker.go
```diff
@@ -0,0 +1,195 @@
...
+// TODO: This is conceptually kinda close to ReferenceGraph except more powerful
+
+// TypeWalkerRemoveType is a special TypeName that informs the type walker to remove the property containing this TypeName
+// entirely.
+var TypeWalkerRemoveType = MakeTypeName(LocalPackageReference{}, "TypeWalkerRemoveProperty")
+
+// TODO: it's awkward to have so much configuration on this thing (3 separate funcs that apply at different places in the walking process?)
+// TODO: but unsure if there's a better way... bring it up in code review.
+
+// TypeWalker performs a depth first search across the types provided, applying the visitor to each TypeDefinition.
```
babbageclunk, line 23:
```suggestion
// TypeWalker performs a depth first traversal across the types provided, applying the visitor to each TypeDefinition.
```

hack/generator/pkg/astmodel/type_walker.go
```diff
@@ -0,0 +1,195 @@
...
+		return nil, err
+	}
+
+	ot, ok := result.(*ObjectType)
+	if !ok {
+		panic(fmt.Sprintf("TypeWalker visitor VisitObjectType must return a *ObjectType, instead returned %T", result))
+	}
+
+	for _, prop := range ot.Properties() {
+		isCycle := isRemoveType(prop.PropertyType())
```
babbageclunk, line 128: Presumably a visitor could return a property with the remove type for other reasons than it containing a cycle? Although I guess that would be a weird thing to do, when the visitor could just remove the property itself in that case rather than needing to ask the walker to do it. The only time it needs to get the walker to do it is if the walker detects a cycle, is that right? And not all of the time, since in that case there'd be no need for a callback. I feel like I need to see the usage to get a feel for how that decision gets made.

Is the WalkCycle interface too flexible? What about just making it a predicate that tells the walker whether to remove the property? I guess that doesn't work since it's in the wrong place. It feels like the interaction between the visitor and the walker is very complicated (but that might be because I can only see half of it).

hack/generator/pkg/astmodel/type_walker.go
```diff
@@ -0,0 +1,195 @@
...
+		if isCycle {
+			ot = ot.WithoutProperty(prop.PropertyName())
+		}
+	}
+
+	return ot, nil
+}
+
+// Walk returns a Types collection constructed by applying the Visitor to each type in the graph of types reachable
+// from the provided TypeDefinition 'def'. Types are visited in a depth-first order. Cycles are not visited.
```
babbageclunk, line 138: Is it right that cycles aren't visited? I think it's that the walk down the graph will stop when it detects a cycle - so the types in the cycle will be visited but only once. But I could definitely be wrong.

hack/generator/pkg/astmodel/type_walker.go
```diff
@@ -0,0 +1,195 @@
...
+		processing: make(map[TypeName]struct{}),
+	}
+
+	t.state.processing[def.Name()] = struct{}{}
+
+	updatedType, err := t.visitor.Visit(def.Type(), ctx)
+	if err != nil {
+		return nil, err
+	}
+	updatedDef, err := t.AfterVisitFunc(def, def.WithType(updatedType), ctx)
```
babbageclunk, line 151: Should there be a call to `BeforeVisit` before the visit? Or is this why the callback shouldn't be called `BeforeVisit`?

babbageclunk (commented):
I think this makes sense but it's hard to be sure without seeing the interplay of the visitor and the walker callback functions.

hack/generator/pkg/astmodel/type_walker.go
matthchr: @babbageclunk good point w.r.t. tweaking in the future if need be

hack/generator/pkg/astmodel/type_walker.go
matthchr: uhh, I don't know =)

Probably from some previous refactor -- fixed.

hack/generator/pkg/astmodel/type_walker.go
```diff
@@ -0,0 +1,195 @@
...
+	// visitor is a copy - modifications won't impact passed visitor
+	visitor.VisitTypeName = typeWalker.visitTypeName
+	visitor.VisitObjectType = typeWalker.visitObjectType
+
+	typeWalker.visitor = visitor
+	typeWalker.AfterVisitFunc = DefaultAfterVisit
+	typeWalker.MakeContextFunc = IdentityMakeContext
+	typeWalker.WalkCycle = IdentityWalkCycle
+
+	return &typeWalker
```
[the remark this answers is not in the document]
matthchr: Agreed.

hack/generator/pkg/astmodel/type_walker.go
matthchr: The tests show some sample usage. I do agree that the interplay seems complicated - I think I was struggling to express that in my initial comment on this PR.

The problem with the visitor removing types/properties is that depending on which handler in the visitor makes the decision to remove, you have a nesting problem...
For example let's say that you'd like to remove all `ObjectType` properties that have a particular `Flag` on the `Type`. You have only one way to do that with the visitor as it is today, which is to put the per-property detection logic into `VisitObjectType`, so you'd do something like:

```
func VisitObjectType(...) (Type, error) {
    // loop over all properties
    // check if somewhere in property type hierarchy there is a flag type
    // Check if that flag type has the flag we're looking for, if so, remove.
}
```

That's fine and it works but the issue is that it requires you to "pull up" the processing that could otherwise be done in the `VisitType` for the property. (imagine if instead of removal we just wanted to change the name of every Flag on a FlagType from `Foo` to `Bar` -- we would be doing that in the `VisitFlagType` bit of the visitor). 
In the `TypeWalker` context this "pulling up" is problematic because the function where we want the removal logic to reside is `VisitTypeName` (since that's what's walking to the next type in the graph and where we can do cycle detection easily) - _but_ you can't remove a type in `VisitTypeName`, so instead we would have to pull it up to `VisitObjectType` and do something like:

```
func VisitObjectType(...) (Type, error) {
    // loop over all properties
    // check if one of the has a typename that would trigger a cycle, probably by doing basically what my `VisitTypeName` is doing today
    // Remove the property
    // Call identityVisitObjectType which will eventually call `VisitTypeName` again on all of the existing properties to actually do the walk.
}
```

So it's not impossible to avoid the need for the `TypeWalkerRemoveType` entirely, it just means that the call-graph of what the visitor is doing gets a bit more complicated because there end up being 2 entrypoints to `VisitTypeName` - one that's a pre-check for removal, and then the real one for when we know we don't need to remove.

*Note: In actuality it'd probably be `VisitTypeName` calling some worker function and that worker function being reused in a slight different context (with different args?) in `VisitObjectType`, since as discussed above `VisitTypeName` can't actually convey the concept of removal with its signature today (at least not without some sentinel `RemovalType` like I've added.

Maybe it would make sense to promote this removal capability into visitor itself somehow? (That does just move complexity around a bit but might be more natural)

matthchr: To your original comment though, yes the variable here shouldn't be `isCycle`, since the idea was you can remove for reasons other than cycles.

hack/generator/pkg/astmodel/type_walker.go
matthchr: No you're exactly right. I've corrected this language to:
> Types are visited in a depth-first order. Cycles are not followed (so each type in a cycle will be visited only once).

hack/generator/pkg/astmodel/type_walker.go
matthchr: Hmm, yeah I see your point here.

Since `MakeContextFunc` is all about managing the `ctx` for each type in the graph, it's not called here because the seed `ctx` is used for the first call. The N+1st call uses a `ctx` created by calling `MakeContextFunc(nthCtx)`.

hack/generator/pkg/astmodel/type_walker.go
matthchr: Nice

hack/generator/pkg/astmodel/type_walker.go
matthchr: Fixed

hack/generator/pkg/astmodel/type_walker.go
```diff
@@ -0,0 +1,195 @@
...
+
+// TypeWalkerRemoveType is a special TypeName that informs the type walker to remove the property containing this TypeName
+// entirely.
+var TypeWalkerRemoveType = MakeTypeName(LocalPackageReference{}, "TypeWalkerRemoveProperty")
+
+// TODO: it's awkward to have so much configuration on this thing (3 separate funcs that apply at different places in the walking process?)
+// TODO: but unsure if there's a better way... bring it up in code review.
+
+// TypeWalker performs a depth first search across the types provided, applying the visitor to each TypeDefinition.
+// MakeContextFunc is called before each visit, and AfterVisitFunc is called after each visit. WalkCycle is called
```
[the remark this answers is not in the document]
matthchr: I've dropped the `Func`.

Even though `MakeContext` is called before the `Visit`, I think it needs a more descriptive name than `BeforeVisit`. I would expect `BeforeVisit` to have a signature something like `AfterVisit` does, whereas `MakeContext` has a very different signature.

hack/generator/pkg/astmodel/type_walker.go
matthchr: I've pushed a draft PR that shows actual usage here: https://github.com/Azure/k8s-infra/pull/422/files#diff-cd9aa548dd1752cd7d7998f99bc00ef5c2e6e69154d26efc1e8e02e376cb9de3R268

This PR looks bigger than it really is because it includes this change and a bunch of golden files tests, so don't be too afraid of it. Don't bother reviewing it yet though, I'll mark it as not draft once it's ready for a real review.

hack/generator/pkg/astmodel/type_walker.go
matthchr: As discussed, changed it so that there is no context on `Walk` and the context is always manufactured by `MakeContext`. This allowed me to reduce a bit of duplication between `Walk` and `visitTypeName`. It does mean that the `MakeContext` function needs to handle its own base case (if it's nil) though, which does push a small bit of complexity to the caller.

I think it makes "where did this context come from" easier to reason about though.

hack/generator/pkg/astmodel/type_walker.go
```diff
@@ -0,0 +1,194 @@
...
+	allTypes Types
+	visitor  TypeVisitor
+
+	// MakeContext is called before any type but the root type is visited. It is given the current context and returns
+	// a new context for use in the upcoming Visit.
+	MakeContext func(it TypeName, ctx interface{}) (interface{}, error)
+	// AfterVisit is called after the type walker has applied the visitor to a TypeDefinition.
+	AfterVisit func(original TypeDefinition, updated TypeDefinition, ctx interface{}) (TypeDefinition, error)
+	// RemoveCycle is called if a cycle is detected. If true is returned the cycle will be pruned, otherwise it will be preserved as-is.
+	RemoveCycle func(def TypeDefinition, ctx interface{}) (bool, error)
```
babbageclunk, line 36: This sounds like the callback should do the removal - what about calling it `ShouldRemoveCycle`?

hack/generator/pkg/astmodel/type_walker.go
```diff
@@ -0,0 +1,194 @@
...
+// if a cycle is detected.
+type TypeWalker struct {
+	allTypes Types
+	visitor  TypeVisitor
+
+	// MakeContext is called before any type but the root type is visited. It is given the current context and returns
+	// a new context for use in the upcoming Visit.
+	MakeContext func(it TypeName, ctx interface{}) (interface{}, error)
+	// AfterVisit is called after the type walker has applied the visitor to a TypeDefinition.
+	AfterVisit func(original TypeDefinition, updated TypeDefinition, ctx interface{}) (TypeDefinition, error)
```
babbageclunk, line 34: Now I really want these to be called `PushContext` and `PopContext` (or `PushVisit` and `PopVisit`?). I think those make sense with how they're used, even the nil being passed in for the initial context (like the next pointer in a linked list being used as a stack).

The MakeContext doc needs updating now that it's also called for the root type.

babbageclunk (approved):
I'm much more comfortable with this now - just some suggestions for different names for the callbacks.

hack/generator/pkg/astmodel/type_walker.go
matthchr: Agree.

hack/generator/pkg/astmodel/type_walker.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+
+	def, ok := t.allTypes[it]
+	if !ok {
+		return nil, errors.Errorf("couldn't find type %q", it)
+	}
+
+	// Prevent loops by bypassing this type if it's currently being processed.
+	if _, ok := t.state.processing[def.Name()]; ok {
+		remove, err := t.RemoveCycle(def, updatedCtx)
```
theunrepentantgeek, line 92: Does it matter that this can be called with any element of the cycle?

'cause I don't believe this is deterministic. The `TypeVisitor`' method `IdentityVisitOfObjectType` iterates over `properties` without sorting (it's a map), so a cycle might be entered at a different point from run to run.

To illustrate:

``` go
type Person struct {
    enrolments []Enrolment
}

type Enrolment {
    student Person
    course  Course
}

type Course struct {
    lecturer Person
    tutors   []Person
    students []Enrolment
}
```
One cycle is:  
Person.enrolments -> Enrolment.student -> Person

Another is
Person.enrolments -> Enrolment.course -> Course.lecturer -> Person

Another is
Person.enrolments -> Enrolment.course -> Course.students -> Enrolment

Depending on which cycle is encountered first, won't you end up with different results from the walk?

hack/generator/pkg/astmodel/type_walker.go
```diff
@@ -0,0 +1,194 @@
...
+	_, err := t.visitor.Visit(def.Name(), nil)
+	if err != nil {
+		return nil, err
+	}
+
+	return t.state.result, nil
+}
+
+// DefaultAfterVisit is the default AfterVisit function for TypeWalker. It returns the TypeDefinition from Visit unmodified
+func DefaultAfterVisit(_ TypeDefinition, updated TypeDefinition, _ interface{}) (TypeDefinition, error) {
```
theunrepentantgeek, line 165: Should this be called `IdentityAfterVisit`?

hack/generator/pkg/astmodel/types.go
```diff
@@ -54,6 +54,24 @@ func (types Types) AddTypes(otherTypes Types) {
 	}
 }
 
+// AddWithEqualityCheck attempts to add the specified definition to the types collection.
+// Multiple adds of a type with the same shape are allowed, but attempting to add two
+// types with the same name but different shape will trigger an error.
+func (types Types) AddWithEqualityCheck(def TypeDefinition) error {
```
theunrepentantgeek, line 60: Does this need to be a separate method? I'm thinking that it could be merged into the existing `Add()` method as it really is safe to add something that's already there.

theunrepentantgeek (changes requested):

hack/generator/pkg/astmodel/type_walker.go
matthchr: Wow really good @theunrepentantgeek.

Yes I found and fixed this bug in my future PR. I will backport it to this PR as you point it it can cause problems here.
I specifically wrote a golden file test for this case (which will stay in my upcoming PR) because as you say, it can happen!

Really impressed you caught that just with a quick review, it took me a while to figure out what was happening when I ran into it!

hack/generator/pkg/astmodel/type_walker.go
matthchr: Probably

hack/generator/pkg/astmodel/types.go
matthchr: Aren't they expressing two different intents?

`Add` as it exists today panic's if there are duplicates. This is sorta nice for cases where you really don't expect to ever have duplicates (which is a lot of the time when you're say processing every type in the `Types` collection once, which a lot of our pipeline stages do).

`AddWithEqualityCheck` lets you expand what you're willing to tolerate (some duplicates, but only if they are structurally identical).

A lot of the common cases where we call `Add` today we really don't expect any duplicates at all because we're effectively doing a 1-1 mapping from an old `Types` set to a new `Types` set. It's only when you start treating things as a graph you run into duplicates and end up needing this.

I'd be fine combining them though if folks felt we don't really care about the above subtlety

hack/generator/pkg/astmodel/type_walker.go
matthchr: Pushed my fix for this (which makes `IdentityVisitOfObjectType` iterate properties in a deterministic order).

hack/generator/pkg/astmodel/type_walker.go
```diff
@@ -0,0 +1,194 @@
...
+
+	typeWalker.visitor = visitor
+
+	return &typeWalker
+}
+
+func (t *TypeWalker) visitTypeName(this *TypeVisitor, it TypeName, ctx interface{}) (Type, error) {
+	updatedCtx, err := t.MakeContext(it, ctx)
+	if err != nil {
+		return nil, err
```
theunrepentantgeek, line 73: Do these different error returns all need to use `errors.Wrapf()` to provide some context?
(Some do, some don't.)

hack/generator/pkg/astmodel/type_walker.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+
+	visitedTypeName, err := t.originalVisitTypeName(this, it, updatedCtx)
+	if err != nil {
+		return nil, err
+	}
+	it, ok := visitedTypeName.(TypeName)
```
theunrepentantgeek, line 80: Do you want exactly a `TypeName` or should you use `AsTypeName()`?

hack/generator/pkg/astmodel/type_walker.go
```diff
@@ -0,0 +1,194 @@
...
+func IdentityMakeContext(_ TypeName, ctx interface{}) (interface{}, error) {
+	return ctx, nil
+}
+
+// IdentityShouldRemoveCycle is the default cycle removal behavior. It preserves all cycles unmodified.
+func IdentityShouldRemoveCycle(_ TypeDefinition, _ interface{}) (bool, error) {
+	return false, nil
+}
+
+func isRemoveType(t Type) bool {
```
theunrepentantgeek, line 179: Perhaps there's a better name? Here's my stab at something more intention revealing:

```suggestion
func shouldBeRemoved(t Type) bool {
```

hack/generator/pkg/astmodel/types.go
theunrepentantgeek: 🤔 

I was thinking about it from an outcome perspective - adding something that's already there is safe because it doesn't change the result.

But I see you're thinking about it from the intent perspective - that we should never try to add something that's already there in the first place. This is a good safety check.

So I'd suggest keeping it as is, but can we find a better name for it?

theunrepentantgeek (commented):
Some minor questions, but mostly looks good to me.

hack/generator/pkg/astmodel/type_walker.go
theunrepentantgeek: @matthchr it's mostly a case of damaged shins. I've walked into the furniture of non-deterministic execution enough times that I've got (metaphorical) scars.

hack/generator/pkg/astmodel/type_walker.go
matthchr: Fixed

hack/generator/pkg/astmodel/type_walker.go
matthchr: Added for all but the last one coming from `AddWithEqualityCheck` (which will already have enough details). Definitely agree it adds clarity to add some wrapping for the cases where we're calling user-provided functions though, as that can help them understand which function of theirs was failing

hack/generator/pkg/astmodel/type_walker.go
matthchr: I want exactly a `TypeName`. As mentioned in the comment on line 46:
> // The provided visitor VisitTypeName function must return a TypeName and VisitObjectType must return an ObjectType or calls to Walk will panic.

hack/generator/pkg/astmodel/type_walker.go
matthchr: I dropped the `Be` and just renamed it to `shouldRemove`.

hack/generator/pkg/astmodel/types.go
matthchr: I'm definitely open to a better name -- do you have suggestions? `Add` is obviously taken. We could make it `AddAllowDuplicates`, or maybe something like `AddStructural` (not sure I love the second one myself).

hack/generator/pkg/astmodel/type_walker.go
theunrepentantgeek: Thanks.

hack/generator/pkg/astmodel/type_walker.go
theunrepentantgeek: Fair enough.

theunrepentantgeek (approved):

PR #425, opened 2021-03-29
Default controller ENVTEST and RECORD_REPLAY to 1

matthchr:
This means that just running a controller test by default does what you probably want. Other modes are still supported by explicitly turning ENVTEST or RECORD_REPLAY off.


babbageclunk (approved):
Nice!

Porges (approved):

Porges:
Good idea!

PR #426, opened 2021-03-29
Autogenerate RBAC kubebuilder annotation

matthchr:


babbageclunk (approved):

hack/generator/pkg/astmodel/resource.go
```diff
@@ -312,6 +314,20 @@ func (resource *ResourceType) AsDeclarations(codeGenerationContext *CodeGenerati
...
+		lpr, ok := declContext.Name.PackageReference.AsLocalPackage()
+		if !ok {
+			panic(fmt.Sprintf("expected resource package reference to be local: %q", declContext.Name))
+		}
+		group := strings.ToLower(lpr.Group() + GroupSuffix)
+		resourceName := strings.ToLower(declContext.Name.Plural().Name())
+
+		astbuilder.AddComment(&comments, fmt.Sprintf("// +kubebuilder:rbac:groups=%s,resources=%s,verbs=get;list;watch;create;update;patch;delete", group, resourceName))
+		astbuilder.AddComment(&comments, fmt.Sprintf("// +kubebuilder:rbac:groups=%s,resources={%s/status,%s/finalizers},verbs=get;update;patch", group, resourceName, resourceName))
+		comments = append(comments, "\n") // Force a newline just to make things a bit prettier
```
Porges, line 328: I actually think this is needed, semantically. The kubebuilder comment _must_ apply to the file/module and not the type.

Porges: There's a similar thing with the webhooks. I called the things on the interface that apply to the file/module an "annotation" but maybe should change how it works…

babbageclunk: Ooh, good point.

matthchr: Ah you may be right -- let me test and see and then if it's true I'll make sure to make a clear note.

matthchr: Yep, you're right

PR #427, opened 2021-03-30
Simplify property conversions

theunrepentantgeek:
Handle assignments _**to optional**_ destination fields once, instead of requiring each kind of property conversion to handle it independently. This halves the number of conversion factory methods we need.

Also

* Modify both Array and Map generation to create a local shadow of the loop variable to avoid aliasing errors during conversions
* Modify handling of optional source fields to create local variables for clarity
* Change warnings to level 3 logs to reduce noise
* Fix bug where potential suffixes for local variables were not being used
* Introduce `astbuilder.Statements()` for building `[]dst.Stmt` from a sequence of statements

### DIFF
--- a/hack/generator/pkg/astmodel/storage_conversion_factories.go
+++ b/hack/generator/pkg/astmodel/storage_conversion_factories.go
@@ -15,8 +15,9 @@ import (
 )
 
 // StorageTypeConversion generates the AST for a given conversion.
-// reader is an expression to read the original value
-// writer is a function that creates one or more statements to write the converted value
+// reader is an expression to read the original value.
+// writer is a function that accepts an expression for reading a value and creates one or more
+// statements to write that value.
 // Both of these might be complex expressions, possibly involving indexing into arrays or maps.
 type StorageTypeConversion func(reader dst.Expr, writer func(dst.Expr) []dst.Stmt, generationContext *CodeGenerationContext) []dst.Stmt
 
@@ -234,6 +235,8 @@ func assignPrimitiveTypeFromOptionalPrimitiveType(
 //
 // <arr> := make([]<type>, len(<reader>))
 // for <index>, <value> := range <reader> {
+//     // Shadow the loop variable to avoid aliasing
+//     <value> := <value>
 //     <arr>[<index>] := <value> // Or other conversion as required
 // }
 // <writer> = <arr>
--- a/hack/generator/pkg/astmodel/storage_conversion_factories.go
+++ b/hack/generator/pkg/astmodel/storage_conversion_factories.go
@@ -350,7 +350,7 @@ func assignArrayFromArray(
 			token.DEFINE,
 			astbuilder.MakeList(dstArray.AsType(generationContext), astbuilder.CallFunc("len", reader)))
 
-		assignToElement := func(expr dst.Expr) []dst.Stmt {
+		writeToElement := func(expr dst.Expr) []dst.Stmt {
 			return []dst.Stmt{
 				astbuilder.SimpleAssignment(
 					&dst.IndexExpr{
@@ -362,10 +362,18 @@ func assignArrayFromArray(
 			}
 		}
 
-		loopBody := conversion(
+		avoidAliasing := astbuilder.SimpleAssignment(dst.NewIdent(itemId), token.DEFINE, dst.NewIdent(itemId))
+		avoidAliasing.Decs.Start.Append("// Shadow the loop variable to avoid aliasing")
+		avoidAliasing.Decs.Before = dst.NewLine
+
+		loopBody := []dst.Stmt{
+			avoidAliasing,
+		}
+
+		loopBody = append(loopBody, conversion(
 			dst.NewIdent(itemId),
-			assignToElement,
-			generationContext)
+			writeToElement,
+			generationContext)...)
 
 		assign := writer(dst.NewIdent(tempId))
 
@@ -446,11 +454,19 @@ func assignMapFromMap(
 			}
 		}
 
-		loopBody := conversion(
+		avoidAliasing := astbuilder.SimpleAssignment(dst.NewIdent(itemId), token.DEFINE, dst.NewIdent(itemId))
+		avoidAliasing.Decs.Start.Append("// Shadow the loop variable to avoid aliasing")
+		avoidAliasing.Decs.Before = dst.NewLine
+
+		loopBody := []dst.Stmt{
+			avoidAliasing,
+		}
+
+		loopBody = append(loopBody, conversion(
 			dst.NewIdent(itemId),
 			assignToItem,
 			generationContext,
-		)
+		)...)
 
 		assign := writer(dst.NewIdent(tempId))
 
--- a/hack/generator/pkg/astmodel/storage_conversion_factories.go
+++ b/hack/generator/pkg/astmodel/storage_conversion_factories.go
@@ -160,7 +160,7 @@ func assignOptionalPrimitiveTypeFromPrimitiveType(
 		return nil
 	}
 
-	copyVar := destinationEndpoint.CreateLocal("Value")
+	copyVar := destinationEndpoint.CreateLocal("", "Value")
 
 	return func(reader dst.Expr, writer dst.Expr, generationContext *CodeGenerationContext) []dst.Stmt {
 		return []dst.Stmt{
@@ -272,7 +272,7 @@ func assignOptionalPrimitiveTypeFromOptionalPrimitiveType(
 		return nil
 	}
 
-	copyVar := destinationEndpoint.CreateLocal("Value")
+	copyVar := destinationEndpoint.CreateLocal("", "Value")
 
 	return func(reader dst.Expr, writer dst.Expr, ctx *CodeGenerationContext) []dst.Stmt {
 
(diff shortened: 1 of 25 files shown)


hack/generator/pkg/astmodel/storage_conversion_factories.go
```diff
@@ -342,36 +263,40 @@ func assignArrayFromArray(
...
+						X:     dst.NewIdent(tempId),
+						Index: dst.NewIdent(indexId),
+					},
+					token.ASSIGN,
+					expr),
+			}
+		}
 
-		assign := astbuilder.SimpleAssignment(writer, token.ASSIGN, dst.NewIdent(tempId))
+		avoidAliasing := astbuilder.SimpleAssignment(dst.NewIdent(itemId), token.DEFINE, dst.NewIdent(itemId))
```
babbageclunk, line 289: Add this to the example code in the doc comment?

hack/generator/pkg/astmodel/storage_conversion_factories.go
```diff
@@ -551,186 +488,41 @@ func assignEnumTypeFromOptionalEnumType(
...
-				token.DEFINE,
-				astbuilder.CallFunc(dstName.name, reader)),
+		// Need to check for null and only assign if we have a value
+		cond := astbuilder.NotEqual(reader, dst.NewIdent("nil"))
 
-			astbuilder.SimpleAssignment(
-				writer,
-				token.ASSIGN,
-				astbuilder.AddrOf(dst.NewIdent(copyVar))),
+		writeLocal := func(expr dst.Expr) []dst.Stmt {
```
babbageclunk, line 500: I think this is the same as the other writeLocal, isn't it?

babbageclunk (approved):
Nice - I think the code is clearer as well as being shorter! Any chance you can do something similar with the FromOptional cases?

hack/generator/pkg/astmodel/storage_conversion_factories.go
```diff
@@ -77,75 +73,66 @@ func createTypeConversion(
...
+	if !dstIsOpt {
+		// Destination is not optional
 		return nil
 	}
 
-	// Require destination to be a primitive type
-	dstPrim, dstOk := AsPrimitiveType(destinationEndpoint.Type())
-	if !dstOk {
+	// Require a conversion between the wrapped type and our source
+	dstEp := destinationEndpoint.WithType(dstOpt.element)
```
matthchr, line 94: minor: `dstEp` is just an abbreviation for a variable that you already have (`destinationEndpoint`). Wondering if this would be clearer as `innerEndpoint`, `nestedEndpoint` or `elementEndpoint`? Possibly sprinkle a `destination` in there too for `destinationElementEndpoint`? Then feel free to abbreviate as appropriate if desired?

hack/generator/pkg/astmodel/storage_conversion_factories.go
matthchr, line 289: You mentioned in the PR description: 
> Modify both Array and Map generation to create a local shadow of the loop variable to avoid aliasing errors during conversions

Can you expand on how exactly that becomes an issue? My understanding was that as long as you're not doing any capturing lambdas there's no need to shadow. I looked through the golden files tests and I didn't find any instance where you actually needed to do a shadow (although I'm not a compiler so maybe I missed it).

It does seem to result in a decent amount of "awkward" code where the assignment is clearly not required, such as:
```
// Shadow the loop variable to avoid aliasing
nastyValue1 := nastyValue1
nasty := nastyValue1
```

matthchr (commented):
Definitely looks like an improvement (and seems to match how you were solving the problem for List/Map types?)

Had a few minor questions

hack/generator/pkg/astmodel/storage_conversion_factories.go
```diff
@@ -77,75 +73,66 @@ func createTypeConversion(
...
+	conversion, _ := createTypeConversion(sourceEndpoint, dstEp, conversionContext)
```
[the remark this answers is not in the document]
theunrepentantgeek: I tried - but didn't manage to handle all the edge cases properly. Might be able to do this later on though.

hack/generator/pkg/astmodel/storage_conversion_factories.go
```diff
@@ -16,9 +16,9 @@ import (
 
 // StorageTypeConversion generates the AST for a given conversion.
 // reader is an expression to read the original value
-// writer is an expression to write the converted value
+// writer is a function that creates one or more statements to write the converted value
```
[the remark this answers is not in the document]
theunrepentantgeek: I'll update the comments to make this clearer.

hack/generator/pkg/astmodel/storage_conversion_factories.go
```diff
@@ -77,75 +73,66 @@ func createTypeConversion(
...
+	//copyVar := destinationEndpoint.CreateLocal("", "Temp")
 
-	return func(reader dst.Expr, writer dst.Expr, generationContext *CodeGenerationContext) []dst.Stmt {
-		return []dst.Stmt{
-			astbuilder.SimpleAssignment(writer, token.ASSIGN, reader),
+	return func(reader dst.Expr, writer func(dst.Expr) []dst.Stmt, generationContext *CodeGenerationContext) []dst.Stmt {
+		// Create a writer that takes the address of the passed expression
+		// Note that we are dependent on any wrapping conversion to ensure no aliasing occurs
+		addrOfWriter := func(expr dst.Expr) []dst.Stmt {
+			return writer(astbuilder.AddrOf(expr))
```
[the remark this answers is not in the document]
theunrepentantgeek: The writer is a func because it can be more complicated than just an assignment statement (e.g. for later conversions it can be a call). 

For example, in PR #429, I modify the writer here to check to see whether the `expr` passed in is a local variable - can't take the address of an arbitrary expression, so if `expr` is _not_ a local, we introduce one.

hack/generator/pkg/astmodel/storage_conversion_factories.go
```diff
@@ -211,100 +195,37 @@ func assignPrimitiveTypeFromOptionalPrimitiveType(
...
-		readValue.Decs.Start.Append("// Copy to a local to avoid aliasing")
-		readValue.Decs.Before = dst.NewLine
+		writeLocal := func(expr dst.Expr) []dst.Stmt {
+			return []dst.Stmt{
+				astbuilder.SimpleAssignment(
+					dst.NewIdent(local),
+					token.ASSIGN,
+					expr),
+			}
+		}
```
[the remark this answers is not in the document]
theunrepentantgeek: Because it's passed as a `writer`, this must have the signature `func(dst.Expr) []dst.Stmt`.

hack/generator/pkg/astmodel/storage_conversion_factories.go
theunrepentantgeek: If you don't shadow the loop variable and take it's address, dereferencing the pointer will always see the current value.

In PR #429 I modify the AddressOff use to automatically introduce the local if one is needed, so the awkwardness of the code goes away.

I could have also removed the automatic shadowing of the array & map code generation, but I like how it makes the code generation more robust. If a future maintainer inadventently directly takes the address of the loop variable, it will still be safe.

hack/generator/pkg/astmodel/storage_conversion_factories.go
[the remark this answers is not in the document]
theunrepentantgeek: Done.

hack/generator/pkg/astmodel/storage_conversion_factories.go
```diff
@@ -77,75 +73,66 @@ func createTypeConversion(
...
+	conversion, _ := createTypeConversion(sourceEndpoint, dstEp, conversionContext)
+	if conversion == nil {
 		return nil
 	}
 
-	// Require both properties to have the same primitive type
-	if !srcPrim.Equals(dstPrim) {
-		return nil
-	}
+	//copyVar := destinationEndpoint.CreateLocal("", "Temp")
```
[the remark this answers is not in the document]
theunrepentantgeek: Deleted. There are changes here in my followup PR that make things more robust.

hack/generator/pkg/astmodel/storage_conversion_factories.go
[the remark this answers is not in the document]
theunrepentantgeek: Which I guess means there is something trickier going on. 😁

hack/generator/pkg/astmodel/storage_conversion_factories.go
theunrepentantgeek: Done.

hack/generator/pkg/astmodel/storage_conversion_factories.go
theunrepentantgeek: Yes, but again it needs to be a local func because of the constrained signature.

hack/generator/pkg/astmodel/storage_conversion_factories.go
theunrepentantgeek: I've made a pass through favouring longer variable names rather than contractions like `source` => `src`, `destination` => `dst` and so on.

PR #428, opened 2021-03-30
Rename AddWithEqualityCheck to AddAllowDuplicates

matthchr:
- Other Add methods also now have the "New" suffix to match.
  - AddWithEqualityCheck renamed to Add.


matthchr:
I was sold on this and after having done it I am less sold again... curious at others thoughts

matthchr:
As discussed elsewhere, what about leaving `Add` as `Add` and making the new method `AddAllowDuplicates` or `AddAllowCollision`?

Porges (approved):

theunrepentantgeek (approved):
Looks good enough to me. I can't think of a better name.

PR #429, opened 2021-03-31
Support conversions between Enums and primitive types

theunrepentantgeek:
Introduces conversion compatibility between enum-typed properties and primitive-typed properties as long as the primitive type is the base type for the enumeration. Includes golden tests to verify the conversions generate the expected code

Other changes

* Eliminated some unrequired intermediate variables
* Ensured variable shadowing within loops to avoid aliasing
* If a local variable matching the name of the property has already been used, will use a suffix based on the name of the enumeration
* Improves some of the errors to include type information about the attempted conversion
* Changed the technique used to generate locals for map & list iteration, reducing conflicts

_**Note**_: Some of the detected file renames are _**not correct**_

~Requires #427 (Simplify Property Conversions) to be merged first~ PR #427 has now been merged.


hack/generator/pkg/astmodel/testdata/ConvertBetweenEnumAndBaseType.golden
```diff
@@ -0,0 +1,30 @@
...
+
+	// No error
+	return nil
+}
+
+// ConvertToVNext populates the provided destination Person from our Person
+func (person Person) ConvertToVNext(destination *vNext.Person) error {
+
+	// sku
+	destination.sku = Bucket(person.sku)
```
babbageclunk, line 26: Should `Bucket` be `vNext.Bucket`?

hack/generator/pkg/astmodel/testdata/ConvertBetweenOptionalEnumAndOptionalBaseType.golden
```diff
@@ -0,0 +1,43 @@
...
+func (person Person) ConvertFromVNext(source *vNext.Person) error {
+
+	// sku
+	var sku string
+	if source.sku != nil {
+		sku = string(*source.sku)
+	} else {
+		sku = ""
+	}
+	person.sku = &sku
```
babbageclunk, line 22: If `source.sku` is nil then `person.sku` should be nil rather than `&""`, right?

hack/generator/pkg/astmodel/testdata/ConvertBetweenOptionalEnumAndOptionalBaseType.golden
```diff
@@ -0,0 +1,43 @@
...
+
+	// sku
+	var sku string
+	if person.sku != nil {
+		sku = *person.sku
+	} else {
+		sku = ""
+	}
+	skuTemp := Bucket(person.sku)
+	destination.sku = &skuTemp
```
babbageclunk, line 39: Similarly here - `destination.sku` should be nil if `person.sku` is nil.

babbageclunk (changes requested):
Mostly looks good but I think the optional -> optional conversions are wrong: destination should be nil if source is nil. 

I'm sorry, I think I missed this in the last change, because I can see the same problem in unchanged files as well.

theunrepentantgeek:
> Mostly looks good but I think the optional -> optional conversions are wrong: destination should be nil if source is nil.

In PR #433 I change them so they are. 

To illustrate, here's an excerpt from `SetOptionalStringFromOptionalString.golden` from that PR:

```
func (person Person) ConvertFromVNext(source *vNext.Person) error {

	// name
	if source.name != nil {
		name := *source.name
		person.name = &name
	} else {
		person.name = nil
	}

	// No error
	return nil
}
```

hack/generator/pkg/astmodel/storage_conversion_factories.go
```diff
@@ -98,17 +101,80 @@ func assignToOptionalType(
...
 		}
 
 		return conversion(reader, addrOfWriter, generationContext)
 	}
 }
 
+// assignToEnumerationType will generate a conversion where the destination is an enumeration if
+// the source is type compatible with the base type of the enumeration
+//
+// <destination> = &<source>
```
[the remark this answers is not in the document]
theunrepentantgeek: No, it's not - fixed.

hack/generator/pkg/astmodel/testdata/ConvertBetweenEnumAndBaseType.golden
theunrepentantgeek: Yes - the test was using the wrong enum when creating the property for testing (was using one from the vCurrent package instead of the vNext package). Fixed.

hack/generator/pkg/astmodel/testdata/ConvertBetweenOptionalEnumAndOptionalBaseType.golden
theunrepentantgeek: Yes - this is corrected in PR #433. To illustrate:

```
func (person Person) ConvertFromVNext(source *vNext.Person) error {

	// sku
	if source.sku != nil {
		sku := string(*source.sku)
		person.sku = &sku
	} else {
		person.sku = nil
	}

	// No error
	return nil
}
```

hack/generator/pkg/astmodel/testdata/ConvertBetweenOptionalEnumAndOptionalBaseType.golden
theunrepentantgeek: Ditto.

babbageclunk (approved):
Looks great! The updated nil-handling code from #433 is much better thanks.

PR #431, opened 2021-04-02
Refactor generic controller

matthchr:
- Make GenericController support both handcrafted and
    autogenerated resources by introducing an additional
    Reconciler interface.


matthchr:
I had a bit of time to hack on this problem that had been nagging at me for a while.
This doesn't actually solve the "support handcrafted resources" problem quite yet, as we still need a way to associate a particular reconciler with a GVK, but at least it gives us some basic interfaces that we can expand, and it makes sure that we separate the code generated reconciliation (all about deployments) from the controller itself which should be agnostic of "actual creation mechanism"

There are still lots of problems:
1. The `genruntime.MetaObject` type may not actually be the "correct" shape for all resources as it's sorta tied/related to the `ArmSpec` stuff which is obviously specific to an ARM resource (as opposed to a SQL user or something), but right now the controller expects that everything looks like `genruntime.MetaObject` and not something more generic.
2. Some stuff which is currently being done in the new `AzureDeploymentReconciler` probably should be pulled up into the `GenericController`, so that if/when we do have handcrafted resources we don't duplicate some of the state management that they're all going to have to do.

really my intent with this was just a code-moving exercise to at least make the separation of the controller concept from the notion of Azure deployments more official, so that we don't accidentally introduce further coupling between them down the road. We should follow this up with more PRs to clean up the interface further in the future.

hack/generated/pkg/genruntime/base_types.go
```diff
@@ -17,6 +20,14 @@ type MetaObject interface {
 	KubernetesResource
 }
 
+type Reconciler interface { // TODO: Sorta awkward interface name
+	CreateOrUpdate(ctx context.Context) (ctrl.Result, error)
+
+	Delete(ctx context.Context) (ctrl.Result, error)
+
+	// TODO: Do we want a "Diff"? Or is the expectation that logic just exists in CreateOrUpdate?
```
theunrepentantgeek, line 28: I don't think so - at least, not until we have a proven need for it. Keeping the interface as minimal as possible is useful.

hack/generated/pkg/reconcilers/azure_deployment_reconciler.go
```diff
@@ -0,0 +1,869 @@
...
+// TODO: Do we actually want this at the controller level or this level?
+const GenericControllerFinalizer = "generated.infra.azure.com/finalizer"
+
+type CreateOrUpdateAction string
+
+const (
+	CreateOrUpdateActionNoAction          = CreateOrUpdateAction("NoAction")
+	CreateOrUpdateActionManageOwnership   = CreateOrUpdateAction("ManageOwnership")
+	CreateOrUpdateActionBeginDeployment   = CreateOrUpdateAction("BeginDeployment")
+	CreateOrUpdateActionMonitorDeployment = CreateOrUpdateAction("MonitorDeployment")
```
theunrepentantgeek, line 64: Given that not everything is "deployed" (e.g. database users, network permissions, etc) perhaps these should be "BeginUpdate" and "MonitorUpdate" ?

hack/generated/pkg/reconcilers/azure_deployment_reconciler.go
```diff
@@ -0,0 +1,869 @@
...
+		recorder:             eventRecorder,
+		ARMClient:            armClient,
+		CreateDeploymentName: createDeploymentName,
+		KubeClient:           kubeClient,
+		ResourceResolver:     resourceResolver,
+	}
+}
+
+func (r *AzureDeploymentReconciler) CreateOrUpdate(ctx context.Context) (ctrl.Result, error) {
+	action, actionFunc, err := r.DetermineCreateOrUpdateAction()
```
theunrepentantgeek, line 111: Since both `action` and `actionFunc` are returned from the same factory method, we shouldn't need to pass the `action` back in again, the func can already have that value captured. 

We do need to return `action` from the method, so that we can do the right logging further on, but that's all.

hack/generated/pkg/reconcilers/azure_deployment_reconciler.go
```diff
@@ -0,0 +1,869 @@
...
+		r.log.Error(err, "Error during CreateOrUpdate", "action", action)
+		r.recorder.Event(r.obj, v1.EventTypeWarning, "CreateOrUpdateActionError", err.Error())
+		return ctrl.Result{}, err
+	}
+
+	return result, nil
+}
+
+func (r *AzureDeploymentReconciler) Delete(ctx context.Context) (ctrl.Result, error) {
+	action, actionFunc, err := r.DetermineDeleteAction()
```
theunrepentantgeek, line 130: Ditto here - I think we can drop the `action` parameter for the delete action.

hack/generated/pkg/reconcilers/azure_deployment_reconciler.go
```diff
@@ -0,0 +1,869 @@
...
+func (r *AzureDeploymentReconciler) GetResourceProvisioningState() armclient.ProvisioningState {
+	return armclient.ProvisioningState(r.obj.GetAnnotations()[ResourceStateAnnotation])
+}
+
+func (r *AzureDeploymentReconciler) GetDeploymentId() (string, bool) {
+	id, ok := r.obj.GetAnnotations()[DeploymentIdAnnotation]
+	return id, ok
+}
+
+func (r *AzureDeploymentReconciler) GetDeploymentIdOrDefault() string {
```
theunrepentantgeek, line 162: Why do we need this method when we have `GetDeploymentId()`? I thought idiomatic Go was to spurn helper methods like this.

theunrepentantgeek (approved):
I've assumed most of the code is just relocated, and thus didn't need a deep read.

hack/generated/pkg/reconcilers/azure_deployment_reconciler.go
matthchr: This is the `deployment_reconciler` though so in this context I think it's ok/correct. The intent was for other things (created through something that isn't the Azure `deployment` ARM infrastructure) to have their own reconcilers, which is why I changed the `generic_controller` to operate at the `reconciler` level and abstracted all of this deployment specific stuff out.

hack/generated/pkg/reconcilers/azure_deployment_reconciler.go
matthchr: Agree - there's a tight coupling between methods and the corresponding action enum as well.

Have fixed this.

hack/generated/pkg/reconcilers/azure_deployment_reconciler.go
matthchr: Fixed

hack/generated/pkg/reconcilers/azure_deployment_reconciler.go
matthchr: Yeah in this instance at least this method isn't buying us much - it'd be clearer to just use `GetDeploymentId`, I've made the change.

Issue #353, opened 2020-12-14
Feature: Store additional properties in property bags


theunrepentantgeek:
Extend creation of the storage types to introduce a property bag that can be used to stash additional information where it's not supported on the target of the conversion. 

Modify the conversions between storage types to use the property bag where one end of the conversion doesn't have the expected property. 

* When converting to the type without the property, store the value in the property bag
* When converting from the type without the property, read the value from the property bag

Prerequisite: #352

PR #433, opened 2021-04-12
Add support for aliases of primitive types

theunrepentantgeek:
Adds conversion support properties using types that are just names (and validation rules) for simple primitive types, including conversion to and from those primitive types where required.

Adds support for assignments between properties that are `KnownResourceReference`.
 
A single handler for every case where we are copying _**from**_ an optional type, reducing the number of handlers required, helping to avoid a combinatorial explosion of handlers.

Also

*  Changes to avoid use of local variables in generated code in some cases
*  Enhance `astbuilder.Statements()` so it filters out nil values
*  Fixes a bug in `assignToEnumerationType` where the expression passed wasn't being used
*  Add clone methods for `KnownResourceReference` and `ResourceReference`
*  A bunch of code gardening improvements to logging, comments and code clarity
*  Enhances zeroValue() to support more than just primitive types

~Requires #429 to be merged first.~ PR #429 has been merged.


hack/generator/pkg/astmodel/storage_conversion_factories.go
```diff
@@ -125,6 +128,83 @@ func assignToOptionalType(
...
+
+		// If we have a value, need to convert it to our destination type
+		writeActualValue := conversion(
+			astbuilder.Dereference(actualReader),
+			writer,
+			generationContext)
+
+		writeZeroValue := writer(
+			&dst.BasicLit{
+				Value: zeroValue(destinationEndpoint.Type(), conversionContext.types),
```
theunrepentantgeek, line 195: Passing `types` to `zeroValue()` feels odd as it's only needed for the case where the type is a TypeName because we need to resolve it to find out whether it's an enum, alias, or something else.

It also feels odd to have `zeroValue()` here as a global func; I'm now wondering whether it has a natural place to live on `Types` - nicely scoped, easy to call, and has access to the set of available types if really needed.

hack/generator/pkg/astmodel/storage_conversion_factories.go
```diff
@@ -913,21 +850,52 @@ func createTypeDeclaration(name TypeName, generationContext *CodeGenerationConte
...
+			return "0"
+		case UInt32Type:
+			return "0"
+		case UInt64Type:
+			return "0"
+		case BoolType:
+			return "false"
+		default:
+			panic(fmt.Sprintf("unexpected primitive type %q", t.String()))
+		}
```
babbageclunk, line 897: It would be clearer to extract this to `primitiveZeroValue` (and replace the recursive call on 874 with a call to it as well).

hack/generator/pkg/astmodel/testdata/SetMapOfOptionalFromMapOfRequired.golden
```diff
@@ -33,13 +33,11 @@ func (person Person) ConvertToVNext(destination *vNext.Person) error {
 	for ratingKey, ratingValue := range person.ratings {
 		// Shadow the loop variable to avoid aliasing
 		ratingValue := ratingValue
-		var rating int
 		if ratingValue != nil {
-			rating = *ratingValue
+			ratingMap[ratingKey] = *ratingValue
 		} else {
-			rating = 0
+			ratingMap[ratingKey] = 0
```
babbageclunk, line 39: This is weird, but I *think* it's right - the source map has a key that maps to nil, that should convert  to a key with a zero value. Someone could argue that it should convert to a missing key?

babbageclunk (approved):
Looks awesome - the optional handling is really nice. Mostly comments about type switches. :)

hack/generator/pkg/astmodel/testdata/SetMapOfOptionalFromMapOfRequired.golden
theunrepentantgeek: I'm pretty sure this is right - we're maintaining the fact that the key value did exist in the original map.

hack/generator/pkg/astmodel/storage_conversion_factories.go
theunrepentantgeek: I changed my mind - have moved it onto `Type` and it seems to have come out cleanly.

hack/generator/pkg/astbuilder/builder.go
```diff
@@ -355,7 +355,10 @@ func EnsureStatementBlock(statement dst.Stmt) *dst.BlockStmt {
...
 	var result []dst.Stmt
 	for _, s := range statements {
-		if stmt, ok := s.(dst.Stmt); ok {
+		if s == nil {
+			// Skip nils
+			continue
+		} else if stmt, ok := s.(dst.Stmt); ok {
 			result = append(result, stmt)
 		} else if stmts, ok := s.([]dst.Stmt); ok {
 			result = append(result, stmts...)
```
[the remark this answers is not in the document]
theunrepentantgeek: Done.

hack/generator/pkg/astmodel/storage_conversion_factories.go
```diff
@@ -125,6 +128,83 @@ func assignToOptionalType(
...
+			cacheOriginal = nil
+			actualReader = reader
+		} else {
+			// Something else, so we cache the original
+			cacheOriginal = astbuilder.SimpleAssignment(
+				dst.NewIdent(local),
+				token.DEFINE,
+				reader)
+			actualReader = dst.NewIdent(local)
+		}
```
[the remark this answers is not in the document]
theunrepentantgeek: Nice change, emphasizes that the treatment of `*dst.Ident` and `*dst.SelectorExpr` is the same. Done.

hack/generator/pkg/astmodel/storage_conversion_factories.go
theunrepentantgeek: I simplified this in a different way.

hack/generator/pkg/astmodel/allof_type.go
```diff
@@ -119,6 +119,12 @@ func (allOf AllOfType) AsDeclarations(_ *CodeGenerationContext, _ DeclarationCon
 	panic(errors.New(allOfPanicMsg))
 }
 
+// AsZero always panics; AllOf cannot be represented by the Go AST and must be
+//// lowered to an object type
```
babbageclunk, line 123: minor: double //

hack/generator/pkg/astmodel/map_type.go
```diff
@@ -52,6 +53,11 @@ func (m *MapType) AsType(codeGenerationContext *CodeGenerationContext) dst.Expr
 	}
 }
 
+// AsZero renders an expression for the "zero" value of a map by calling make()
+func (m *MapType) AsZero(_ Types, ctx *CodeGenerationContext) dst.Expr {
+	return astbuilder.CallFunc("make", m.AsType(ctx))
```
babbageclunk, line 58: The zero value of a map should be nil, rather than an empty map.

hack/generator/pkg/astmodel/object_type.go
```diff
@@ -183,6 +183,13 @@ func (objectType *ObjectType) AsType(codeGenerationContext *CodeGenerationContex
 	}
 }
 
+// AsZero renders an expression for the "zero" value of the type
+// types allows TypeName to resolve to the underlying type
+// ctx allows current imports to be correctly identified where needed
+func (objectType *ObjectType) AsZero(types Types, ctx *CodeGenerationContext) dst.Expr {
+	panic("cannot create a zero value for an object type")
```
babbageclunk, line 190: Why is this the case? Shouldn't the zero value of `SomeType` be `SomeType{}`? Basically it should be what you'd get by doing `var thing SomeType`.

hack/generator/pkg/astmodel/primitive_type.go
```diff
@@ -12,28 +12,29 @@ import (
...
-var FloatType = &PrimitiveType{"float64"}
+var FloatType = &PrimitiveType{"float64", "0"}
 
 // BoolType represents the Go bool type
-var BoolType = &PrimitiveType{"bool"}
+var BoolType = &PrimitiveType{"bool", "false"}
 
 // AnyType represents the root Go interface type, permitting any object
-var AnyType = &PrimitiveType{"interface{}"}
+var AnyType = &PrimitiveType{"interface{}", "&struct{}{}"}
```
babbageclunk, line 37: This should be nil. (Although I guess it's academic since there shouldn't be any `AnyType`s left when this would happen?)

hack/generator/pkg/astmodel/storage_conversion_factories.go
```diff
@@ -849,53 +843,3 @@ func createTypeDeclaration(name TypeName, generationContext *CodeGenerationConte
...
-			return zeroValue(enumType.baseType, types)
-		}
-
-		primitiveType, isPrimitive := AsPrimitiveType(actualType)
-		if isPrimitive {
-			return zeroValue(primitiveType, types)
-		}
-
-		// Otherwise default to an empty instantiation
-		return fmt.Sprintf("%s{}", name.Name())
```
babbageclunk: This is doing the right thing for ObjectType, I think. (Although does it need a package qualification?)

hack/generator/pkg/astmodel/type_name.go
```diff
@@ -66,6 +66,26 @@ func (typeName TypeName) AsType(codeGenerationContext *CodeGenerationContext) ds
...
+	actualType, err := types.FullyResolve(typeName)
+	if err != nil {
+		// This should never happen
+		panic(err)
+	}
+
+	if _, isObject := AsObjectType(actualType); isObject {
+		// We reference an object type, so our zero value is an empty struct
+		return &dst.BasicLit{
+			Value: fmt.Sprintf("%s{}", typeName.Name()),
```
babbageclunk, line 81: Ahhhh, is this why `ObjectType.AsZero` panics, because you can't create the zero value of an unnamed struct?

hack/generator/pkg/astmodel/object_type.go
babbageclunk: I think I might understand this now.

babbageclunk (approved):
Those are some lovely type-switches! A couple of questions about specific zero values.

hack/generator/pkg/astmodel/testdata/SetMapOfOptionalFromMapOfRequired.golden
matthchr: Do we need to stash the "nil-ness" of a particular value in the bag and retrieve it on the way back? That seems tricky but would:
1. Maintain roundtripping exactly.
2. Push the default behavior to be that nil values are "dropped" (shunted to the bag).

Both nil values being dropped and nil values being the zero value of the type have a chance of being wrong (so we'd need to write an override in those cases I think?)

I think this same round-tripping problem appears for all optional types (so optional arrays and plain optional values also have the same problem?)

hack/generator/pkg/astmodel/optional_type.go
```diff
@@ -65,6 +65,14 @@ func (optional *OptionalType) AsType(codeGenerationContext *CodeGenerationContex
 	}
 }
 
+// AsZero renders an expression for the "zero" value of the type
+// by returning a literal "nil"
+func (optional *OptionalType) AsZero(types Types, ctx *CodeGenerationContext) dst.Expr {
```
matthchr, line 70: minor: Don't use `types` here so could be `_`

hack/generator/pkg/astmodel/type_name.go
matthchr: yeah agree, this makes sense now

hack/generator/pkg/astmodel/storage_conversion_factories.go
```diff
@@ -36,23 +35,25 @@ var typeConversionFactories []StorageTypeConversionFactory
...
 		assignPrimitiveTypeFromEnumType,
-		assignEnumTypeFromOptionalEnumType,
-		assignPrimitiveTypeFromOptionalEnumType,
 		// Complex object types
 		assignObjectTypeFromObjectType,
-		assignObjectTypeFromOptionalObjectType,
+		// Known types
+		assignKnownReferenceFromKnownReference,
 		// Meta-conversions
+		assignFromOptionalType, // Must go before assignToOptionalType
```
matthchr, line 52: might be worth noting why?

hack/generator/pkg/astmodel/testdata/SetMapOfOptionalFromMapOfRequired.golden
theunrepentantgeek: We don't (yet!) have a property bag, but it would certainly be possible to use it to tunnel extra information through. I've added this to Issue #353

theunrepentantgeek: Yeah, anytime there's a transition between optional and non-optional we have this problem. It occurs both ways in ARM APIs - optional things that become manditory, and mandatory things that become optional. Stashing additional information in the property bag sounds like a good solution.

hack/generator/pkg/astmodel/allof_type.go
theunrepentantgeek: Fixed.

hack/generator/pkg/astmodel/map_type.go
theunrepentantgeek: Fixed.

hack/generator/pkg/astmodel/object_type.go
theunrepentantgeek: The problem is that an ObjectType has no name - so we can't generate `SomeType{}`. Where the type has a name, the zero value is generated by `TypeName`. I've added a comment to make this more explicit.

hack/generator/pkg/astmodel/primitive_type.go
theunrepentantgeek: Academic, but let's fix anyway.

hack/generator/pkg/astmodel/storage_conversion_factories.go
theunrepentantgeek: Updated `TypeName.AsZero()` to use a package name if needed.

hack/generator/pkg/astmodel/type_name.go
theunrepentantgeek: I've added a comment to `ObjectType.AsZero()` to make this more clear.

hack/generated/pkg/genruntime/resource_reference.go
```diff
@@ -50,3 +50,19 @@ func LookupOwnerGroupKind(v interface{}) (string, string) {
 
 	return group, kind
 }
+
+// Copy makes an independent copy of the KnownResourceReference
+func (ref KnownResourceReference) Copy() KnownResourceReference {
```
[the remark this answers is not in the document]
theunrepentantgeek: Indeed. Done.

hack/generator/pkg/astmodel/array_type.go
```diff
@@ -40,6 +41,11 @@ func (array *ArrayType) AsType(codeGenerationContext *CodeGenerationContext) dst
 	}
 }
 
+// AsZero renders an expression for the "zero" value of the array by calling make()
+func (array *ArrayType) AsZero(_ Types, ctx *CodeGenerationContext) dst.Expr {
+	return astbuilder.CallFunc("make", array.AsType(ctx))
```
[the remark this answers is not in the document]
theunrepentantgeek: Yeah, I think you're right.

hack/generator/pkg/astmodel/optional_type.go
theunrepentantgeek: Fixed. I suspect our linting will complain if I don't.

hack/generator/pkg/astmodel/storage_conversion_factories.go
theunrepentantgeek: If we don't to `FromOptionalType` first, we don't generate the correct zero value for the destination type.

For example, instead of this correct code:

``` go
	// age
	if source.age != nil {
		age := Age(*source.age)
		person.age = &age
	} else {
		person.age = nil
	}
```

we generate this incorrect result:

``` go
	// age
	if source.age != nil {
		ageTemp := Age(*source.age)
		person.age = &ageTemp
	} else {
		ageTemp := 0
		person.age = &ageTemp
	}
```

I've updated the comment to give a clue why the order is important.

hack/generator/pkg/astmodel/testdata/SetMapOfOptionalFromMapOfRequired.golden
matthchr: Might be worth leaving a note in the code as well (TODO) just as a reminder to do this eventually, although I suppose eventually we're going to have roundtripping tests as well that would notice it in any case.

matthchr (approved):
Left a few more minor comments, but LGTM overall!

hack/generator/pkg/astmodel/testdata/SetMapOfOptionalFromMapOfRequired.golden
theunrepentantgeek: I've added this to Issue #353 so I think that's enough.

PR #434, opened 2021-04-14
Add some documentation on how type conversion generates the code

theunrepentantgeek:


babbageclunk (approved):
Nice trace through - it definitely makes the more complicated conversions a bit clearer.

PR #435, opened 2021-04-15
Debug descriptions for types

theunrepentantgeek:
The existing `String()` implementations on `Type` are insufficient for troubleshooting some issues because they stop at a `TypeName`, not showing what lies underneath.

Without the ability to add a `Types` parameter to `String()` to allow resolution of type names, I've added instead `WriteDebugDescription()`.

### DIFF
--- a/hack/generator/pkg/astmodel/allof_type.go
+++ b/hack/generator/pkg/astmodel/allof_type.go
@@ -159,3 +159,19 @@ func (allOf *AllOfType) String() string {
 
 	return fmt.Sprintf("(allOf: %s)", strings.Join(subStrings, ", "))
 }
+
+// DebugDescription adds a description of the current AnyOf type to the passed builder
+// builder receives the full description, including nested types
+// types is a dictionary for resolving named types
+func (allOf *AllOfType) WriteDebugDescription(builder *strings.Builder, types Types) {
+	builder.WriteString("AllOf[")
+	first := true
+	allOf.types.ForEach(func(t Type, ix int) {
+		if !first {
+			builder.WriteString("|")
+		}
+		t.WriteDebugDescription(builder, types)
+		first = false
+	})
+	builder.WriteString("]")
+}
--- a/hack/generator/pkg/astmodel/allof_type.go
+++ b/hack/generator/pkg/astmodel/allof_type.go
@@ -165,13 +165,11 @@ func (allOf *AllOfType) String() string {
 // types is a dictionary for resolving named types
 func (allOf *AllOfType) WriteDebugDescription(builder *strings.Builder, types Types) {
 	builder.WriteString("AllOf[")
-	first := true
 	allOf.types.ForEach(func(t Type, ix int) {
-		if !first {
+		if ix > 0 {
 			builder.WriteString("|")
 		}
 		t.WriteDebugDescription(builder, types)
-		first = false
 	})
 	builder.WriteString("]")
 }
--- a/hack/generator/pkg/astmodel/allof_type.go
+++ b/hack/generator/pkg/astmodel/allof_type.go
@@ -160,7 +160,7 @@ func (allOf *AllOfType) String() string {
 	return fmt.Sprintf("(allOf: %s)", strings.Join(subStrings, ", "))
 }
 
-// DebugDescription adds a description of the current AnyOf type to the passed builder
+// WriteDebugDescription adds a description of the current AnyOf type to the passed builder
 // builder receives the full description, including nested types
 // types is a dictionary for resolving named types
 func (allOf *AllOfType) WriteDebugDescription(builder *strings.Builder, types Types) {
--- a/hack/generator/pkg/astmodel/flagged_type.go
+++ b/hack/generator/pkg/astmodel/flagged_type.go
@@ -176,7 +176,7 @@ func (ft *FlaggedType) Unwrap() Type {
 	return ft.element
 }
 
-// DebugDescription adds a description of the current type to the passed builder
+// WriteDebugDescription adds a description of the current type to the passed builder
 // builder receives the full description, including nested types
 // types is a dictionary for resolving named types
 func (ft *FlaggedType) WriteDebugDescription(builder *strings.Builder, types Types) {
--- a/hack/generator/pkg/astmodel/flagged_type.go
+++ b/hack/generator/pkg/astmodel/flagged_type.go
@@ -175,3 +175,15 @@ func (ft *FlaggedType) String() string {
 func (ft *FlaggedType) Unwrap() Type {
 	return ft.element
 }
+
+// DebugDescription adds a description of the current type to the passed builder
+// builder receives the full description, including nested types
+// types is a dictionary for resolving named types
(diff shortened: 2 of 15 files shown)


hack/generator/pkg/astmodel/allof_type.go
```diff
@@ -159,3 +159,19 @@ func (allOf *AllOfType) String() string {
 
 	return fmt.Sprintf("(allOf: %s)", strings.Join(subStrings, ", "))
 }
+
+// DebugDescription adds a description of the current AnyOf type to the passed builder
```
matthchr, line 163: minor: Should be `WriteDebugDescription`?

hack/generator/pkg/astmodel/allof_type.go
```diff
@@ -159,3 +159,19 @@ func (allOf *AllOfType) String() string {
...
+// builder receives the full description, including nested types
+// types is a dictionary for resolving named types
+func (allOf *AllOfType) WriteDebugDescription(builder *strings.Builder, types Types) {
+	builder.WriteString("AllOf[")
+	first := true
```
matthchr, line 168: minor: You have an index in the `ForEach` func so you could just do away with this and check if `ix > 0` I think?

matthchr: This applies on other function implementations too but won't repeat comment there

hack/generator/pkg/astmodel/allof_type.go
```diff
@@ -159,3 +159,19 @@ func (allOf *AllOfType) String() string {
...
+	allOf.types.ForEach(func(t Type, ix int) {
+		if !first {
+			builder.WriteString("|")
+		}
+		t.WriteDebugDescription(builder, types)
```
matthchr, line 173: minor: Wondering if newlines would help any in displaying things?

hack/generator/pkg/astmodel/object_type.go
```diff
@@ -526,3 +527,10 @@ func extractEmbeddedTypeName(t Type) (TypeName, error) {
 
 	return TypeName{}, errors.Errorf("embedded property type must be TypeName, was: %T", t)
 }
+
+// DebugDescription adds a description of the current type to the passed builder
+// builder receives the full description, including nested types
+// types is a dictionary for resolving named types
+func (objectType *ObjectType) WriteDebugDescription(builder *strings.Builder, types Types) {
+	builder.WriteString("Object")
```
matthchr, line 535: Technically could print the properties and stuff here?

hack/generator/pkg/astmodel/type_name.go
```diff
@@ -166,3 +166,19 @@ func (typeName TypeName) Plural() TypeName {
...
+// DebugDescription adds a description of the current type to the passed builder
+// builder receives the full description, including nested types
+// types is a dictionary for resolving named types
+func (typeName TypeName) WriteDebugDescription(builder *strings.Builder, types Types) {
+	builder.WriteString(typeName.PackageReference.String())
+	builder.WriteString("/")
+	builder.WriteString(typeName.name)
+	builder.WriteString(":")
+
+	if definition, ok := types[typeName]; ok {
```
matthchr, line 179: Might want to examine the TypeName to determine if it is/isn't a `LocalPackageReference` and then only follow the reference if it is? If it's something pointing to an external type then you can instead print something that clarifies that (and avoid the `NOTDEFINED`?)

matthchr (approved):
LGTM -- left some nits.

hack/generator/pkg/astmodel/allof_type.go
theunrepentantgeek: Heh. I'm too used to C#'s **foreach**. Will fix.

hack/generator/pkg/astmodel/allof_type.go
theunrepentantgeek: They frequently end up embedded within an `error` message, so I suspect newlines would cause more problems than they fix.

hack/generator/pkg/astmodel/object_type.go
theunrepentantgeek: I considered it, but was hesitant about flooding the output with a LOT more content.

hack/generator/pkg/astmodel/type_name.go
theunrepentantgeek: Good idea.

hack/generator/pkg/astmodel/allof_type.go
theunrepentantgeek: Fixed.

hack/generator/pkg/astmodel/allof_type.go
theunrepentantgeek: Done.

hack/generator/pkg/astmodel/type_name.go
theunrepentantgeek: Done.

hack/generator/pkg/astmodel/flagged_type.go
```diff
@@ -175,3 +175,15 @@ func (ft *FlaggedType) String() string {
 func (ft *FlaggedType) Unwrap() Type {
 	return ft.element
 }
+
+// WriteDebugDescription adds a description of the current type to the passed builder
+// builder receives the full description, including nested types
+// types is a dictionary for resolving named types
+func (ft *FlaggedType) WriteDebugDescription(builder *strings.Builder, types Types) {
+	ft.element.WriteDebugDescription(builder, types)
+	for f := range ft.flags {
```
Porges, line 184: everything else is prefix so I think this should probably be as well

hack/generator/pkg/astmodel/type_test.go
```diff
@@ -0,0 +1,78 @@
...
+		{"MapOfSuitToAge", NewMapType(suit, age), "Map[local/test/v1/Suit:Enum[string:clubs|diamonds|hearts|spades]]local/test/v1/Age:int"},
+		{"FlaggedAge", armAge, "local/test/v1/Age:int[#arm]"},
+		{"FlaggedSuit", armSuit, "local/test/v1/Suit:Enum[string:clubs|diamonds|hearts|spades][#arm]"},
+		{"ErroredAge", erroredAge, "Error[local/test/v1/Age:int|boom|oh oh]"},
+	}
+
+	for _, c := range cases {
+		c := c
+		t.Run(c.name, func(t *testing.T) {
+			t.Parallel()
```
Porges, line 70: Parallelism probably isn't going to help with speed here

theunrepentantgeek: Yes, it marks the tests as able to be run in parallel - but isn't this also how you signal that the tests are supposed to be independent?

hack/generator/pkg/astmodel/flagged_type.go
theunrepentantgeek: I disagree. 😉

hack/generator/pkg/astmodel/type_test.go
Porges: I think it spins them off as separate go tasks so for really short-running tests it's not worth it

PR #439, opened 2021-04-20
Create TypeVisitorBuilder to make it easier to create TypeVisitors

theunrepentantgeek:
Introduces a new type `TypeVisitorBuilder` as a convenience for making instances of `TypeVisitor`. 

Functions provided for visits are permitted a variety of signatures, reducing the amount of noise in code.

Goal is to make consuming code a little more declarative - this works particularly well when functions with well chosen names are passed used.


hack/generator/pkg/astmodel/type_definition.go
```diff
@@ -136,15 +136,15 @@ func FileNameHint(name TypeName) string {
...
-	}
+	visitor := MakeTypeVisitor(
+		func(_ *TypeVisitor, ot *ObjectType, _ interface{}) (Type, error) {
+			rt, err := transform(ot)
+			if err != nil {
+				return nil, err
+			}
+			visited = true
+			return rt, nil
+		})
```
matthchr, line 147: This sorta seems _less_ clear than it was before (to me). Certainly if it's a clarity/simplicity improvement it's a minor one.

Do we really want to give up compile-time type safety of fields for what feels like a small net gain? (I suppose it's nice that the visitor isn't modified after creation but we didn't actually _forbid_ modification so it seems a small win).

Supporting this also results in additional somewhat involved code to do the reflection, as well as tests for that. Is that additional complexity (and risk, although it's been well mitigated with tests) worth it over the dumb but "obviously correct" way it was before?

matthchr: For example it seems like it would be easy to accidentally pass two functions that had the same signature and have the second overwrite the first, for example. Something that would be more difficult to do with the old property-assignment approach since `visitor.VisitTypeName = myVisitTypeName` where `myVisitTypeName` actually has an `ObjectType` signature won't compile. If you _did_ accidentally do:
```
visitor.VisitObjectType = myVisitObjectType
...
visitor.VisitObjectType = myVisitTypeName
```

it's at least clear-ish from a scan of the code that you've overwritten something.

hack/generator/pkg/codegen/pipeline_check_for_anytype.go
```diff
@@ -76,16 +76,20 @@ func checkForAnyType(description string, packages []string) PipelineStage {
 }
 
 func containsAnyType(theType astmodel.Type) bool {
+
 	var found bool
-	visitor := astmodel.MakeTypeVisitor()
-	visitor.VisitPrimitive = func(_ *astmodel.TypeVisitor, it *astmodel.PrimitiveType, _ interface{}) (astmodel.Type, error) {
+	detectAnyType := func(it *astmodel.PrimitiveType) astmodel.Type {
```
theunrepentantgeek, line 81: Rename to `detectUseOfAnyType`

hack/generator/pkg/codegen/pipeline_convert_allof_and_oneof_to_objects.go
```diff
@@ -33,10 +33,10 @@ func convertAllOfAndOneOfToObjects(idFactory astmodel.IdentifierFactory) Pipelin
 		"allof-anyof-objects",
 		"Convert allOf and oneOf to object types",
 		func(ctx context.Context, defs astmodel.Types) (astmodel.Types, error) {
-			visitor := astmodel.MakeTypeVisitor()
+			builder := astmodel.TypeVisitorBuilder{}
 
 			// the context here is whether we are selecting spec or status fields
-			visitor.VisitAllOfType = func(this *astmodel.TypeVisitor, it *astmodel.AllOfType, ctx interface{}) (astmodel.Type, error) {
+			builder.VisitAllOfType = func(this *astmodel.TypeVisitor, it *astmodel.AllOfType, ctx interface{}) (astmodel.Type, error) {
```
theunrepentantgeek, line 39: I considered giving these anonymous funcs some intention revealing names, but I wasn't 100% on what the methods did - @matthchr if you make some suggestions, I'll roll those in to place.

babbageclunk (approved):
I like this - the convenience wrapping of the visit callbacks makes simple uses much clearer. Neat configure trick in the tests too!

matthchr (commented):
I'm torn on this.

I like the immutable `TypeVisitor`.
I really like the nicely named functions and how they align with the assignment to tell a good story. Obviously this is only a convention but it's a good one.

I still don't like `interface{}` on `TypeVisitorBuilder` properties though. If I were a new person contributing to this repo the lack of typing there would probably throw me. I feel like if we're going to give up compiler-checked typing we should be getting a lot in return and I'm not sure that the tradeoff is worth it here:

What we get:
1. The ability to specify one of 3 flavors of function, saving some unneeded parameter specification or return type specification. This enables us to write less code and/or write less cluttered functions.

What we pay:
1. Lack of compile time validation ensuring that the `TypeVisitorBuilder` was pass valid parameters.
2. Significant extra complexity in implementation of `TypeVisitorBuilder`, requiring extra tests.

The complexity in particular is troublesome for me because we've traded the "common complexity" of having unused parameters or returning `nil` for errors for the "uncommon complexity" of lack of typing and reflection.  (By "common complexity" I just mean it's complexity that you commonly see elsewhere in Go and in our codebase here)

There are some alternative approaches floating through my head:
1. Keep immutability, keep the nicely named functions, entirely drop the ability to specify alternative function shapes (but gain strong typing and a guarantee there's no possible panic lurking in your `TypeVisitorBuilder` usage).
2. Keep immutability, keep the nicely named functions, reduce support for alternative named functions to 2 kinds (pick which 2 based on usage?), double parameter count and tack `Short` on the end for the smaller shaped functions. So you could have `VisitTypeName` and `VisitTypeNameShort` (or simple or whatever suffix you like). Unfortunately I think you'd need to `panic` if you set both so there'd still be a possibility of `panic`. This is sortof a middle ground in my mind.

If other folks on the team prefer this approach to either of the two above then I'm fine to see this merge (there are aspects I like as I mentioned), but I personally prefer either of the above to the current reflection based approach (I _think_ I lean towards option 1).

Happy to discuss further in person as well.

hack/generator/pkg/astmodel/type_definition.go
theunrepentantgeek: Resolved with restructure.

hack/generator/pkg/astmodel/type_visitor_test.go
```diff
@@ -165,3 +175,150 @@ func TestIdentityVisitorReturnsEqualResult(t *testing.T) {
...
+				builder.VisitFlaggedType = func(ft *FlaggedType) (Type, error) {
+					return ft, errors.New(ft.String())
+				}
+			},
+			StorageFlag.ApplyTo(StringType),
+			"storage",
+		},
+	}
+
+	// TODO: Pending tests for
```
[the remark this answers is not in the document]
theunrepentantgeek: Was intending on returning to them later - didn't want to take the time if this PR wasn't going to land.

hack/generator/pkg/astmodel/type_walker_test.go
```diff
@@ -273,8 +273,8 @@ func TestTypeWalker_VisitorApplied(t *testing.T) {
 	g := NewGomegaWithT(t)
 
 	types := makeSimpleTestTypeGraph()
-	visitor := MakeTypeVisitor()
-	visitor.VisitObjectType = func(this *TypeVisitor, it *ObjectType, ctx interface{}) (Type, error) {
+	visitor := TypeVisitorBuilder{}.Build()
+	visitor.visitObjectType = func(this *TypeVisitor, it *ObjectType, ctx interface{}) (Type, error) {
```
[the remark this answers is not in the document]
theunrepentantgeek: Done.

hack/generator/pkg/codegen/pipeline_augment_status.go
```diff
@@ -166,16 +168,6 @@ func generateStatusTypes(swaggerTypes swaggerTypes) (statusTypes, error) {
 	return statusTypes{resourceLookup, otherTypes}, nil
 }
 
-func makeRenamingVisitor(rename func(astmodel.TypeName) astmodel.TypeName) astmodel.TypeVisitor {
```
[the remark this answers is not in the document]
theunrepentantgeek: Fair enough - restored.

hack/generator/pkg/codegen/pipeline_remove_type_aliases.go
```diff
@@ -22,13 +22,17 @@ func removeTypeAliases() PipelineStage {
 		"removeAliases",
 		"Remove type aliases",
 		func(ctx context.Context, types astmodel.Types) (astmodel.Types, error) {
-			visitor := astmodel.MakeTypeVisitor()
-			var result = make(astmodel.Types)
 
-			visitor.VisitTypeName = func(this *astmodel.TypeVisitor, it astmodel.TypeName, ctx interface{}) (astmodel.Type, error) {
+			removeAlias := func(this *astmodel.TypeVisitor, it astmodel.TypeName, ctx interface{}) (astmodel.Type, error) {
```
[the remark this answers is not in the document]
theunrepentantgeek: Renamed to simplifyAliases which is more revealing that there are mulitiple possible outcomes.

PR #441, opened 2021-04-22
Rename Arm -> ARM

matthchr:


matthchr:
This is purely just renames, no actual logic changes

Porges:
@matthchr do you mind if I delay this until I merge my changes? I've renamed a few ARM-related items. (Worst case I can just recreate this PR afterwards.)

matthchr:
How bad do you think the conflicts will be? I also have some stuff based on top of this... but possibly can try rebasing it, which shouldn't be too bad.

Porges:
> How bad do you think the conflicts will be? I also have some stuff based on top of this... but possibly can try rebasing it, which shouldn't be too bad.

Okay, if you have stuff on it then go ahead :)

PR #442, opened 2021-04-22
Document feature ask for Azure REST API team

matthchr:


hack/generator/spec/swagger-resource-references-onepager.md
```diff
@@ -0,0 +1,54 @@
+# The feature ask
+My team would like the ability to look at a particular field in the Swagger and know that field is a resource reference (a reference to an ARM ID). Our understanding is that today that is not possible. Some examples of references in existing specs:
```
theunrepentantgeek, line 2:
```suggestion
My team would like the ability to look at a particular field in the Swagger and know whether that field is a resource reference (a reference to an ARM ID). Our understanding is that today that is not possible. Some examples of references in existing specs:
```

hack/generator/spec/swagger-resource-references-onepager.md
```diff
@@ -0,0 +1,54 @@
...
+1. [Compute](https://github.com/Azure/azure-rest-api-specs/blob/master/specification/compute/resource-manager/Microsoft.Compute/stable/2021-03-01/cloudService.json#L1745)
+2. [Networking](https://github.com/Azure/azure-rest-api-specs/blob/master/specification/network/resource-manager/Microsoft.Network/stable/2020-07-01/network.json#L171)
+3. [Batch](https://github.com/Azure/azure-rest-api-specs/blob/master/specification/batch/resource-manager/Microsoft.Batch/stable/2021-01-01/BatchManagement.json#L2324)
+
+Not every team does references the same way (which is fine), but that means that as a client of the Swagger it's impossible to tell if a given field is a reference or not.
```
theunrepentantgeek, line 7:
```suggestion
Not every team handles ARM references the same way (which is fine), but that means that as a client of the Swagger it is currently impossible to tell if a given field is a reference or not.
```

hack/generator/spec/swagger-resource-references-onepager.md
```diff
@@ -0,0 +1,54 @@
...
+      },
+      "description": "Reference to another subresource."
+    },
+```
+
+## What is Azure Service Operator (ASO)?
+It's a Kubernetes operator which supports creating Azure resources. You can read more [here](https://github.com/Azure/azure-service-operator). It allows customers to create Azure resources directly in their Kubernetes cluster by defining YAML files representing the Azure resource and then doing `kubectl apply -f <my-yaml>`.
+
+## How do resource references work in ASO?
+ASO allows you to represent various Azure resources in Kubernetes. As a consequence of the fact that the resource is reflected in Kubernetes, you also can _refer_ to resources in one of two ways:
```
theunrepentantgeek, line 28:
```suggestion
ASO allows you to represent various Azure resources in Kubernetes. As a consequence of the fact that the resource is reflected in Kubernetes, you may also _refer_ to resources in two different ways:
```

hack/generator/spec/swagger-resource-references-onepager.md
```diff
@@ -0,0 +1,54 @@
...
+
+You can read more about the references design [here](https://github.com/Azure/k8s-infra/blob/master/hack/generator/spec/type_references_and_ownership.md)
+
+## Why do we want this ability?
+In a lot of ways, the maintainers of ASO (that's my team and I) are in the same position as the SDK team: there are tons of Azure services and there's no way we can keep up with the pace of new API versions/services without some sort of automation helping us. We are currently most of the way through developing a code-generator that generates Kubernetes CRDs from the azure-rest-api-specs Swager documents - you can see this generator here: [k8sinfra](https://github.com/Azure/k8s-infra).
+
+In order for the generator to generate the right shape for resources that have references, we need a way to determine that a particular field is a reference, so that we can replace the simple `string` field with something like:
```
theunrepentantgeek, line 37:
```suggestion
In order for the generator to emit the right shape for resources that have references, we need a reliable/authoritative way to determine that a particular field is a reference, so that we can replace the simple `string` field with something like:
```

theunrepentantgeek (commented):
Looks good - made a few wording suggestions, nothing too important.

hack/generator/spec/swagger-resource-references-onepager.md
matthchr: Fixed

hack/generator/spec/swagger-resource-references-onepager.md
matthchr: done

hack/generator/spec/swagger-resource-references-onepager.md
matthchr: Done

hack/generator/spec/swagger-resource-references-onepager.md
matthchr: Done

babbageclunk (approved):

hack/generator/spec/swagger-resource-references-onepager.md
```diff
@@ -0,0 +1,54 @@
...
+2. [Networking](https://github.com/Azure/azure-rest-api-specs/blob/master/specification/network/resource-manager/Microsoft.Network/stable/2020-07-01/network.json#L171)
+3. [Batch](https://github.com/Azure/azure-rest-api-specs/blob/master/specification/batch/resource-manager/Microsoft.Batch/stable/2021-01-01/BatchManagement.json#L2324)
+
+Not every team handles ARM references the same way (which is fine), but that means that as a client of the Swagger it is currently impossible to tell if a given field is a reference or not.
+
+## Digging in
+The easiest way to accomplish this is just to expose "is this field a reference" as an extension in Swagger using `format`. Probably something like: 
+```
+    "SubResource": {
+      "properties": {
```
johanste, line 13: I would suggest adding common semantic types to the common definitions file. This way we can be even more specific and indicate what kind of resource it is, and not just an arbitrary resource (e.g. indicate that something is a NIC type resource rather than an arbitrary resource type).

matthchr: I agree this would be nice -- although I wonder how it plays with how services version. There are multiple references I'm aware of where after-the-fact another type was supported. For example when Shared Image Gallery images came out, I believe you could put those references in the same place the old custom image reference was, with no change to spec or APIVersion. Would the expectation be that the service team would update the spec with that additional information? Just future versions of the spec?

johanste: That is a valid concern. Versioning is hard. But I would also be concerned about adding this kind of support to existing APIs breaking clients regardless of how we model it here - it is not unreasonable that a client would "know" how to follow the reference and get horribly broken if a new, unexpected resource was referenced.

johanste: A client *shouldn't* do that. But there are plenty of scenarios where you pretty much have to follow links. Not sure if there is a more convenient API these days to list public IP addresses for a VM, but when I last implemented it, I had to follow links. And I'm pretty sure if any of those links started pointing to something vastly different than I expected, my code would have blown up. 

So there may not strictly be a formal contract that something will always point to a resource of a given type, but customers will invariably take dependencies on undocumented behavior - and in many cases, that is actually the only way for them to get their workloads to work. 

I would expect a change like what you describe above to come with a tail end of ICMs. And it would be difficult to completely blame the caller for having made a mistake/taken a dependency on undocumented behavior.

hack/generator/spec/swagger-resource-references-onepager.md
```diff
@@ -0,0 +1,54 @@
...
+	// Group is the Kubernetes group of the resource.
+	Group string `json:"group"`
+	// Kind is the Kubernetes kind of the resource.
+	Kind string `json:"kind"`
+	// Namespace is the Kubernetes namespace of the resource.
+	Namespace string `json:"namespace"`
+	// Name is the Kubernetes name of the resource.
+	Name string `json:"name"`
+
+        // +kubebuilder:validation:Pattern="^/subscriptions/(.+?)/resourcegroups/(.+?)/providers/(.+?)/((.+?)/(.+?))+$"
```
m-nash, line 49: Not sure if this matters for the proposal but just FYI.  This isn't the format for all resourceIds.  It could be an id of a resource that lives under a subscription so the resourceGroups won't be there, it could be something that lives under a tenant, so subscription and resourcegroups won't be there.  it could be an extension resource where you have an entire resourceId in the middle of one of these resourceIds.  I believe this also assumes that the end consists of /ref/value/ pairs, but thats not always the case either you can have /ref1/ref2/value/ in fact the number of refs doesn't seem to have a limit.

PR #443, opened 2021-04-23
Add ReflectVisitor and ResolvedReferences

matthchr:
Laying the groundwork for resolving arbitrary cross resource references.

### DIFF
--- a/hack/generated/pkg/genruntime/resolved_references.go
+++ b/hack/generated/pkg/genruntime/resolved_references.go
@@ -8,21 +8,14 @@ package genruntime
 // ResolvedReferences is a set of references which have been resolved for a particular resource.
 // The special self field is the fully qualified ARM ID of the resource that this ResolvedReferences applies to.
 type ResolvedReferences struct {
-	// self is the fully qualified name of the resource in Azure ("a/b/c").
-	self string
 	// references is a map of ResourceReference to ARM ID.
 	references map[ResourceReference]string
 }
 
 // MakeResolvedReferences creates a ResolvedReferences from the fully qualified ARM ID of the resource and
 // and ARM IDs that the resource refers to.
-func MakeResolvedReferences(name string, references map[ResourceReference]string) ResolvedReferences {
-	if name == "" {
-		panic("ResolvedReferences name is required")
-	}
-
+func MakeResolvedReferences(references map[ResourceReference]string) ResolvedReferences {
 	return ResolvedReferences{
-		self:       name,
 		references: references,
 	}
 }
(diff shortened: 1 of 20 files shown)


hack/generated/pkg/reflecthelpers/reflect_visitor.go
```diff
@@ -0,0 +1,158 @@
...
+
+type ReflectVisitor struct {
+	VisitPrimitive func(this *ReflectVisitor, it interface{}, ctx interface{}) error
+	VisitStruct    func(this *ReflectVisitor, it interface{}, ctx interface{}) error
+	VisitPtr       func(this *ReflectVisitor, it interface{}, ctx interface{}) error
+	VisitSlice     func(this *ReflectVisitor, it interface{}, ctx interface{}) error
+	VisitMap       func(this *ReflectVisitor, it interface{}, ctx interface{}) error
+}
+
+func NewReflectVisitor() *ReflectVisitor {
```
matthchr, line 52: Missing comments for many of these functions

hack/generated/pkg/genruntime/resolved_references.go
```diff
@@ -0,0 +1,29 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ */
+
+package genruntime
+
+type ResolvedReferences struct {
```
matthchr, line 8: docs

hack/generated/pkg/genruntime/resolved_references.go
```diff
@@ -0,0 +1,39 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ */
+
+package genruntime
+
+// ResolvedReferences is a set of references which have been resolved for a particular resource.
+// The special self field is the fully qualified ARM ID of the resource that this ResolvedReferences applies to.
+type ResolvedReferences struct {
```
babbageclunk, line 10: I don't get why you're bundling these two things together into one struct. Why is name in here?

babbageclunk (approved):
I like this, but I think name should be a separate argument to ConvertToARM, rather than packaging it up with the resolved references.

The reflect-based visitor is nice! I've seen code in other projects that uses reflection to gather info from arbitrary types in similar but in a much more ad-hoc way - this is much better.

hack/generated/pkg/genruntime/resolved_references.go
matthchr: Pulled it back out

matthchr: Removed name from this

PR #444, opened 2021-04-27
Split ArmConversionFunction

Porges:
Split ArmConversionFunction into two types rather than one type with a boolean.


babbageclunk (approved):
Much neater!

PR #445, opened 2021-04-27
Add TypeMerger

Porges:
This is called `TypeMerger` but it is really multiple dispatch for functions with type `(a <: Type, b <: Type) → Type`. Ideally this would be a bit more generalized to _n_ arguments but Go doesn’t support that kind of thing.

----

Open question: _Should_ this be generalized to allow functions of any arity (in which case I’d probably rename to `TypeFuncDispatcher` or something similarly obtuse)? 

Self answer: probably not. If we need `(a <: Type) → Type` we have type switches already, and I can’t see any reason for a 3-argument version.

### DIFF
--- a/hack/generator/pkg/astmodel/type_merger.go
+++ b/hack/generator/pkg/astmodel/type_merger.go
@@ -15,9 +15,9 @@ import (
 //
 // Conceptually it takes (via Add) a list of functions of the form:
 //
-//     func ([ctx interface{},] a <: Type, b <: Type) (Type, error)
+//     func ([ctx interface{},] left {some Type}, right {some Type}) (Type, error)
 //
-// where `a` and `b` can be concrete types that implement the `Type` interface.
+// where `left` and `right` can be concrete types that implement the `Type` interface.
 //
 // When `TypeMerger.Merge(Type, Type)` is invoked, it will iterate through the
 // provided functions and invoke the first one that matches the concrete types
@@ -58,7 +58,7 @@ func validateMerger(merger interface{}) validatedMerger {
 
 	mergerType := it.Type()
 
-	badArgumentsMsg := "merger must take take arguments of type (left Type, right Type) or (ctx X, left Type, right Type)"
+	badArgumentsMsg := "merger must take take arguments of type (left [some Type], right [some Type]) or (ctx X, left [some Type], right [some Type])"
 	if mergerType.NumIn() < 2 || mergerType.NumIn() > 3 {
 		panic(badArgumentsMsg)
 	}
--- a/hack/generator/pkg/astmodel/type_merger.go
+++ b/hack/generator/pkg/astmodel/type_merger.go
@@ -7,29 +7,33 @@ package astmodel
 
 import (
 	"reflect"
+
+	"github.com/pkg/errors"
 )
 
 // TypeMerger is like a visitor for 2 types.
 // The `ctx` argument can be used to “smuggle” additional data down the call-chain.
 type TypeMerger struct {
-	mergers  map[TypePair]TypeMergerFunc
-	fallback TypeMergerFunc
+	mergers  []mergerRegistration
+	fallback MergerFunc
 }
 
-type TypePair struct{ left, right reflect.Type }
-type TypeMergerFunc func(left, right Type) Type
+type mergerRegistration struct {
+	left, right reflect.Type
+	merge       MergerFunc
+}
 
-func NewTypeMerger(fallback TypeMergerFunc) TypeMerger {
-	return TypeMerger{
-		mergers:  make(map[TypePair]TypeMergerFunc),
-		fallback: fallback,
-	}
+type MergerFunc func(ctx interface{}, left, right Type) (Type, error)
+
+func NewTypeMerger(fallback MergerFunc) TypeMerger {
+	return TypeMerger{fallback: fallback}
 }
 
 var typeInterface reflect.Type = reflect.TypeOf((*Type)(nil)).Elem() // yuck
+var errorInterface reflect.Type = reflect.TypeOf((*error)(nil)).Elem()
+var mergerFuncType reflect.Type = reflect.TypeOf((*MergerFunc)(nil)).Elem()
 
-func (m *TypeMerger) Add(merger interface{}) {
-
+func validateMerger(merger interface{}) (reflect.Value, reflect.Type, reflect.Type, bool) {
 	it := reflect.ValueOf(merger)
 	if it.Kind() != reflect.Func {
 		panic("merger must be a function")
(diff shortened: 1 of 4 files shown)


hack/generator/pkg/astmodel/type_merger.go
```diff
@@ -0,0 +1,178 @@
...
+
+func NewTypeMerger(fallback MergerFunc) TypeMerger {
+	return TypeMerger{fallback: fallback}
+}
+
+var typeInterface reflect.Type = reflect.TypeOf((*Type)(nil)).Elem() // yuck
+var errorInterface reflect.Type = reflect.TypeOf((*error)(nil)).Elem()
+var mergerFuncType reflect.Type = reflect.TypeOf((*MergerFunc)(nil)).Elem()
+
+func validateMerger(merger interface{}) (reflect.Value, reflect.Type, reflect.Type, bool) {
```
matthchr, line 47: Would a little struct rather than a 4-tuple return be clearer?

matthchr: If not, _might_ be worth defining `needsCtx` as `needsCtx := inOffset != 0` so it's at least more clear from the direct return results what that field is?

hack/generator/pkg/astmodel/type_merger.go
```diff
@@ -0,0 +1,178 @@
...
+	if mergerType.NumOut() != 2 ||
+		mergerType.Out(0) != typeInterface ||
+		mergerType.Out(1) != errorInterface {
+		panic("merger must return (Type, error)")
+	}
+
+	return it, leftArg, rightArg, inOffset != 0
+}
+
+func (m *TypeMerger) Add(mergeFunc interface{}) {
```
matthchr, line 76: Doc discussing expected shape of `mergeFunc`? You basically say it above in the docs on `TypeMerger` though...

hack/generator/pkg/astmodel/type_merger.go
```diff
@@ -0,0 +1,178 @@
...
+				if takesCtx {
+					return merger.Call([]reflect.Value{ctxValue, leftValue, rightValue})
+				} else {
+					return merger.Call([]reflect.Value{leftValue, rightValue})
+				}
+			}).Interface().(MergerFunc),
+		})
+}
+
+func (m *TypeMerger) AddUnordered(mergeFunc interface{}) {
```
matthchr, line 99: How is this different from add? Doc would help

hack/generator/pkg/astmodel/type_merger.go
```diff
@@ -0,0 +1,178 @@
...
+				if takesCtx {
+					return merger.Call([]reflect.Value{ctxValue, leftValue, rightValue})
+				} else {
+					return merger.Call([]reflect.Value{leftValue, rightValue})
+				}
+			}).Interface().(MergerFunc),
+		})
+}
+
+// Merge merges the two types according to the provided mergers and fallback
```
matthchr, line 135:
```suggestion
// Merge merges the two types according to the provided mergers and fallback. The ctx parameter is optional and if supplied can only be supplied once
```

maybe?

hack/generator/pkg/astmodel/type_merger.go
```diff
@@ -0,0 +1,178 @@
...
+func (m *TypeMerger) Merge(left, right Type, ctx ...interface{}) (Type, error) {
```
matthchr, line 136: minor: Depending on usage you could have 2 "overloads" `Merge` and `MergeCtx` one of which takes a `ctx` and the other of which doesn't?

Might be cleaner than this pattern - unsure.

hack/generator/pkg/astmodel/type_merger.go
```diff
@@ -0,0 +1,178 @@
...
+
+	if right == nil {
+		return left, nil
+	}
+
+	leftType := reflect.ValueOf(left).Type()
+	rightType := reflect.ValueOf(right).Type()
+
+	for _, merger := range m.mergers {
+		if (merger.left == leftType || merger.left == typeInterface) &&
```
matthchr, line 160: minor: Consider lifting out into named vars just for a bit more clarity?
```
                leftMatches := merger.left == leftType || merger.left == typeInterface
                rightMatches := merger.right == rightType || merger.right == typeInterface
		if leftMatches && rightMatches {
```

matthchr (approved):
Left a few minor comments but LGTM overall

hack/generator/pkg/astmodel/type_merger.go
Porges: I considered this and I do think it would be cleaner overall

matthchr (approved):

babbageclunk (approved):
Looks great!

hack/generator/pkg/astmodel/type_merger.go
```diff
@@ -0,0 +1,172 @@
...
+
+	if mergerType.NumIn() != 2 && mergerType.NumIn() != 3 {
+		panic("merger must take 2 arguments (left Type, right Type) or 3 arguments (ctx X, left Type, right Type)")
+	}
+
+	argOffset := mergerType.NumIn() - 2
+
+	needsCtx := argOffset != 0
+	leftArg := mergerType.In(argOffset + 0)
+	rightArg := mergerType.In(argOffset + 1)
```
theunrepentantgeek, line 69: I suspect these would be simpler if context was an optional third parameter, with left and right always in positions `0` and `1`

hack/generator/pkg/astmodel/type_merger.go
```diff
@@ -0,0 +1,172 @@
...
+
+	if !leftArg.AssignableTo(typeInterface) || !rightArg.AssignableTo(typeInterface) {
+		panic("merger must take in types assignable to Type")
+	}
```
theunrepentantgeek, line 73: Perhaps split this test so you can be explicit about whether it's _left_ or _right_ that's wrong?

hack/generator/pkg/astmodel/type_merger.go
Porges: Yes, I originally had it that way around, but it greatly simplifies the use to have it first so you can pass it member functions (see use in the allof/oneof pipeline).

hack/generator/pkg/astmodel/type_merger.go
```diff
@@ -0,0 +1,170 @@
...
+	"reflect"
+
+	"github.com/pkg/errors"
+)
+
+// TypeMerger is like a visitor for 2 types.
+//
+// Conceptually it takes (via Add) a list of functions of the form:
+//
+//     func ([ctx interface{},] a <: Type, b <: Type) (Type, error)
```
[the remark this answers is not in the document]
Porges: 1st question: some concrete type that implements `Type`, or the type `Type` itself.
2nd question: see: https://github.com/Azure/k8s-infra/pull/445#discussion_r623545281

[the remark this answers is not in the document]
Porges: I guess I could write something more C#y like `func<A: Type, B: Type> ([ctx interface{},] A, B) (Type, error)`

hack/generator/pkg/astmodel/type_merger.go
Porges: Thought about it a bit, I'm actually going to go in the other direction and unify this message with the previous one, so if you mess up the argument type at all, it tells you what you need to pass.

hack/generator/pkg/astmodel/type_merger.go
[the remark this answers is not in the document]
Porges: updated with different notation, which should be clearer

hack/generator/pkg/astmodel/type_merger.go
```diff
@@ -0,0 +1,170 @@
...
+// passed. If none match then the fallback provided to `NewTypeMerger` will be invoked.
+//
+// The `ctx` argument can optionally be used to “smuggle” additional data down the call-chain.
+type TypeMerger struct {
+	mergers  []mergerRegistration
+	fallback MergerFunc
+}
+
+type mergerRegistration struct {
+	left, right reflect.Type
```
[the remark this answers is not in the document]
Porges: done!

PR #448, opened 2021-05-04
Add ConversionFunctionBuilder

matthchr:
- Refactor ConvertTo/From ARM handlers to use it.
    This reduces duplicate code and fixes a small bug in
    the ConvertToARM handler.

### DIFF
--- a/hack/generator/pkg/astmodel/armconversion/convert_from_arm_function_builder.go
+++ b/hack/generator/pkg/astmodel/armconversion/convert_from_arm_function_builder.go
@@ -17,8 +17,9 @@ import (
 
 type convertFromARMBuilder struct {
 	conversionBuilder
-	typedInputIdent string
-	inputIdent      string
+	typedInputIdent       string
+	inputIdent            string
+	typeConversionBuilder *astmodel.ConversionFunctionBuilder
 }
 
 func newConvertFromARMFunctionBuilder(
@@ -45,8 +46,16 @@ func newConvertFromARMFunctionBuilder(
 			isSpecType:            c.isSpecType,
 			codeGenerationContext: codeGenerationContext,
 		},
+		typeConversionBuilder: astmodel.NewConversionFunctionBuilder(c.idFactory, codeGenerationContext),
 	}
 
+	// It's a bit awkward that there are two levels of "handler" here, but they serve different purposes:
+	// The top level propertyConversionHandlers is about determining which properties are involved: given a property on the destination type it
+	// determines which property (if any) on the source type will be converted to the destination.
+	// The "inner" handler (typeConversionBuilder) is about determining how to convert between two types: given a
+	// source type and a destination type, figure out how to make the assignment work. It has no knowledge of broader object strucutre
+	// or other properties.
+	result.typeConversionBuilder.AddConversionHandlers(result.convertComplexTypeNameProperty)
 	result.propertyConversionHandlers = []propertyConversionHandler{
 		result.namePropertyHandler,
 		result.ownerPropertyHandler,
@@ -71,10 +80,7 @@ func (builder *convertFromARMBuilder) functionDeclaration() *dst.FuncDecl {
 	fn.AddComments("populates a Kubernetes CRD object from an Azure ARM object")
 	fn.AddParameter(
 		builder.idFactory.CreateIdentifier(astmodel.OwnerProperty, astmodel.NotExported),
-		&dst.SelectorExpr{
-			X:   dst.NewIdent(astmodel.GenRuntimePackageName),
-			Sel: dst.NewIdent("KnownResourceReference"),
-		})
+		astbuilder.Selector(dst.NewIdent(astmodel.GenRuntimePackageName), "KnownResourceReference"))
 
 	fn.AddParameter(builder.inputIdent, dst.NewIdent("interface{}"))
 	fn.AddReturns("error")
@@ -161,10 +167,7 @@ func (builder *convertFromARMBuilder) namePropertyHandler(
 				astbuilder.CallQualifiedFunc(
 					astmodel.GenRuntimePackageName,
 					"ExtractKubernetesResourceNameFromARMName",
-					&dst.SelectorExpr{
-						X:   dst.NewIdent(builder.typedInputIdent),
-						Sel: dst.NewIdent(string(fromProp.PropertyName())),
-					}),
+					astbuilder.Selector(dst.NewIdent(builder.typedInputIdent), string(fromProp.PropertyName()))),
 			),
 		},
 	}
@@ -179,10 +182,7 @@ func (builder *convertFromARMBuilder) ownerPropertyHandler(
 	}
 
 	result := astbuilder.SimpleAssignment(
-		&dst.SelectorExpr{
-			X:   dst.NewIdent(builder.receiverIdent),
-			Sel: dst.NewIdent(string(toProp.PropertyName())),
-		},
+		astbuilder.Selector(dst.NewIdent(builder.receiverIdent), string(toProp.PropertyName())),
 		token.ASSIGN,
 		dst.NewIdent(builder.idFactory.CreateIdentifier(astmodel.OwnerProperty, astmodel.NotExported)))
 	return []dst.Stmt{result}
@@ -208,42 +208,17 @@ func (builder *convertFromARMBuilder) propertiesWithSameNameAndTypeHandler(
 		return nil
 	}
 
-	if typeRequiresCopying(toType) {
-		// We can't get away with just assigning this field, since
-		// it's a reference type. Use the conversion code to copy the
-		// elements.
-
-		return builder.fromARMComplexPropertyConversion(
-			complexPropertyConversionParameters{
-				source: &dst.SelectorExpr{
-					X:   dst.NewIdent(builder.typedInputIdent),
-					Sel: dst.NewIdent(string(toProp.PropertyName())),
-				},
-				destination: &dst.SelectorExpr{
-					X:   dst.NewIdent(builder.receiverIdent),
-					Sel: dst.NewIdent(string(toProp.PropertyName())),
-				},
-				destinationType:   toType,
-				nameHint:          string(toProp.PropertyName()),
-				conversionContext: nil,
-				assignmentHandler: nil,
-				sameTypes:         true,
-			},
-		)
-	}
-
-	result := astbuilder.SimpleAssignment(
-		&dst.SelectorExpr{
-			X:   dst.NewIdent(builder.receiverIdent),
-			Sel: dst.NewIdent(string(fromProp.PropertyName())),
+	return builder.typeConversionBuilder.BuildConversion(
+		astmodel.ConversionParameters{
+			Source:            astbuilder.Selector(dst.NewIdent(builder.typedInputIdent), string(fromProp.PropertyName())),
+			SourceType:        fromProp.PropertyType(),
+			Destination:       astbuilder.Selector(dst.NewIdent(builder.receiverIdent), string(toProp.PropertyName())),
+			DestinationType:   toProp.PropertyType(),
+			NameHint:          string(toProp.PropertyName()),
+			ConversionContext: nil,
+			AssignmentHandler: nil,
 		},
-		token.ASSIGN,
-		&dst.SelectorExpr{
-			X:   dst.NewIdent(builder.typedInputIdent),
-			Sel: dst.NewIdent(string(toProp.PropertyName())),
-		})
-
-	return []dst.Stmt{result}
+	)
 }
 
 func (builder *convertFromARMBuilder) propertiesWithSameNameButDifferentTypeHandler() propertyConversionHandler {
@@ -265,20 +240,15 @@ func (builder *convertFromARMBuilder) propertiesWithSameNameButDifferentTypeHand
 			definedErrVar = true
 		}
 
-		complexConversion := builder.fromARMComplexPropertyConversion(
-			complexPropertyConversionParameters{
-				source: &dst.SelectorExpr{
-					X:   dst.NewIdent(builder.typedInputIdent),
-					Sel: dst.NewIdent(string(fromProp.PropertyName())),
-				},
-				destination: &dst.SelectorExpr{
-					X:   dst.NewIdent(builder.receiverIdent),
-					Sel: dst.NewIdent(string(toProp.PropertyName())),
-				},
-				destinationType:   toProp.PropertyType(),
-				nameHint:          string(toProp.PropertyName()),
-				conversionContext: nil,
-				assignmentHandler: nil,
+		complexConversion := builder.typeConversionBuilder.BuildConversion(
+			astmodel.ConversionParameters{
+				Source:            astbuilder.Selector(dst.NewIdent(builder.typedInputIdent), string(fromProp.PropertyName())),
+				SourceType:        fromProp.PropertyType(),
+				Destination:       astbuilder.Selector(dst.NewIdent(builder.receiverIdent), string(toProp.PropertyName())),
+				DestinationType:   toProp.PropertyType(),
+				NameHint:          string(toProp.PropertyName()),
+				ConversionContext: nil,
+				AssignmentHandler: nil,
 			})
 
 		result = append(result, complexConversion...)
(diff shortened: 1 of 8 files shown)


hack/generator/pkg/astmodel/conversion_function_builder.go
```diff
@@ -0,0 +1,466 @@
...
+import (
+	"fmt"
+	"go/token"
+
+	"github.com/dave/dst"
+
+	"github.com/Azure/k8s-infra/hack/generator/pkg/astbuilder"
+)
+
+type ConversionParameters struct {
```
matthchr, line 17: Probably should have comments on these functions/types

hack/generator/pkg/astmodel/conversion_function_builder.go
```diff
@@ -0,0 +1,466 @@
...
+// It has a set of built-in conversions and can be configured with additional conversions.
+type ConversionFunctionBuilder struct {
+	// TODO: Better way to let you fuss with this? How can you pick out what I've already put in here to overwrite it?
+	conversions []ConversionHandler
+
+	IdFactory             IdentifierFactory
+	CodeGenerationContext *CodeGenerationContext
+}
+
+func NewConversionFunctionBuilder(idFactory IdentifierFactory, codeGenerationContext *CodeGenerationContext) *ConversionFunctionBuilder {
```
matthchr, line 125: It'd be nice to both: "have an order" to things that are processed, but also allow them to be overwritable/movable (like `TypeVisitor`'s are).

I didn't spend a ton of time thinking how to accomplish that for this first pass but possibly something to think about if this is going to be used more broadly.

hack/generator/pkg/astmodel/armconversion/convert_from_arm_function_builder.go
```diff
@@ -45,8 +46,11 @@ func newConvertFromARMFunctionBuilder(
 			isSpecType:            c.isSpecType,
 			codeGenerationContext: codeGenerationContext,
 		},
+		typeConversionBuilder: astmodel.NewConversionFunctionBuilder(c.idFactory, codeGenerationContext),
 	}
 
+	result.typeConversionBuilder.AddConversionHandlers(result.convertComplexTypeNameProperty)
+
 	result.propertyConversionHandlers = []propertyConversionHandler{
```
matthchr, line 54: I initially wanted to do away with this double-nesting, (we have handlers and then one of the handlers uses the `typeConversionBuilder` which itself has handlers), but actually it makes sense...

The upper handler (`propertyConversionHandlers`) is about determining which properties are involved: given a property on the destination type it determines which property (if any) on the source type will be converted to the destination.
The "inner" handler `typeConversionBuilder` is about determining how to convert between two types: given a source type and a destination type, figure out how to make an assignment. It has no knowledge of broader object structure of "other" properties.

hack/generator/pkg/astmodel/armconversion/convert_to_arm_function_builder.go
```diff
@@ -173,308 +175,35 @@ func (builder *convertToARMBuilder) fixedValuePropertyHandler(propertyName astmo
...
-//////////////////////////////////////////////////////////////////////////////////
-// Complex property conversion (for when properties aren't simple primitive types)
-//////////////////////////////////////////////////////////////////////////////////
-
-func (builder *convertToARMBuilder) toARMComplexPropertyConversion(
-	params complexPropertyConversionParameters) []dst.Stmt {
-
-	switch concrete := params.destinationType.(type) {
-	case *astmodel.OptionalType:
-		return builder.convertComplexOptionalProperty(params)
```
theunrepentantgeek, line 266: This seems to have a built in assumption that the source and destination properties are the same type, which possibly makes this a simpler application than the conversions I've written for storage types.

hack/generator/pkg/astmodel/conversion_function_builder.go
```diff
@@ -0,0 +1,494 @@
...
+	Destination       dst.Expr
+	SourceType        Type
+	DestinationType   Type
+	NameHint          string
+	ConversionContext []Type
+	AssignmentHandler func(destination, source dst.Expr) dst.Stmt
+}
+
+// ClonedSource clones the Source field.
+func (params ConversionParameters) ClonedSource() dst.Expr {
```
theunrepentantgeek, line 29: Do you need to call out that it is cloned? It's kind of an implementation detail.

What about making the field `source` and providing a method `Source()` that just happens to return a clone.

hack/generator/pkg/astmodel/conversion_function_builder.go
```diff
@@ -0,0 +1,494 @@
...
+	return dst.Clone(params.Source).(dst.Expr)
+}
+
+// ClonedDestination clones the Destination field.
+func (params ConversionParameters) ClonedDestination() dst.Expr {
```
theunrepentantgeek, line 34: Same idea here
```suggestion
func (params ConversionParameters) Destination() dst.Expr {
```

hack/generator/pkg/astmodel/conversion_function_builder.go
```diff
@@ -0,0 +1,494 @@
...
+		return AssignmentHandlerAssign
+	}
+	return params.AssignmentHandler
+}
+
+// TODO: Consider a better way to do this, using KnownLocalsSet?
+// CountArraysAndMapsInConversionContext returns the number of arrays/maps which are in the conversion context.
+// This is to aid in situations where there are deeply nested conversions (i.e. array of map of maps). In these contexts,
+// just using a simple assignment such as "elem := ..." isn't sufficient because elem my already have been defined above by
+// an enclosing map/array conversion context. We use the depth to do "elem1 := ..." or "elem7 := ...".
```
theunrepentantgeek, line 91: For the storage conversions I reduced the number of collisions by using distinct suffixes - `Key` and `Value` for maps, `Item` and `Index` (if I recall correctly) for lists, and so on.

hack/generator/pkg/astmodel/conversion_function_builder.go
```diff
@@ -0,0 +1,494 @@
...
+
+func (builder *ConversionFunctionBuilder) BuildConversion(params ConversionParameters) []dst.Stmt {
+	for _, conversion := range builder.conversions {
+		result := conversion(builder, params)
+		if len(result) > 0 {
+			return result
+		}
+	}
+
+	panic(fmt.Sprintf("don't know how to perform conversion for %T -> %T", params.SourceType, params.DestinationType))
```
theunrepentantgeek, line 173: Suggest using the debug descriptions of the types to provide more information about each

hack/generator/pkg/astmodel/conversion_function_builder.go
```diff
@@ -0,0 +1,494 @@
...
+			WithSource(newSource).
+			WithSourceType(sourceType.Element()))
+
+	// Tack on the final assignment
+	innerStatements = append(
+		innerStatements,
+		astbuilder.SimpleAssignment(
+			params.ClonedDestination(),
+			token.ASSIGN,
+			&dst.UnaryExpr{
```
theunrepentantgeek, line 215: Use `astbuilder.AddrOf()`

hack/generator/pkg/astmodel/conversion_function_builder.go
```diff
@@ -0,0 +1,494 @@
...
+
+	if _, ok := destinationType.KeyType().(*PrimitiveType); !ok {
+		panic(fmt.Sprintf("map had non-primitive key type: %v", destinationType.KeyType()))
+	}
+
+	depth := params.CountArraysAndMapsInConversionContext()
+
+	keyIdent := "key"
+	valueIdent := "value"
+	elemIdent := "elem"
```
theunrepentantgeek, line 330: Maybe `item`?

hack/generator/pkg/astmodel/conversion_function_builder.go
```diff
@@ -0,0 +1,494 @@
...
+	}
+
+	return []dst.Stmt{result}
+}
+
+// IdentityAssignTypeName handles conversion for TypeName's that are the same
+// Note that because this handler is dealing with TypeName's and not Optional<TypeName>, it is safe to
+// perform a simple assignment rather than a copy.
+// This function generates code that looks like this:
+//// <destination> <assignmentHandler> <source>
```
theunrepentantgeek, line 398:
```suggestion
//
// <destination> <assignmentHandler> <source>
//
```

hack/generator/pkg/astmodel/conversion_function_builder.go
```diff
@@ -0,0 +1,494 @@
...
+// It generates code that looks like:
+//     <destination> = *<source>.DeepCopy()
+func IdentityDeepCopyJSON(builder *ConversionFunctionBuilder, params ConversionParameters) []dst.Stmt {
+
+	if !params.DestinationType.Equals(JSONType) {
+		return nil
+	}
+
+	newSource := &dst.UnaryExpr{
+		X: &dst.CallExpr{
```
theunrepentantgeek, line 471: Use `astbuilder.CallQualifiedFunc()`

hack/generator/pkg/astmodel/conversion_function_builder.go
theunrepentantgeek: Or if this is the list, how about taking the singular form of the property name and suffixing `List` ?

theunrepentantgeek (changes requested):

hack/generator/pkg/astmodel/conversion_function_builder.go
```diff
@@ -0,0 +1,494 @@
...
+	// TODO: Dislike Typed suffix here -- change
+	tempVarIdent := builder.IdFactory.CreateIdentifier(params.NameHint+"Typed", NotExported)
+
+	newSource := &dst.UnaryExpr{
+		X:  params.ClonedSource(),
+		Op: token.MUL,
+	}
+
+	innerStatements := builder.BuildConversion(
+		params.WithDestination(dst.NewIdent(tempVarIdent)).
```
babbageclunk, line 202: I think this would be clearer if you just construct a new `ConversionParameters` and explicitly copy some fields from `params` and set some with new values. Otherwise you have to refer back to the struct def to work out which fields are the same and which are different.

babbageclunk (approved):
Looks good - although I agree with @theunrepentantgeek's suggestions too.

hack/generator/pkg/astmodel/conversion_function_builder.go
```diff
@@ -0,0 +1,494 @@
...
+		}
+	}
+
+	return result
+}
+
+func (params ConversionParameters) copy() ConversionParameters {
+	result := params
+	result.ConversionContext = nil
+	result.ConversionContext = append(result.ConversionContext, params.ConversionContext...)
```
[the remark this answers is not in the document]
matthchr: I think you misread this as 
`result.ConversionContext = append(result.ConversionContext, result.ConversionContext...)`
and not
`result.ConversionContext = append(result.ConversionContext, params.ConversionContext...)`

It's not duplicating the slice it's appending the `params` `ConversionContext` to `result`'s

hack/generator/pkg/astmodel/armconversion/convert_to_arm_function_builder.go
```diff
@@ -41,15 +42,16 @@ func newConvertToARMFunctionBuilder(
...
+		typeConversionBuilder: astmodel.NewConversionFunctionBuilder(c.idFactory, codeGenerationContext),
 	}
 
+	result.typeConversionBuilder.AddConversionHandlers(result.convertComplexTypeNameProperty)
 	result.propertyConversionHandlers = []propertyConversionHandler{
 		result.namePropertyHandler,
 		result.fixedValuePropertyHandler("Type"),
 		result.fixedValuePropertyHandler("ApiVersion"),
-		result.propertiesWithSameNameAndTypeHandler,
-		result.propertiesWithSameNameButDifferentTypeHandler,
```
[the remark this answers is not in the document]
matthchr: Because of the `definedErrVar` which I'm currently using to stick a `var err` at the top of functions that need it.
We need this because the following isn't valid code:
```
err := populateFromARM(foo)
err := populateFromARM(bar)
```

The current way I have things is kinda icky though because it means that these top level handlers need to understand a bit about what the lower level handler is going to do in order to determine if it needs to start with that `var err` or not. I think I see a way to fix this and make the `from` handler the same as the `to` handler... will investigate

hack/generator/pkg/astmodel/armconversion/convert_from_arm_function_builder.go
matthchr: Done

hack/generator/pkg/astmodel/armconversion/convert_from_arm_function_builder.go
```diff
@@ -208,42 +212,23 @@ func (builder *convertFromARMBuilder) propertiesWithSameNameAndTypeHandler(
...
-				destinationType:   toType,
-				nameHint:          string(toProp.PropertyName()),
-				conversionContext: nil,
-				assignmentHandler: nil,
-				sameTypes:         true,
+	return builder.typeConversionBuilder.BuildConversion(
+		astmodel.ConversionParameters{
+			Source: &dst.SelectorExpr{
+				X:   dst.NewIdent(builder.typedInputIdent),
+				Sel: dst.NewIdent(string(toProp.PropertyName())),
```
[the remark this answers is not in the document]
matthchr: Done (across all occurrences here and in the `_to_arm` builder as well).

hack/generator/pkg/astmodel/armconversion/convert_to_arm_function_builder.go
matthchr: Do you specifically mean the optional property handling here? Yes, in the convert to/from ARM case optionality is preserved between the two types always (no loss of information possible).

Definitely agree that these are some hurdles we would have to cross before merging the infrastructure for these two types of handlers (storage vs ARM). Part of merging them would also probably be more fine-grained control so that each usage (storage vs arm) could easily override the bits that made sense for it.

For now though I just maintained this behavior as that's what I had before and it wasn't clear that there was a generic solution that would work for both - which is to say I suspect the storage cases where only one side is optional we'd need handlers for those specific cases (in addition to the default "both" case)?

hack/generator/pkg/astmodel/conversion_function_builder.go
matthchr: Currently this is in the astmodel package and the ARM stuff is in a subpackage, so I can't make these fields private (not exported?) as that would prevent the usage from constructing them.

I could change the methods to `GetSource` + `GetDestination`?

hack/generator/pkg/astmodel/conversion_function_builder.go
matthchr: See comment above.

hack/generator/pkg/astmodel/conversion_function_builder.go
matthchr: We don't have easy access to the full `types` collection here... we do have `builder.CodeGenerationContext.GetTypesInCurrentPackage()` but that's only the types in _this_ package so won't work for types in other packages.

I added `CodeGenerationContext.GetAllReachableTypes()` and used that instead.

hack/generator/pkg/astmodel/conversion_function_builder.go
matthchr: Done

hack/generator/pkg/astmodel/conversion_function_builder.go
matthchr: Done - fixed other usages of UnaryExpr too

hack/generator/pkg/astmodel/conversion_function_builder.go
matthchr: Fixed

hack/generator/pkg/astmodel/conversion_function_builder.go
matthchr: `CallQualifiedFunc` expects that I have a `string` for `qualifier` but here I have an `Expr`. Given that seems to be an "abnormal" case (we have a good number of other usages of `CallQualifiedFunc`, and the fact that we have helpers for the other bits here I'm not sure it makes sense adding a helper to `astbuilder` to cater to this scenario.

The code I have right now is:
```
newSource := astbuilder.Dereference(
		&dst.CallExpr{
			Fun: astbuilder.Selector(params.ClonedSource(), "DeepCopy"),
			Args: []dst.Expr{},
		})
```

which isn't too bad?

hack/generator/pkg/astmodel/conversion_function_builder.go
matthchr: I'd like to make some improvements to the naming, but I think I should leave that part for a follow up PR, as any changes to how we generate names in the code is going to explode the PR side in the generated files - thoughts?

hack/generator/pkg/astmodel/conversion_function_builder.go
matthchr: Same comment as above about holding off on generated name changes until a second PR to ease review burden.

hack/generator/pkg/astmodel/conversion_function_builder.go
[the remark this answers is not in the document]
matthchr: Nevermind ignore the above - you're right my own code confused me too. Changed.

hack/generator/pkg/astmodel/armconversion/convert_to_arm_function_builder.go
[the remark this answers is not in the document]
matthchr: I have a fix for this - it's with a bunch of other renaming stuff though so will include it in a second PR I think

hack/generator/pkg/astmodel/conversion_function_builder.go
theunrepentantgeek: What if you introduced a constructor/factory method so that direct construction wasn't needed?

If that's not feasible, then `GetSource()` etc would be fine.

hack/generator/pkg/astmodel/conversion_function_builder.go
theunrepentantgeek: Followup PR is fine.

hack/generator/pkg/astmodel/conversion_function_builder.go
theunrepentantgeek: Yeah, that's better than it was. 👍

theunrepentantgeek (approved):

hack/generator/pkg/astmodel/conversion_function_builder.go
matthchr: The issue with constructor/factory method is that this type has 8 properties, some of which are regularly `nil`. An 8 parameter function is significantly less understandable/readable than an 8 parameter struct instantiation I think.

One approach that could get these names nicer would be to duplicate the `ConversionParameters` structure and turn it into a `parameters` struct that's only used to supply arguments to the factory function. Then make everything not exported on `ConversionParameters` so that we could use `Source` and `Destination`. That's quite a bit of extra code though and feels like it might be a bit overengineered.

Will just change to `GetX` for now I think

PR #449, opened 2021-05-04
Resource Properties and Functions

theunrepentantgeek:
In order to generate storage conversion functions for resource types, they need to have properties - and for the generated function to be written, they need to have functions too.

Since the change isn't really to do with storage functions, I figured it made sense to split it out as a separate PR.

PR includes a couple of simple interfaces, but regrettably can't include the Withers due to lack of generics and covariance on return types.


hack/generator/pkg/astmodel/resource_type.go
```diff
@@ -284,32 +370,34 @@ func (resource *ResourceType) RequiredPackageReferences() *PackageReferenceSet {
...
-			fields = append(fields, defineField("Status", statusType, "`json:\"status,omitempty\"`"))
+	for _, property := range resource.Properties() {
+		f := property.AsField(codeGenerationContext)
+		if f != nil {
+			fields = append(fields, f)
 		}
 	}
 
+	if len(fields) > 0 {
+		// if first field has Before:EmptyLine decoration, switch it to NewLine
```
babbageclunk, line 396: Is this comment right? Should the code be checking for that decoration somewhere?

babbageclunk (approved):
Looks good. How are you going to handle the `With*` methods - something like this?

```go
func WithFunction(target FunctionContainer, f Function) FunctionContainer {
    // Type-switch to dispatch to the different WithFunction methods
}
```

hack/generator/pkg/astmodel/resource_type.go
theunrepentantgeek: Comment originated in code I copied from `ObjectType` - I've ammended both to make them more accurate.

theunrepentantgeek:
> Looks good. How are you going to handle the With* methods

I wasn't going to do anything special - can't express them in the interface, so I'm leaving it for consumers to detect the implementing type and do the right thing. Admittedly, it's a bit of a wart.

PR #450, opened 2021-05-04
Add spec for defaulter/validator

matthchr:


hack/generator/spec/defaulter-validator.md
```diff
@@ -0,0 +1,100 @@
...
+func (storageAccount *StorageAccount) Default() {
+	defaults := storageAccount.defaults()
+	var temp runtime.Object = storageAccount
+	if runtimeDefaulter, ok := temp.(genruntime.Defaulter); ok {
+		defaults = runtimeDefaulter.Defaults()
+	}
+
+	for _, defaultFunc := range defaults {
+		defaultFunc()
+	}
```
babbageclunk, line 38: I didn't click why you needed the cast to `runtime.Object` at first - it's because you can't do type assertions on non-interface values, right? I don't think there is a better way - although since it's nothing specific to `runtime.Object` maybe 
`interface{}` would be a better choice. 

It's quite a weird situation really: "I'm doing an interface check on my own type because I don't know whether I implement this method".

How do you envisage providing custom Defaults implementations?

hack/generator/spec/defaulter-validator.md
```diff
@@ -0,0 +1,100 @@
...
+// +kubebuilder:webhook:path=/mutate-microsoft-storage-infra-azure-com-v1alpha1api20190401-storageaccount,mutating=true,sideEffects=None,matchPolicy=Exact,failurePolicy=fail,groups=microsoft.storage.infra.azure.com,resources=storageaccounts,verbs=create;update,versions=v1alpha1api20190401,name=default.v1alpha1api20190401.storageaccounts.microsoft.storage.infra.azure.com,admissionReviewVersions=v1beta1
+
+var _ admission.Defaulter = &StorageAccount{}
+
+// Default defaults the Azure name of the resource to the Kubernetes name
+func (storageAccount *StorageAccount) Default() {
+	defaults := storageAccount.defaults()
+	var temp runtime.Object = storageAccount
+	if runtimeDefaulter, ok := temp.(genruntime.Defaulter); ok {
+		defaults = runtimeDefaulter.Defaults()
```
babbageclunk, line 33: Wouldn't we want the collected default funcs to be added with the base one? Allowing users to override azure name defaulting (or more likely forget to do it in a default func they provide themselves) seems dangerous.

hack/generator/spec/defaulter-validator.md
```diff
@@ -0,0 +1,100 @@
...
+func (storageAccount *StorageAccount) ValidateCreate() error {
+	validations := storageAccount.createValidations()
+	var temp runtime.Object = storageAccount
+	if runtimeValidator, ok := temp.(genruntime.Validator); ok {
+		validations = runtimeValidator.CreateValidations()
+	}
+
+	for _, validation := range validations {
+		err := validation()
+		if err != nil {
```
babbageclunk, line 70: It might be worth running all the validations and aggregating errors here.

hack/generator/spec/defaulter-validator.md
matthchr: In a handcrafted file alongside the generated code was the assumption. In general I feel like this is a pattern we might want in multiple places (not just here). For example I was imagining @theunrepentantgeek would end up with something similar for conversions eventually for the places we need to deal with manual fixups for them.

hack/generator/spec/defaulter-validator.md
matthchr: I was thinking that the implementer could choose to call this in their `genruntime.Defauler` implementation if they wanted the existing defaults, otherwise there's no way to bypass an autogenerated default... but maybe we dont want that anyway. We could start out without the ability to do that I guess and then just add it if we ever needed it - fair enough.

hack/generator/spec/defaulter-validator.md
matthchr: Changed this to `interface{}`. Also agree that this is weird... is there a less weird way to do this?

Basically I want to encode a hook into the generated code such that we can put handcrafted code "somewhere else" and have it called from here.

hack/generator/spec/defaulter-validator.md
matthchr: Updated

hack/generator/spec/defaulter-validator.md
matthchr: @theunrepentantgeek actually talks about this same approach here: https://github.com/Azure/k8s-infra/blob/master/hack/generator/spec/versioning.md#standard-extension-points

The only difference is that his proposal has an interface per-type and here we just have a single generic interface, but the same "implement an interface that you don't know about" approach is mentioned.

babbageclunk: > In a handcrafted file alongside the generated code was the assumption.

Oh, I was worried about something like that going away when we regenerate, but I guess we'll avoid deleting those non-generated files.

hack/generator/spec/defaulter-validator.md
babbageclunk: Ah - in the example it doesn't look like the base implementation would be available to call so if someone provided their own implementation of Defaulter they'd need to rewrite the base one, but you're saying it would be public. In that case I think it's fine - definitely more flexibility if they can choose whether it happens.

matthchr: It doesn't actually have to be public does it? Since the override has to come from inside this package?

hack/generator/spec/defaulter-validator.md
matthchr: Yeah we already avoid deleting the non-generated files. I think we look at the `_gen` suffix as well as the first N lines for the comment saying they're generated.

Porges (approved):

PR #451, opened 2021-05-06
Turn on shadow-checking in golangci-lint configuration

babbageclunk:
This might make it easier to find bugs caused by inadvertently creating a new binding rather than assigning to one in an outer scope. This has bitten most of us at some time or another and always takes way longer to work out than I'd like.

Unfortunately the linter can only find potential shadowing problems if the `generator:lint` or `controller:lint` tasks are run, so it probably needs to be paired with editor support (like [this feature](https://blog.jetbrains.com/go/2017/09/04/gogland-eap-13-better-completion-new-inspections-fixed-performance-bugs-and-more/) in Goland). In the ASO repo we run `go vet` before building - is it worth doing that here for this kind of thing? (That's likely side-stepped by IDEs or running `go test` directly though.)


Porges (approved):

matthchr:
> Unfortunately the linter can only find potential shadowing problems if the generator:lint or controller:lint tasks are run

At least for me, I run `task` (which runs everything) pretty regularly when developing, so this doesn't seem like a huge limitation. I think that the shadowing feature in GoLand is on by default now? I think it's on for me but I never turned it on.

matthchr (approved):

PR #455, opened 2021-05-10
Create storage conversion helper functions

theunrepentantgeek:
Create storage variants of all resources and complex object types, injecting helper conversion functions for conversion back and forth. These helper functions will be used to build the actual conversions in a later PR.

**Reviewer guide**

The crux of operation is the StorageTypeFactory (see storage_type_factory.go) which uses three distinct TypeVisitors to make the required changes to each class.

Also includes

* Some improvements to debugger output for various types
* Use of new abstractions PropertyContainer and FunctionContainer
* Improvements to field & parameter names, and other code gardening


hack/generator/pkg/astmodel/testdata/ConvertBetweenAliasAndAliasType.golden
```diff
@@ -10,7 +10,7 @@ type Person struct {
 }
 
 // ConvertFromVNext populates our Person from the provided source Person
-func (person Person) ConvertFromVNext(source *vNext.Person) error {
+func (person *Person) ConvertFromVNext(source *vNext.Person) error {
```
babbageclunk, line 13: I can see why this one needs a pointer, since it's mutating person.

hack/generator/pkg/astmodel/testdata/ConvertBetweenAliasAndAliasType.golden
```diff
@@ -20,7 +20,7 @@ func (person Person) ConvertFromVNext(source *vNext.Person) error {
 }
 
 // ConvertToVNext populates the provided destination Person from our Person
-func (person Person) ConvertToVNext(destination *vNext.Person) error {
+func (person *Person) ConvertToVNext(destination *vNext.Person) error {
```
babbageclunk, line 23: No sure about this one - is it just for consistency? I guess they might be quite big.

hack/generator/pkg/astmodel/type_flag.go
```diff
@@ -29,6 +29,7 @@ func (f TypeFlag) ApplyTo(t Type) *FlaggedType {
 
 // RemoveFrom applies the tag to the provided type
 func (f TypeFlag) RemoveFrom(t Type) (Type, error) {
+
```
babbageclunk, line 32: Rogue newline?

hack/generator/pkg/astmodel/type_visitor.go
```diff
@@ -178,12 +178,12 @@ func IdentityVisitOfObjectType(this *TypeVisitor, it *ObjectType, ctx interface{
 func IdentityVisitOfMapType(this *TypeVisitor, it *MapType, ctx interface{}) (Type, error) {
 	visitedKey, err := this.Visit(it.key, ctx)
 	if err != nil {
-		return nil, errors.Wrapf(err, "failed to visit map key type %T", it.key)
+		return nil, errors.Wrapf(err, "failed to visit map key type %q", it.key.String())
```
babbageclunk, line 181: You don't need the `.String()` calls here: the `%q` formatter will call `.String()` on any `Stringer`.

hack/generator/pkg/codegen/storage/storage_type_factory.go
```diff
@@ -0,0 +1,465 @@
...
+	pendingMarkAsHubVersion    astmodel.TypeNameQueue     // Queue of types that need to be flagged as the hub storage version
+	idFactory                  astmodel.IdentifierFactory // Factory for creating identifiers
+	storageConverter           astmodel.TypeVisitor       // a cached type visitor used to create storage variants
+	propertyConverter          astmodel.TypeVisitor       // a cached type visitor used to simplify property types
+	functionInjector           astmodel.TypeVisitor       // a cached type visitor used to inject functions into definitions
+	resourceHubMarker          astmodel.TypeVisitor       // a cached type visitor used to mark resources as Storage Versions
+
+	// Map of conversion links for creating our conversion graph
+	// (Can't use PackageReferences as keys, so keyed by the full package path)
+	conversionMap map[string]astmodel.PackageReference
```
babbageclunk, line 33: Is this right? There are other places we use `PackageReference`s as keys (like in `PackageReferenceSet`). Didn't you mention that all the implementations have value semantics?

hack/generator/pkg/codegen/storage/storage_type_factory.go
```diff
@@ -0,0 +1,465 @@
...
+	}.Build()
+
+	result.functionInjector = astmodel.TypeVisitorBuilder{
+		VisitObjectType:   result.injectFunctionIntoObject,
+		VisitResourceType: result.injectFunctionIntoResource,
+	}.Build()
+
+	result.resourceHubMarker = astmodel.TypeVisitorBuilder{
+		VisitResourceType: result.markResourceAsStorageVersion,
+	}.Build()
```
babbageclunk, line 74: I think these 4 type visitors should be separated out of `StorageTypeFactory` - probably as 4 separate types each with their own callback methods. I haven't read through them yet but I'm assuming the callbacks aren't all manipulating shared state on result? Splitting them out into separate types would make that clearer (or at least make it obvious what state is shared between them), and hopefully make each visitor more coherent - as it is they're all intermingled into one type.

babbageclunk (changes requested):
I'm submitting this not-quite-finished because I need to head but wanted you to be able to see the current comments. Main thing is that I think the type visitors + callback methods in StorageTypeFactory should be split out into smaller types to make clearer what state they're using and sharing between them.

hack/generator/pkg/codegen/pipeline_create_storage_types.go
```diff
@@ -7,247 +7,69 @@ package codegen
...
-					// Skip Enum definitions as we use the base type for storage
+				factory, ok := factories[ref.Group()]
+				if !ok {
+					klog.V(3).Infof("Creating storage factory for %s", ref.Group())
+					factory = storage.NewStorageTypeFactory(ref.Group(), idFactory)
+					factories[ref.Group()] = factory
+				}
+
+				isArm := astmodel.ARMFlag.IsOn(def.Type())
+				if isArm {
```
matthchr, line 45: minor: probably can drop the `isArm` here, as it's not adding a lot of clarity.

If you keep it, it should be `isARM`.

hack/generator/pkg/codegen/pipeline_ensure_type_has_arm_type.go
```diff
@@ -45,7 +45,13 @@ func validateAllTypesHaveARMType(definitions astmodel.Types) error {
 
 	var errs []error
 
-	for _, def := range definitions {
+	for name, def := range definitions {
```
matthchr, line 48: minor: Wonder if this method name should be changed to `valdiateExpectedTypesHaveARMType` or something?

hack/generator/pkg/codegen/storage/storage_type_factory.go
matthchr: yeah IIRC we can use `PackageReference` as keys and it works.

hack/generator/pkg/codegen/storage/storage_type_factory.go
```diff
@@ -0,0 +1,465 @@
...
+
+	return result
+}
+
+func (f *StorageTypeFactory) Add(def astmodel.TypeDefinition) {
```
matthchr, line 79: minor: missing comment

hack/generator/pkg/codegen/storage/storage_type_factory.go
```diff
@@ -0,0 +1,465 @@
...
+
+	"github.com/Azure/k8s-infra/hack/generator/pkg/astmodel"
+	"github.com/pkg/errors"
+	kerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/klog/v2"
+)
+
+// StorageTypeFactory is used to create storage types for a specific service
+type StorageTypeFactory struct {
+	service                    string                     // Name of the service we're handling (used mostly for logging)
```
matthchr, line 19: A bit thrown by this nomenclature - is `service` a group? Why not just call it `group` (which is nomenclature we already have) rather than introduce a new concept (even if it really is the same concept).

hack/generator/pkg/codegen/storage/storage_type_factory.go
```diff
@@ -0,0 +1,465 @@
...
+ */
+
+// useBaseTypeForEnumerations replaces an enumeration with its underlying base type
+func (f *StorageTypeFactory) useBaseTypeForEnumerations(
+	tv *astmodel.TypeVisitor, et *astmodel.EnumType, ctx interface{}) (astmodel.Type, error) {
+	return tv.Visit(et.BaseType(), ctx)
+}
+
+// shortCircuitNamesOfSimpleTypes redirects TypeNames that reference resources or objects into our
+// storage namespace, and replaces TypeNames that point to simple types (enumerations or
```
matthchr, line 408: Is the main reason that you need to do this for things like validated types where they had an alias before but since you're dropped validations there's no need for an alias now?

matthchr (approved):
LGTM - left a laundry list of relatively minor comments

hack/generator/pkg/astmodel/storage_conversion_factories.go
```diff
@@ -123,10 +124,16 @@ func createTypeConversion(
 		}
 	}
 
+	var debugDescriptionOfDestination strings.Builder
+	destinationEndpoint.Type().WriteDebugDescription(&debugDescriptionOfDestination, conversionContext.types)
+
+	var debugDescriptionOfSource strings.Builder
+	sourceEndpoint.Type().WriteDebugDescription(&debugDescriptionOfSource, conversionContext.types)
```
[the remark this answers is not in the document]
theunrepentantgeek: Fair enough. I'll do a separate PR.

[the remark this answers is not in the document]
theunrepentantgeek: https://github.com/Azure/azure-service-operator/pull/1486

hack/generator/pkg/astmodel/storage_conversion_context.go
```diff
@@ -12,21 +12,40 @@ type StorageConversionContext struct {
 	types Types
 	// functionName is the name of the function we're currently generating
 	functionName string
+	// knownLocals is a reference to our set of local variables
+	knownLocals *KnownLocalsSet
```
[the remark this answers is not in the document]
babbageclunk: Yeah, fair points - reading that link I think I was wrong about this. My feeling was that `KnownLocalsSet` is kind of a map and so the distinction between nil and empty is confusing - but actually we have that same confusing distinction with maps. Comment rescinded!

PR #456, opened 2021-05-10
Add ResourceReferences pipeline stage

matthchr:
This new stage detects cross resource references and replaces them with the ResourceReference type.


matthchr:
Review in ASO repo instead.

